PR #15, opened 2021-09-26
Feat/signup page

maximenathan:
Enable already created `signup page` and add a `forgot password` page

link to https://github.com/GrottoCenter/Grottocenter3/pull/516

Peek 2021-09-26 16-40

**TODO**: Check if emails are sent


urien:
pour moi c'est parfait

Clm-Roig (approved):

PR #57, opened 2021-11-09
[STYLE] Replace React logo by Grottocenter one + update description and title

Clm-Roig:
When sharing on Slack for example: 

## BEFORE

image

## AFTER

image


urien:
Très bonne idée, merci

bsoufflet (approved):

PR #387, opened 2022-04-06
[DOC FILE] increase file size from 100Mo to 2Go

Clm-Roig:
A message specifying the max size allowed is now displayed.


urien (approved):
Merci

PR #467, opened 2022-05-20
fix(340): err create massif

lageta:
fix error on create massif page when massifValues was undefined, access to the geoJson proprety raise an error


urien (approved):
Cela foncrionne

Clm-Roig (approved):

PR #714, opened 2022-08-12
[REFACTOR] Entrance actions

Clm-Roig:
- Rename entry => entrance
- Use an action entrance folder 

EntranceReducer will be splitted in multiple reducers (CreateEntrance, UpdateEntrance) later.


packages/web-app/src/reducers/RandomEntranceReducer.js
```diff
@@ -0,0 +1,26 @@
+import {
+  FETCH_RANDOM_ENTRANCE,
+  FETCH_RANDOM_ENTRANCE_SUCCESS,
+  FETCH_RANDOM_ENTRANCE_FAILURE
+} from '../actions/RandomEntrance';
+
+const initialState = {
+  isFetching: false, // show loading spinner
+  entrance: undefined, // random entry
```
maximenathan:
```suggestion
const initialState = {
  isFetching: false,
  entrance: undefined,
```

maximenathan (approved):
C'est bon pour moi!

Issue #445, opened 2022-05-11
Add a loading screen while the react app is preparing and JS is loading


bsoufflet:
The Grottocenter front app is slow to load due to large javascript file to load the first time (then in cache) and due to the react DOM rendering.

We should maybe display a loading screen instead of a white page which may feel like a bug.

Maybe this solution could work while waiting for the React app to render all components : https://betterprogramming.pub/a-quick-and-easy-react-js-loading-screen-with-hooks-940feccd553f
But the above solution may not work during the downloading of the large javascript files (8+ Mo)...

Clm-Roig:
Why not rendering a spinner directly in the index.html file ? And remove it when the window is ready using vanilla JS : 

```js 
window.onload = function(){
    // remove spinner
};
```

Souterweb:
Testé le logo d'attente : c'est excellent ! 
On en arrive presque a regretter l'époque où c'était bien plus lent pour avoir le temps de l'admirer...  ;o)
Bravo.
Domi

urien:
Traité par Vincent : super fonctionnalité !

PR #819, opened 2023-01-01
Performance: Reduce the initial js bundle size

vmarseguerra:
By making use of react code splitting and removing barely used heavy libraries

Also adds a loader while the application is loading (#445)

Current size 4.5Mo
Major changes:
- lodash: 110Ko (removed)
- jwt: 460ko (removed)
- lang: 803ko (loaded dynamicly)
- swagger-ui: 1Mo (splitted)
- D3.js + hexbin: 330Ko (splitted)
- date-fns: 100Ko (splitted)
- proj4: 85Ko (splitted)
- leaflet-draw 70Ko (splitted)

New initial bundle size 1.68Mo

Each locale strings are now loaded dynamically when the user switch. For performance the fetching of the initial locale strings starts directly from the html page.

All the locale strings file have been moved to the `public` folder, is there also configurations to change for the transifex bot ?

To analyse the bundle I use the following commands:
```
yarn build
npx source-map-explorer packages/web-app/build/static/js/*.js
```

Current bundle + New bundle + Loader preview: 
[preview.zip](https://github.com/GrottoCenter/grottocenter-front/files/10328896/preview.zip)


urien (commented):
C'est super, le temps de chargement est beaucoup plus court et l'image d'attente est très sympa et permettra de faire comprendre aux utilisateurs avec une connexion limité que des choses sont en cours de chargement
Merci !

Clm-Roig:
> All the locale strings file have been moved to the public folder, is there also configurations to change for the transifex bot ?

Yes there is something to change ! I can handle this part when the PR is merged. 

I will review your code later, thank you very much for your work.

packages/web-app/src/components/common/Pageloader.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+const Container = styled.div`
+  width: 100%;
+  height: 200px;
+  display: flex;
+  justify-content: center;
+  align-items: flex-end;
+`;
+
+const Pageloader = () => (
```
Clm-Roig: `PageLoader` with a "L" instead of "l"

packages/web-app/public/index.html
```diff
@@ -47,6 +41,137 @@
...
+            stroke="#056598"
+            stroke-width="25"
+            stroke-linecap="round"
+            stroke-dashoffset="391"
+            stroke-dasharray="195.498402"
+          />
+        </svg>
+      </div>
+    </div>
+    <style>
```
Clm-Roig, line 41: Maybe it should be in a separate `style.css` file ?

packages/web-app/public/index.html
```diff
@@ -47,6 +41,137 @@
...
+      }
+      #root {
+        transition: opacity 0.4s;
+        opacity: 0;
+      }
+      .rootDisplay {
+        opacity: 1 !important;
+      }
+    </style>
+    <script>
```
Clm-Roig, line 143: Same as my previous comment: move it into another script file ?

packages/web-app/src/actions/Intl.js
```diff
@@ -1,6 +1,50 @@
...
+    .then(checkAndGetStatus)
+    .then(response => response.json())
+    .then(data => dispatch(changeLocaleLoadSuccess(locale, data)))
+    .catch(errorMessage => {
+      dispatch(changeLocaleLoadFailure(errorMessage));
+    });
+};
+
+export const bootstrapIntl = () => dispatch => {
+  /* global intlBootstrap */
```
Clm-Roig: Add 
```js
globals: {
    intlBootstrap: true
}
```

to `web-app/.eslintrc.js` instead.

packages/web-app/src/pages/Application.jsx
```diff
@@ -42,12 +42,32 @@ import OrganizationEdit from './EntityEdit/Organization/OrganizationEdit';
 import CountryPage from './Country';
 import NotificationsPage from './Notifications';
 
+async function transitionToReact() {
+  /* global intlBootstrap */
```
Clm-Roig: Put it in eslintrc (see previous comment about it)

packages/web-app/src/reducers/IntlReducer.js
```diff
@@ -1,27 +1,55 @@
...
+  CHANGE_LOCALE_SUCCESS,
+  CHANGE_LOCALE_LOAD_SUCCESS,
+  CHANGE_LOCALE_LOAD_FAILURE
+} from '../actions/Intl';
 
-const navigatorLocale = navigator.language.split(/[-_]/)[0];
-const initialLocale = includes(navigatorLocale, keys(AVAILABLE_LANGUAGES))
-  ? navigatorLocale
-  : DEFAULT_LANGUAGE;
+/* global intlBootstrap */
```
Clm-Roig: Put it in eslintrc instead (see previous comment about it)

packages/web-app/src/reducers/LoginReducer.js
```diff
@@ -1,4 +1,4 @@
-import { decode } from 'jsonwebtoken';
+// import { decode } from 'jsonwebtoken';
```
Clm-Roig: To delete

Clm-Roig (changes requested):

packages/web-app/public/index.html
vmarseguerra: Usually I would agree, but in this case specifically, we want the loader to be displayed as fast as possible and splitting this CSS and JS in other files will delay the display

Clm-Roig: I understand so importing it is not an option as it will cause the page to "flash" with the unstyled content before the CSS is loaded ? 

But `style` is not allowed in the `body` tag.

According to [this post](https://stackoverflow.com/questions/49523627/i-want-to-load-specific-stylesheet-first-in-html-before-other-sylesheets-and-bod), the best solution is to hardcode the CSS in `head` : it's W3C compliant + it's loaded before the `body`.

vmarseguerra: Whoa! I didn't know that the style tag in body what such a controversial topic, and that the _current_ spec does not allow it ...
https://stackoverflow.com/questions/2830296/using-style-tags-in-the-body-with-other-html#answer-50154225
https://github.com/whatwg/html/issues/1605

Ok let move the CSS into the head

packages/web-app/public/index.html
```diff
@@ -34,19 +37,142 @@
...
+      const intlBootstrap = {
+        defaultLocale: 'en',
+        allLanguages: [
+          'ar',
+          'bg',
+          'ca',
+          'de',
+          'el',
+          'en',
+          'es',
```
bsoufflet, line 154: Is there a way to not hardcode this list ? Otherwise we will need to remember to update it if we add an extra language no ?

bsoufflet (approved):

packages/web-app/public/index.html
vmarseguerra: This list is duplicated from the `src/conf/config.js` file, which at this time is not yet loaded. This not ideal but it is a tradeoff for a faster loading speed.

To make it less bad, there is at boot a check between the intlBootstrap and the React config that log a warning if the length of both lists are not equal. I also just added comments on both lists with a reminder to update both of them

bsoufflet (approved):

PR #835, opened 2023-02-01
feat(performance): only loads api languages when required

vmarseguerra:
Only loads languages from the api when they are needed instead of every time at startup.

Also removes unused files.


bsoufflet (approved):

urien:
Excuses moi Vincent, j'ai mergé pour que les étudiants aient accès aux serveurs de dev.

Issue #847, opened 2023-02-05
cannot set a negative lat/lon coordinate


robertolumbreras:
Editing or adding a coordinate the web interface just rejects the minus sign and only allows entering a positive number, but you can move the point and the negative lat/lon are displayed and set.

urien:
Edited by Vincent. Thank you for this issue @rover0

PR #848, opened 2023-02-06
fix(forms): entity forms fixes

vmarseguerra:
- Fixed: InputCoordinate does not allow negative (#847)
- Fixed: PolygonMap edit on massif only works once (#838)
- Fixed: Massif description is not created (#834)
- Fixed: Minor React errors (propTypes and others)


urien:
Merci Vincent. Je ne peux pas tester mais c'est super que tu ais pris en compte ces issues.

bsoufflet (approved):

PR #860, opened 2023-02-14
feat(map): add new layers like ign scan 25

### LINKED ISSUE #681: [MAP] Add Basemaps
Pour les utilisateurs il serait très utile de disposer de davantage de fonds de carte.

Les utilisateurs français nous réclament en particulier l'accès aux fond de carte IGN : SCAN 25, fond de carte géologique
Les modalités sont décrites ici
https://geoservices.ign.fr/documentation/services/utilisation-web/extension-pour-leaflet

D'autres fond de carte sont présentés ici
https://leaflet-extras.github.io/leaflet-providers/preview/

bsoufflet:
Also : 
- add new layers like ign scan 25 (+ BRGM layers)
- add back depth, length and cave name in map popup.
- save in localstorage the latest selected base layer and reuse it
- Use our custom LayersControl component in all maps
- fix zoom to close on entrance page

Fix #681

Warning : IGN scan 25 may not be visible on test server because of the IGN key being protected by referer website

Warning : Require this change to be deployed on the API : https://github.com/GrottoCenter/grottocenter-api/pull/1137

## 📸 Previews
image


urien (commented):
Super !

packages/web-app/src/components/common/Maps/common/LayersControl.jsx
```diff
@@ -20,7 +21,21 @@ const LayersControl = ({
...
+          />
+        )}
+        {layer.type === 'WMS' && (
+          <WMSTileLayer
+            attribution={layer.attribution}
+            layers={layer.layers}
+            url={layer.url}
+          />
+        )}
       </LeafletLayersControl.BaseLayer>
```
vmarseguerra, line 43: The layers list is also used in [PolygonMap.jsx](https://github.com/GrottoCenter/grottocenter-front/blob/develop/packages/web-app/src/components/appli/EntitiesForm/Massif/PolygonMap.jsx) and [MapMarkerSelector.jsx](https://github.com/GrottoCenter/grottocenter-front/blob/develop/packages/web-app/src/components/appli/EntitiesForm/utils/MapMarkerSelector.jsx)

Is it possible to make the change to take into account the layer type also there? 
Or may be it can be factored to avoid code duplication?

packages/web-app/src/components/common/Maps/common/Markers/Components/EntrancePopup.jsx
```diff
@@ -25,17 +28,17 @@ export const EntrancePopup = ({ entrance }) => (
...
     )}
-    {entrance.cave && entrance.cave.length && (
+    {entrance.length && (
       <Property
         secondary
-        value={`${entrance.cave.length} m`}
+        value={`${entrance.length} m`}
         icon={<CustomIcon size={25} type="length" />}
       />
     )}
```
vmarseguerra, line 44: Maybe I missed something, but I couldn't test those changes

This marker seems to be used on the main map and on the massif map:
- On the massif map they are not displayed because the API will never return any entrances (the code is commented) [grottocenter-api/api/controllers/v1/massif/find.js](https://github.com/GrottoCenter/grottocenter-api/blob/develop/api/controllers/v1/massif/find.js#L33)
- On the main map the API will also never return the length, depth or cave name because of missing quotes in the SQL Query. Without quotes, select alias are lowercased (idCave -> idcave) but the code doesn't expect the case to have changed ([grottocenter-api/api/services/GeoLocService.js](https://github.com/GrottoCenter/grottocenter-api/blob/develop/api/services/GeoLocService.js#L4))

After adding quotes in the SQL, the entrance object will look like this: 
```
{
    id: 21,
    name: 'Saints de Glace (Les)',
    city: 'Autrans-Méaudre-en-Vercors',
    region: 'Isère (38), Auvergne-Rhône-Alpes (ARA)',
    cave: { id: 75075, name: null, depth: 670, length: 49300 },
    longitude: 5.520071983337401,
    latitude: 45.152929703286,
    quality: 10
}
```

Here the key path to depth, length where correct no need to change them.

vmarseguerra (changes requested):

packages/web-app/src/components/common/Maps/common/Markers/Components/EntrancePopup.jsx
bsoufflet: @vmarseguerra : I forgot to create the PR for the backend part. Please see it here : https://github.com/GrottoCenter/grottocenter-api/pull/1137

With this change in the backend API it should make more sense.

packages/web-app/src/components/common/Maps/common/LayersControl.jsx
bsoufflet: I will try to refactor to use the same base layer component for all maps of grottocenter...

bsoufflet:
I will also try to save in local storage the base layer selected by the user :)

bsoufflet:
@urien @vmarseguerra : 
- save in localstorage the latest selected base layer and reuse it
- Use our custom LayersControl component in all maps
- fix zoom to close on entrance page

vmarseguerra (approved):

PR #893, opened 2023-03-21
fix(entrance): broken "all revisions" page due to translation

bsoufflet:
Fix BUG  #891

## 🤔 What  
"All revisions" page does not display due to a front end bug with the translation of a title...

## 🧪 Testing
Go to `ui/entrances/73545/snapshots?isNetwork=false&all=true` using the azure link


urien (approved):
Cela fonctionne bien
Merci Benjamin

PR #930, opened 2023-08-20
Fix Build and Deploy CI

vmarseguerra:
# Fix Build and Deploy CI

This is an attempt to fix the [Build and Deploy github action](https://github.com/GrottoCenter/grottocenter-front/actions/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml) that has [failed to run](https://github.com/GrottoCenter/grottocenter-front/actions/runs/5894032591/job/16044266574) in the [last PR](https://github.com/GrottoCenter/grottocenter-front/pull/929). The error was during the yarn install:
`YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.`

The last time this action [was successful](https://github.com/GrottoCenter/grottocenter-front/actions/runs/5849018555/job/15856900862) was just a few days before on the same PR when it was deployed on the stage site.
It seems the only difference is that it uses a different version of Yarn: Before it was `4.0.0-rc.48` and now it is `4.0.0-rc.49`

It is weird that the CI image does not use the specified yarn version in the `package.json` which should be `yarn@3.2.1`

In the build log, from the line: `Found yarn version spec to follow: '>=3.0.*'` It seams Oryx (which is the tool used to build the project) use the `engines` spec to [determine the yarn version](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl#L26).

So changing the `engines` spec should I hope make the CI use the correct yarn version

I have also updated yarn to the latest stable version available (3.6.2) using `yarn set version stable`


urien (approved):
Cela fonctionne bien sur le serveur de Dev.
Merci Vincent

PR #958, opened 2024-03-22
Fixes small issues about documents

vmarseguerra:
Fixes small issues about documents

- Cannot select more than one bbs subject in form
- Broken caver link in document details

https://grottocenter.org/ui/documents/85186

Thanks @ClemRz for finding them


urien (approved):
Cela fonctionne parfaitement désormais

PR #961, opened 2024-03-25
fix(entrance): error when adding sub entities

vmarseguerra:
Fixes the error that occur error when trying to add a sub entities (location, description, rigging, document, comment)


urien (approved):
Cela fonctionne très bien.
Merci

PR #977, opened 2024-08-20
fix(caver): error when selecting a user in dashboard #972

### LINKED ISSUE #972: [Admin] select user
**Describe the bug**
Comme administrateur si je souhaite sélectionner un utilisateur, le moteur de recherche fonctionne mais quand je clique sur la personne à laquelle je souhaite donner des droits, cela conduit a la page d'erreur.

vmarseguerra:
Fixes the error when selecting a user in dashboard.

When the caver is not part of any groups, the `groups` key in the search result is sometimes not present, causing issues

Should fix #972


urien (changes requested):
Cela fonctionne parfaitement

urien (approved):

PR #989, opened 2024-09-10
Adds delete / restore for Cave and Entrance + Refactoring

vmarseguerra:
Adds the possibility to delete, view deleted, restore, delete permanently for the following entities:

- Network (cave)
- Entrance

Only moderators can do these actions.
When deleting, there is a possibility to redirect to another entity of the same type.
When permanently deleting, there is a possibility to merge the linked entities to another entity of the same type.

As a cave can contain 1 or more entrances, the following strategy is applied when permanently deleting:
- For a cave, another cave is required to be selected to move the associated entrances
- For an entrance, the associated cave is also deleted if it was the only associated entrance of this cave

Refactoring and cleanup were also made.

In order to test this, the PR linked in the API (https://github.com/GrottoCenter/grottocenter-api/pull/1308) needs to be deployed


urien:
Bravo, le processus que tu as mis en place est parfait

bsoufflet (approved):

PR #991, opened 2024-09-23
Adds delete for person + Refactoring

vmarseguerra:
Adds the possibility to delete permanently Caver
When permanently deleting, there is a possibility to merge the linked entities to another entity of the same type.

In Grottocenter there are 2 types of user (person):
- Caver: User who registered, have a login and password (~3000 users)
- Author: User created automatically, attached to documents, email ends with '@mail.no' and does not have a password (~56600 users)

There is no soft delete for caver, deletion is permanent and no restore is possible.

- Only administrators can delete a caver user.
- Only moderators can delete an author user.

Refactoring and cleanup were also made:
- The caver edit page is now a single form instead of a multistep form
- A button to view the caver full page was added in the dashboard

In order to test this, the PR linked in the API (https://github.com/GrottoCenter/grottocenter-api/pull/1316) needs to be deployed


urien:
Le fonctionnement proposé est parfait.
Merci Vincent

bsoufflet (approved):

PR #1099, opened 2025-11-28
fix(translations): fixes the CSV import tool translations

ClemRz:


ClemRz:
I added SCNAT logo. Solves issue #1074

PR #1131, opened 2025-12-24
fix(map): remove geolocation blocking on map load and pan to user location when accepted

ClemRz:
# Fix map loading behavior with geolocation

## 🤔 What

- Changed map loading behavior to no longer block on geolocation permission
- Maps now load immediately with default coordinates or existing data
- When user accepts location sharing, maps pan to user location (only if no other data exists)
- Removed unused `hasError` property from `useGeolocation` hook
- Simplified zoom level logic across all map components

## 🤷♂️ Why

Previously, all maps (Map.jsx, MapMarkerSelector.jsx, PolygonMap.jsx) would show a loading spinner until the user accepted or denied location sharing. This created a poor user experience where:
- Users had to wait before seeing any map
- The map wouldn't load if geolocation was blocked or unavailable
- Users couldn't interact with the map while the permission dialog was open

### Core Changes

1. **useGeolocation hook** (`hooks/useGeolocation.js`)
   - Changed from `isReady` to `hasLocation` flag
   - No longer blocks - returns immediately with default coordinates
   - Sets `hasLocation: true` only when user accepts and GPS coordinates are acquired
   - Removed unused `hasError` property
   - Removed optional error callback (not needed)

2. **Map.jsx** (main map page)
   - Loads immediately with default coordinates (0,0) at zoom 2
   - When user accepts location sharing AND no target coordinates exist, pans to user location with zoom 13
   - Treats URL target `0,0,2` as "no real target" to allow panning to user location
   - Uses `mapRef` to directly call `map.setView()` for programmatic panning

3. **MapMarkerSelector.jsx** (entry/cave forms)
   - Loads immediately with form coordinates or default coordinates
   - Pans to user location only when accepted AND no valid form data exists
   - Simplified zoom level logic using logical OR

4. **PolygonMap.jsx** (massif forms)
   - Loads immediately with polygon centroid or default coordinates
   - Pans to user location only when accepted AND no polygon data exists
   - Simplified to use `initialCenter` instead of stateful `center`
   - Removed unnecessary `setCenter` calls
   - Simplified zoom level logic using logical OR

5. **CustomMapContainer** (`common/Maps/common/MapContainer.jsx`)
   - Added `mapRef` prop support to expose Leaflet map instance
   - Kept `forceCentering` and `Centerer` component for Region/Country components

## 🔗 Related API PR

None - frontend only changes.

### Map.jsx (Main map at `/ui/map`)
- [ ] Load `/ui/map` without location permission → map shows at (0,0) zoom 2
- [ ] Accept location sharing → map pans to user location with zoom 13, URL updates
- [ ] Deny location sharing → map stays at (0,0) zoom 2
- [ ] Load `/ui/map/48.8566,2.3522,13` → map loads at those coordinates, ignores user location
- [ ] Manual panning/zooming → smooth interaction, URL updates correctly
- [ ] Refresh page after panning → map loads at URL coordinates

### MapMarkerSelector.jsx (Entry/cave forms)
- [ ] Load form with empty lat/lng → map shows at default coordinates zoom 2
- [ ] Load form with existing lat/lng → map centers on those coordinates zoom 13
- [ ] Accept location sharing with empty form → map pans to user location zoom 13
- [ ] Accept location sharing with filled form → map stays on form coordinates
- [ ] Drag map → form lat/lng fields update correctly
- [ ] Type in lat/lng fields → map updates position

### PolygonMap.jsx (Massif forms)
- [ ] Load form with existing polygon → map fits bounds to polygon
- [ ] Load form without polygon → map shows at default coordinates zoom 2
- [ ] Accept location sharing without polygon → map pans to user location zoom 13
- [ ] Accept location sharing with polygon → map stays on polygon bounds
- [ ] Draw/edit polygon → smooth interaction

### Edge Cases
- [ ] Browser without geolocation support → maps load at defaults, no errors

### DIFF
--- a/packages/web-app/src/pages/Map.jsx
+++ b/packages/web-app/src/pages/Map.jsx
@@ -63,7 +63,7 @@ const Map = () => {
     });
     navigate(newPath, { replace: true });
     dispatch(changeLocation(center));
-    dispatch(changeZoom(zoom));
+    dispatch(changeZoom(newZoom));
 
     const criteria = {
       /* eslint-disable no-underscore-dangle */
@@ -93,6 +93,10 @@ const Map = () => {
 
   useEffect(() => {
     dispatch(fetchProjections());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
     const target = decodeMapTarget(params.target);
     if (target) {
       dispatch(changeLocation({ lat: target.lat, lng: target.lng }));
--- a/packages/web-app/src/pages/Map.jsx
+++ b/packages/web-app/src/pages/Map.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, Suspense } from 'react';
+import React, { useEffect, useRef, Suspense } from 'react';
 import { includes } from 'ramda';
 import { useNavigate, generatePath, useParams } from 'react-router-dom';
 import { useDispatch, useSelector } from 'react-redux';
@@ -16,7 +16,7 @@ import {
 import { fetchProjections } from '../actions/Projections';
 import useGeolocation from '../hooks/useGeolocation';
 import 'leaflet/dist/leaflet.css';
-import { defaultZoom, focusZoom } from '../conf/config';
+import { defaultZoom, defaultCoord, focusZoom } from '../conf/config';
 
 const MapClusters = React.lazy(
   () => import('../components/common/Maps/MapClusters')
@@ -40,7 +40,8 @@ const Map = () => {
   const navigate = useNavigate();
   const dispatch = useDispatch();
   const params = useParams();
-  const { location: geoLocation, isReady: isGeoReady, hasError: geoHasError } = useGeolocation();
+  const { location: geoLocation, hasLocation } = useGeolocation();
+  const mapRef = useRef(null);
   const {
     location,
     zoom,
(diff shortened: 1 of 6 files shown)


packages/web-app/src/pages/Map.jsx
```diff
@@ -92,26 +93,24 @@ const Map = () => {
 
   useEffect(() => {
     dispatch(fetchProjections());
```
vmarseguerra, line 95: Projections are now fetched every time the map is moved or zoomed

packages/web-app/src/pages/Map.jsx
```diff
@@ -126,6 +125,7 @@ const Map = () => {
         onUpdate={handleUpdate}
         isSideMenuOpen={open}
         projectionsList={projections}
+        mapRef={mapRef}
```
vmarseguerra, line 132: Is this required? Why not using `changeLocation()` and `changeZoom()` like elsewhere in the file?

vmarseguerra (changes requested):

packages/web-app/src/pages/Map.jsx
ClemRz: Good catch!

packages/web-app/src/pages/Map.jsx
ClemRz: The `mapRef` is used in the third `useEffect` (lines 113-119) to directly call `mapRef.current.setView()` when the user's geolocation becomes available.

```jsx
if (hasLocation && (!params.target || isDefaultTarget) && mapRef.current) {
  mapRef.current.setView([geoLocation.lat, geoLocation.lng], focusZoom);
}
```

**Pros:**
- Direct map manipulation bypasses the Redux state update cycle, providing immediate visual feedback
- Avoids triggering the `handleUpdate` callback, which would fetch data and update the URL unnecessarily
- The `setView` method is synchronous and guaranteed to execute immediately
- Prevents potential infinite loops that could occur if Redux state changes triggered re-renders during geolocation updates

**Cons:**
- Creates two different patterns for map updates in the same file (direct manipulation vs. Redux actions)
- Breaks the unidirectional data flow principle
- The map state (what's displayed) can temporarily diverge from Redux state
- Makes the component harder to test since it relies on imperative API calls

## Alternative Approach (Using changeLocation/changeZoom)

**Pros:**
- Maintains consistency with the rest of the file
- Keeps all state changes flowing through Redux
- More predictable and easier to debug
- Better aligns with React's declarative paradigm

**Cons:**
- Would trigger `handleUpdate` via the `onUpdate` prop, causing unnecessary data fetches
- Would update the URL, which may not be desired for automatic geolocation
- Adds extra re-render cycles
- The second `useEffect` (lines 103-111) would need modification to avoid conflicts

In conclusion, I think that the `mapRef` approach is justified here because this specific use case requires direct map manipulation without triggering the full update cycle. The geolocation feature needs to move the map view without fetching new data or updating the URL, which wouldn't be possible using `changeLocation()` and `changeZoom()` alone.

bsoufflet (approved):

PR #1132, opened 2025-12-24
feat(search): persists EntityTable user preferences in localStorage

ClemRz:
# Persist EntityTable user preferences (rows per page and column visibility)

## 🤔 What

This PR adds persistence for user preferences in the EntityTable component:

- Persist the selected rows per page across sessions using localStorage
- Persist column visibility preferences per entity type
- Apply stored preferences on initial load and when performing advanced searches
- Add "Change columns" translation key in English, Spanish, and French

## 🤷♂️ Why

Users had to reconfigure their table preferences (rows per page and visible columns) every time they refreshed the page or performed a new search. This created a poor user experience as their preferences were not remembered between sessions.

## 🔍 How

**EntityTable component changes:**
- Extracted `getStoredRowsPerPage()` as an exported utility function to retrieve the stored rows per page preference from localStorage
- Created `applyColumnVisibility()` helper function to restore column visibility from localStorage
- Modified state initialization to use lazy initialization with stored values
- Added `useEffect` hooks to persist column visibility changes to localStorage
- Added `useEffect` to trigger page change on mount if stored rowsPerPage differs from default
- Stored the rows per page value in localStorage when changed via `handleChangeRowsPerPage`
- Moved `DEFAULT_PAGE_SIZE_OPTIONS` to config.js for reusability

**Advanced Search components:**
- Imported and used `getStoredRowsPerPage()` in all search components (DocumentSearch, EntrancesSearch, MassifsSearch, OrganizationsSearch, PersonSearch)
- Pass the stored page size as the `size` parameter when fetching search results to ensure initial results match user preferences

**Storage keys:**
- `entityTable_rowsPerPage`: stores the selected rows per page (shared across all entity types)
- `entityTable_${entityType}_columns`: stores column visibility per entity type (e.g., `entityTable_entrances_columns`)

## 🧪 Testing

1. Open any advanced search page (e.g., Entrances, Documents, Organizations)
2. Change the rows per page from the default (e.g., from 20 to 100)
3. Refresh the page - verify the rows per page preference is maintained
4. Click the column visibility icon and hide/show some columns
5. Refresh the page - verify the column visibility preferences are maintained
6. Switch to a different entity type - verify each entity type maintains its own column visibility preferences
7. Perform a new search - verify the stored rows per page is used for the initial query

### DIFF
--- a/packages/web-app/src/components/common/EntityTable/EntityTable.jsx
+++ b/packages/web-app/src/components/common/EntityTable/EntityTable.jsx
@@ -33,7 +33,30 @@ import { LoadingTableHead, LoadingTableBodyInner } from './LoadingTable';
 import Alert from '../Alert';
 import Translate from '../Translate';
 
+const DEFAULT_PAGE_SIZE_OPTIONS = [20, 100, 200];
 const MAX_DOCUMENTS_TO_EXPORT_IN_CSV = 10000; // This limit is also enforced server side
+
+const applyColumnVisibility = (columns, storedVisibility) => {
+  try {
+    const visibleColumns = JSON.parse(storedVisibility);
+    return columns.map(col => {
+      const newCol = [...col];
+      newCol[0] = visibleColumns.includes(col[1]);
+      return newCol;
+    });
+  } catch (e) {
+    return columns;
+  }
+};
+
+export const getStoredRowsPerPage = (pageSizeOptions = DEFAULT_PAGE_SIZE_OPTIONS) => {
+  const stored = localStorage.getItem('entityTable_rowsPerPage');
+  if (stored) {
+    const value = parseInt(stored, 10);
+    if (pageSizeOptions.includes(value)) return value;
+  }
+  return pageSizeOptions[0];
+};
 
 const StyledTablePagination = styled(TablePagination)`
   p {
@@ -105,7 +128,7 @@ EntityTableHead.propTypes = {
   rowCount: PropTypes.number.isRequired
 };
 
-const VisibleColumnsMenu = ({ columns, setColumns }) => {
+const VisibleColumnsMenu = ({ columns, setColumns, entityType }) => {
   const { formatMessage } = useIntl();
   const [anchorEl, setAnchorEl] = useState(null);
   return (
@@ -128,7 +151,13 @@ const VisibleColumnsMenu = ({ columns, setColumns }) => {
             key={column[1]}
             onClick={() => {
               column[0] = !column[0]; // eslint-disable-line no-param-reassign
-              setColumns([...columns]);
+              const newColumns = [...columns];
+              setColumns(newColumns);
+              if (entityType) {
+                const storageKey = `entityTable_${entityType}_columns`;
+                const visibleColumns = newColumns.filter(col => col[0]).map(col => col[1]);
+                localStorage.setItem(storageKey, JSON.stringify(visibleColumns));
+              }
             }}>
             <Checkbox
               size="small"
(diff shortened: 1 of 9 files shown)


urien (approved):
Merci,
Cela fonctionne pour toutes les entités : entrées, massifs, organisations, documents personnes

packages/web-app/src/conf/config.js
```diff
@@ -91,3 +91,5 @@ export const NOMINATIM_API_URL = 'https://nominatim.openstreetmap.org/search';
 
 export const AUTOCOMPLETE_DEBOUNCE_DELAY = 500; // in milliseconds
 export const AUTOCOMPLETE_MIN_CHARACTERS = 2;
+
+export const DEFAULT_PAGE_SIZE_OPTIONS = [20, 100, 200];
```
vmarseguerra: As this is not a global option may it can be put inside the `EntityTable.jsx‎` file alongside `MAX_DOCUMENTS_TO_EXPORT_IN_CSV`

packages/web-app/src/components/common/EntityTable/EntityTable.jsx
```diff
@@ -253,12 +294,27 @@ const EntityTable = ({
...
   }, [entityType]);
 
+  useEffect(() => {
+    if (!entityType) return;
+    const storageKey = `entityTable_${entityType}_columns`;
+    const visibility = entityColumns.reduce((acc, col) => {
+      acc[col[1]] = col[0];
+      return acc;
+    }, {});
+    localStorage.setItem(storageKey, JSON.stringify(visibility));
```
vmarseguerra: May be [too avoid using](https://react.dev/learn/you-might-not-need-an-effect) `useEffect` this can be put inside the `VisibleColumnsMenu` element

packages/web-app/src/components/common/EntityTable/EntityTable.jsx
```diff
@@ -167,12 +192,20 @@ const EntityTable = ({
...
+    getStoredRowsPerPage(pageSizeOptions)
+  );
   const [order, setOrder] = useState('');
   const [orderBy, setOrderBy] = useState('');
   const [selected, setSelected] = useState([]);
   const entityConfig = entitiesConfig[entityType ?? 'placeholder'];
-  const [entityColumns, setEntityColumns] = useState(entityConfig.columns);
+  const [entityColumns, setEntityColumns] = useState(() => {
+    const storageKey = `entityTable_${entityType}_columns`;
+    const stored = localStorage.getItem(storageKey);
```
vmarseguerra, line 209: This can be factorized inside `applyColumnVisibility()`

packages/web-app/src/components/common/EntityTable/EntityTable.jsx
```diff
@@ -32,9 +32,34 @@ import entitiesConfig from './entitiesConfig';
 import { LoadingTableHead, LoadingTableBodyInner } from './LoadingTable';
 import Alert from '../Alert';
 import Translate from '../Translate';
+import { DEFAULT_PAGE_SIZE_OPTIONS } from '../../../conf/config';
 
 const MAX_DOCUMENTS_TO_EXPORT_IN_CSV = 10000; // This limit is also enforced server side
 
+const applyColumnVisibility = (columns, storedVisibility) => {
```
vmarseguerra, line 39: Instead of copying each element manually and returning a new array, modifying only the `[0]` element will be more futureproof (in case elements are added/removed).
Also storing only the visible column in a string array could be simpler than in an object

vmarseguerra (changes requested):

bsoufflet (approved):

PR #1170, opened 2026-02-13
fix(advancedSearch): use nested Typesense field paths for CSV export columns

ClemRz:
# Fix CSV export columns for nested Typesense fields

## 🤔 What

- Fix the document CSV export sending parent object keys (`authors`, `iso3166`) instead of Typesense nested field paths (`authors.nickname`, `iso3166.iso`).
- This caused the backend to receive raw nested objects and serialize them as `[object Object]` in the CSV output.

## 🤷‍♂️ Why

The export endpoint expects Typesense nested field paths to extract scalar values. Sending the parent key returns the full nested object, which can't be meaningfully serialized to CSV.

## 🔍 How

The column key in `entitiesConfig.jsx` is used both for table cell rendering (via `getObjectPath`) and for the export payload. Changing it directly would break the table since fields like `doc.authors` are arrays, not objects with a `.nickname` property.

To solve this without breaking table rendering:

1. Added an optional 6th element (`exportKey`) to the column tuples in `entitiesConfig.jsx` for `authors` and `iso3166`.
2. Updated the export logic in `EntityTable.jsx` to prefer `e[5]` (export key) over `e[1]` (display key) when building the `columns` array: `c.map(e => e[5] || e[1])`.

This is backward-compatible — columns without a 6th element behave exactly as before.

## 🔗 Related API PR

N/A — the backend already has a `toPrimitive` safety net in `advanced-search-export.js`, but this fix ensures the frontend sends correct field paths for explicit, predictable results.

## 🧪 Testing

1. Go to Advanced Search, search for documents.
2. Click "Export to CSV".
3. Verify the CSV contains actual author nicknames and ISO codes, not `[object Object]`.
4. Verify the table still renders authors and country/region columns correctly.

## 📸 Previews

N/A — no visual changes; the table renders identically. The fix only affects the export payload.


ClemRz:
I'm gonna merge right away. @urien, @bsoufflet & @vmarseguerra feel free to comment even after the fact if there is anything concerning to you :)

PR #1172, opened 2026-02-13
feat: preview images

Paul-AUB:
# Afficher une preview des images directement sur la page des Entrances

## 🤔 What

Implementation of : https://github.com/GrottoCenter/grottocenter-front/issues/1071
closes #1071 

What has been implemented:

- Thumbnail preview of all images (support various images files types: `['.jpg', '.jpeg', '.png', '.gif']`)
- Image lightbox with zoom + move capabilites when clicking on the thumbnail
- lightbox also display the description as a caption
- Explicit icon and button to download the image
- Also display the thumbnail/lightbox on personnal page
- Updated translations for all new texts
- Bonus: added an AGENT.md file to support AI development best practices

## 🤷‍♂️ Why

See : https://github.com/GrottoCenter/grottocenter-front/issues/1071

## 🔍 How

I have to admit that I used a bit of AI...

## 🔗 Related API PR

N/a

## 🧪 Testing

- Tested visually for images on entrances
- Tested visually for images on personal profile
- Tested visually that doesn't break other documents

## 📸 Previews

image

image

image

PR #1174, opened 2026-02-15
feat: preview images

### LINKED ISSUE #1071: Afficher une preview des images directement sur la page des Entrances
La gestion documentaire sur Grottocenter est très précise et complète, mais l'expérience utilisateur sur la page des Entrances n'est pas très agréable. Exemple, ici il y a la photo de l'entrée d'une cavité :

Image

La photo n'est pas affichée directement dans la page, et il faut cliquer dessus pour voir à quoi ça ressemble. Paradoxalement, les infos prennent beaucoup de place à l'écran, et il y a 2 liens cliquables (un vers la page de l'image, et un vers l'image en elle-même).

- Avoir un moyen d'afficher une preview des images (photos, topos, etc.) directement sur la page pour voir à quoi ça ressemble (quitte à avoir un lien pour ouvrir la page avec les détails ensuite en cliquant dessus.
- Faire en sorte que les métadonnées prennent moins de place à l'écran 

Si on compare à Camp to Camp, la visualisation est plus facile et naturelle, et prend paradoxalement moins de place :

Image

Idem sur Visorando:

Image

On a bien une preview des images, et ensuite sur un clic une page de détails:

Image

**Describe alternatives you've considered**
Je suis conscient que la solution proposée s'applique essentiellement aux images, et pas forcément à tous les documents (genre pdf, encore que...). Peut être faudrait-il appliquer un fonctionnement différent suivant le type de document ? Ou le type de fichier ?

**Additional context**
Grottocenter est vraiment génial de part la quantité d'information qu'il contient, mais en discutant avec les spéléo autour de moi, il semblerait que le principal frein à son adoption est son UX. Une amélioration de cette dernière serait un boost non négligeable pour l'app !

Paul-AUB:
Direct copy from #1172 (no more from a fork).
Most of review change requests have been addressed in the last commit.


Paul-AUB:
NB: it seems the check failed was related to the PR from the fork, not this one. This branch has successfully been deployed in the staging environment, I let you review the functionnality.

urien:
C'est une très belle fonctionnalité.
merci Paul. Je vais proposer aux membres de l'association de donner leur avis
Je laisse un développeur regarder ton code

packages/web-app/public/lang/ar.json
```diff
@@ -351,5 +351,9 @@
   "Zambia": "زامبيا",
   "Zimbabwe": "زيمبابوي",
   "oaiTooltip": "خادم OAI-PMH",
-  "z3950Tooltip": "خادم Z39.50"
+  "z3950Tooltip": "خادم Z39.50",
+  "Download image": "تنزيل الصورة",
+  "Image {current} of {total}": "صورة {current} من {total}",
+  "Next image": "الصورة التالية",
+  "Previous image": "الصورة السابقة"
```
ClemRz: Please sort all the translation files with `yarn translations:sort <path-to-json-file>`

AGENTS.md
ClemRz, line 1: Please don't include this file in your PR

ClemRz (changes requested):
There are 2 special entrances in GrottoCenter that are worth looking at:
- [Mine de Charbon (Grotte de la)](https://orange-rock-0d4f87503-1174.westeurope.azurestaticapps.net/ui/entrances/25759) - this is the entrance with most documents
- [Cave 18 A (Fig Tree) Orkerenket Lorngaboli](https://orange-rock-0d4f87503-1174.westeurope.azurestaticapps.net/ui/entrances/93177) - this is the entrance with most images

In the second one I feel like the images display could be more compact by limiting the text length in the buttons (use ellipsis "..."):

Image


I think that if the text is more than 21 characters long then it should be cropped at 18 chars + ellipsis (3 chars):

Screenshot 2026-02-16 at 11 08 32 a m

Paul-AUB:
All fixed ! I created a separated commit rather than fixup + rebase so that it should be easily reviewable. I enlarged a bit the size of the thumbnail, still readable in mobiles :

image

image

image

.gitignore
```diff
@@ -68,4 +68,7 @@ typings/
 .vscode
 .idea
 *.iml
-.amazonq
\ No newline at end of file
+.amazonq
```
ClemRz: This one is also for AI ^ maybe you could move it in the # AI section you created?

ClemRz (changes requested):
This is beautiful!

Quick question: I see that one can navigate through the images of the same document when clicking on it. When there are several one-imaged documents I cannot do the same. Would it be possible to allow navigation across all documents at once?

Some screenshots from the mentioned entrances:

When a document has several pictures I can navigate (left-right):
Image

When a document has only one picture I cannot, even if the entrance has more documents with more pictures:
Image

.gitignore
Paul-AUB: Yep

Paul-AUB:
Well, actually not *so* difficult^^

I just pushed the changes to do so, e.g. [here](https://orange-rock-0d4f87503-1174.westeurope.azurestaticapps.net/ui/entrances/25759):

image

ClemRz (approved):

ClemRz:
Can you squash all your commits into a single one please?

Paul-AUB:
Done !

ClemRz:
Thank you @Paul-AUB ! Lets wait on @vmarseguerra feedback :)

ClemRz (changes requested):
## Code Review — PR #1174: feat: preview images

Nice feature — image thumbnails with a lightbox viewer is a solid UX improvement for the documents list. Here's what I found:

---

### 1. Full-size images loaded as thumbnails (performance)

`ImageThumbnail` loads the full `completePath` URL directly into a 240×180 card. If users upload high-res cave photos (5+ MB), every thumbnail will download the full image just to display it at 240px wide. This could make document lists with several images very slow to load.

The `loading="lazy"` attribute helps with offscreen images, but for visible ones it's still full-resolution. Not a blocker for this PR, but it would be worth opening an issue on the API side to support thumbnail generation — e.g. a small microservice that automatically creates and stores a resized thumbnail when a picture gets uploaded. The frontend could then request a `thumbnailPath` instead of `completePath` for the grid view.

---

### 2. Dual lightbox instances — potential confusion

`Files.jsx` creates its own `ImageLightbox` as a fallback when `onImageClick` is not provided, while `DocumentsList.jsx` also renders one at the list level. This means there are two lightbox management paths:

- Parent-managed (via `onImageClick` + `imageIndexOffset`) for cross-document navigation
- Self-managed (inside `Files.jsx`) as a fallback

This works, but it's a bit subtle. If `Files` is ever used outside of `DocumentsList` without passing `onImageClick`, the fallback lightbox won't have cross-document navigation. A comment in `Files.jsx` explaining when each path is used would help future maintainers.

---

### 3. `contentRef` callback doesn't clean up the wheel listener

In `ImageLightbox.jsx`:

```js
const contentRef = useCallback(
  node => {
    if (node) {
      node.addEventListener('wheel', handleWheel, { passive: false });
    }
  },
  [handleWheel]
);
```

Callback refs don't automatically remove event listeners when the node unmounts or the ref changes. If the component re-renders and the ref callback fires again with a new node, the old listener on the previous node is never removed. Consider tracking the node in a regular ref and cleaning up in a `useEffect`:

```js
const contentRef = useRef(null);

useEffect(() => {
  const node = contentRef.current;
  if (node) {
    node.addEventListener('wheel', handleWheel, { passive: false });
    return () => node.removeEventListener('wheel', handleWheel);
  }
  return undefined;
}, [handleWheel]);
```

---

### 4. Zoom/pan state not reset on close

The `useEffect` that resets zoom and position depends on `[currentIndex, open]`:

```js
useEffect(() => {
  setZoom(1);
  setPosition({ x: 0, y: 0 });
}, [currentIndex, open]);
```

This works when `open` toggles, but if the user zooms in, closes, and reopens on the same image (same `currentIndex` and `open` goes false→true), the effect does fire. That's fine. However, if `initialIndex` changes while `open` stays `true` (e.g., clicking a different thumbnail while lightbox is already open), `currentIndex` updates via the other effect but the zoom reset races with it. Minor edge case, but worth being aware of.

---

### 5. `height: calc(100vh - 190px)` in lightbox — should use `dvh`?

```js
const LightboxContent = styled(Box)`
  height: calc(100vh - 190px);
```

Given that PR #1177 is systematically migrating `vh` to `dvh` across the codebase for mobile viewport correctness, this new code should probably follow the same pattern with a `vh` fallback + `dvh` override.

---

### 6. `.gitignore` change — `AGENTS.md`

The PR adds `AGENTS.md` to `.gitignore`. This seems unrelated to the image preview feature — probably a personal dev config file. Not a blocker, but might be cleaner as a separate commit or PR to keep the diff focused.

---

### 7. `isImageFile` — `.svg` might not be desirable as a thumbnail

The image extension list includes `.svg`:

```js
const imageExtensions = ['.jpg', '.jpeg', '.png', '.gif', '.webp', '.svg'];
```

SVGs in a document context could be technical diagrams, cave survey exports, or vector maps that won't render well as 240×180 thumbnails with `object-fit: cover`. Might be worth excluding `.svg` or handling it differently.

---

### Minor / positive notes

- The skeleton loading state in `ImageThumbnail` with opacity transition is a nice touch.
- Keyboard navigation (arrow keys) in the lightbox is good for accessibility.
- The `imageOffsets` computation in `DocumentsList` to map per-document image indices to global indices is clean.
- Error fallback with the `Description` icon is a reasonable degradation.
- Proper use of `useCallback` for stable handler references.

Overall solid work. The main items to address are the wheel listener cleanup, considering `dvh` for the lightbox height, and documenting the dual-lightbox pattern.

Paul-AUB:
I fixed the wheel issue (I had to adapt a bit the suggestion), the `dvh`, documented the lightbox in `Files.jsx`. I squashed everything in the same commit.

I also opened https://github.com/GrottoCenter/grottocenter-api/issues/1468 to track the optimization feature.

Should be all good, thx for the review !

ClemRz:
Thank you @Paul-AUB !! This is awesome!

ClemRz (approved):

ClemRz:
I feel confident in merging this to the trunk. @bsoufflet @vmarseguerra please have a code review anyway and feel free to let us know if you detect anything.

PR #1176, opened 2026-02-16
feat(RelevanceOrdering): add relevance-based ordering and reorder controls for entity lists

ClemRz:
**THIS REQUIRES API PR [#1466](https://github.com/GrottoCenter/grottocenter-api/pull/1466) TO BE MERGED FIRST**

# feat(RelevanceOrdering): add relevance-based ordering and reorder controls for entity lists

## 🤔 What

- Adds relevance-based display ordering for 5 entity types: Locations, Descriptions, Comments, Riggings, and Histories
- Adds up/down arrow reorder controls inside ActionButtons for authenticated users
- Adds success snackbar with undo link after each successful move
- Closes #314

## 🤷‍♂️ Why

Entity lists (locations, descriptions, comments, riggings, histories) had no user-facing ordering mechanism. The API already supports a `relevance` integer field and `PATCH` endpoints to swap adjacent items. This PR adds the frontend counterpart so cavers can prioritize content within lists.

## 🔍 How

- `sortByRelevance` helper sorts entity arrays by `relevance ASC` (nulls last, stable sort)
- `createMoveRelevanceAction` factory generates Redux thunks for each entity type following the `MOVE_<ENTITY>_RELEVANCE` / `_SUCCESS` / `_FAILURE` pattern
- `swapRelevance` reducer helper uses nested `arrFindReplaceOrAdd` to replace both moved and swapped entities in the store
- EntranceReducer, CaveReducer, and MassifReducer handle the success actions for their respective entity lists
- ActionButtons extended with optional `onMoveUp`/`onMoveDown`/`isFirst`/`isLast`/`isMoveLoading` props — arrows are hidden at boundaries, spinner shown during move
- `useMoveRelevanceWithUndo` hook encapsulates dispatch + notistack snackbar with undo link (6s auto-dismiss, opposite direction on undo click, 3s confirmation on undo success)
- All 5 entity list components (Locations, Descriptions, Comments, Riggings, Histories) updated to sort by relevance and use the hook
- API route helpers added to `apiRoutes.js` for all 5 entity types
- Translation keys added to en/fr/es

## 🔗 Related API PR

The API already supports the `PATCH /api/v1/<entityType>/:id/move-relevance` endpoints — no API PR needed.

## 🧪 Testing

- 30 tests across 6 suites (all passing):
  - Property tests: URL builder correctness, relevance sort invariant, reducer swap correctness, reorder controls visibility, undo direction inversion, move-then-undo round trip
[...]

Issue #329, opened 2022-02-28
[PICTO] Review, uniformize & refactor


urien:
**Is your feature request related to a problem? Please describe.**
Les pictos n'ont pas tous exactement la même taille, l'ombrage n'est pas exactement le même, il y a beaucoup de pictos qui sont conservés et qui ne serviront jamais, il manque peut-être des pictos. 
De plus ils y a des pictos dans des dossiers différents

**Describe the solution you'd like**
Je propose que  l'ensemble des pictos soient repris afin qu'ils aient les mêmes caractéristiques. Eventuellement il faudrait les placer tous dans le même dossier, ce qui nécessite de reprendre l'application là où ils sont utilisés

bsoufflet:
Comment from @BTHOM regarding Picto to improve : 

> le logo qui accompagne l'altitude de l'entrée est un peu spartiate et ambigu. On pourrait compléter sa flèche verticale par un triangle blanc sur fond marron, suggérant un pic montagneux, donc une dénivelée extérieure. Et son style serait plus en harmonie avec celui des pictogrammes "Profondeur" et "Développement".

Paul-AUB:
> Comment from [@BTHOM](https://github.com/BTHOM) regarding Picto to improve :
> 
> > le logo qui accompagne l'altitude de l'entrée est un peu spartiate et ambigu. On pourrait compléter sa flèche verticale par un triangle blanc sur fond marron, suggérant un pic montagneux, donc une dénivelée extérieure. Et son style serait plus en harmonie avec celui des pictogrammes "Profondeur" et "Développement".

Proposal for `altitude`, trying to keep consistence with other icons (here's the real svg):

![Image](https://github.com/user-attachments/assets/05e2c297-3903-4b01-8765-b81bcb64c07c)

<img width="605" height="53" alt="Image" src="https://github.com/user-attachments/assets/5d9c011b-c494-4151-b112-aa65f46eecea" />

urien:
Cela me convient, bonne idée

Paul-AUB:
# Architecture plan : Image & Logo Management Modernization

## Context

The project has accumulated technical debt around visual asset management:
- 3 coexisting icon systems (obsolete `gc-icon` font, PNG icons8 set, SVG iconsV3 set)
- Orphaned and backup files polluting the production build
- No clear separation between `public/` (static, stable URLs) and `src/assets/` (bundled, hashed)

[...]

PR #1177, opened 2026-02-16
feat: search on map

### LINKED ISSUE #1173: [MAP] Rechercher des entrées et réseaux sur la carte
Actuellement, il n'est pas possible de trouver des entrées ou des réseaux directement sur la carte. 
Par exemple: 

Image
ne permet pas de trouver "Chique (Scialet de la)", alors qu'une recherche "normale" donne bien :

Image

[Grottomap](https://grottomap.org/) permet bien de faire cette recherche :

Image

Le but de cette issue serait d'implémenter la recherche de la même façon, avec l'icon "entrée" lorsque la recherche trouve une entrée. On pourrait imaginer qqch comme ça :

Image

**Describe alternatives you've considered**
Utiliser grottomap à la place 😂, mais c'est un peu dommage, y'a d'autres trucs très cool ici !

**Additional context**
N/a

Paul-AUB:
# Search for entities on map

## 🤔 What

See https://github.com/GrottoCenter/grottocenter-front/issues/1173

- Feature to allow to search anything in the map, the results would be in that order:
  - Entrances (5 results max)
  - Network (3 results max)
  - Organizations (3 results max)
  - Other locations (5 results max), that's the legacy behavior
- If an entity has no coordinates, do not display it in the result (as they can't be found in the map)
- For networks, I used the first entry coordinates (and considered a network = a cave with more than one entry ; definition of network vs cave is a bit... confused it seems - at least for me) 
- Fix a bug in mobile where the result list was messy with touch gesture
- Fix a bug where the map size didn't fit the screen (see below)
- Fix overflow in research bar with other button (see below
- Fix double tooltip + hoover message in mobile (bad management of hoovering with touch)
- Fix the size of the computed sceen visible was faulty (usage of `vh` instead of `dvh`)
- Improve tooltip density on map

## 🤷‍♂️ Why

Because it's a caving map, so i want to search for caves stuff xD

## 🧪 Testing

Manual test on firefox and mobile (simulated and real one).
If there's some edges cases known, tel me I can test them.

### Main feature

image

image

image

Also tested on mobile:

image

### Improvements

Before:
image

After:

image

### Bugs

image

image

image


Paul-AUB:
For searching purpose, I used basic `matchAllFields: false`, so I return results even if the keyword is in the description, etc.
E.g. 

image

Spot the organization because:

image

I don't think it's a "bad" behavior because it ensure to find what you look for (even if you may fin *too much stuff* ; but you just have to specify more your search).

Paul-AUB:
NB: I let you deploy when you want / when staging env is available ;)

ClemRz:
I freed up one staging env (the #1176 one) to give this priority.

ClemRz:
When I search for an organization and click on one, the map pans to the right place but the balloon does not open and the marker is not shown. Maybe we should consider checking automatically the organizations marker (top-right menu on the map) when an org. is selected?

ClemRz:
Unrelated bug: the tip of the balloon's arrow is anchored to the bottom of the marker which has the undesirable effect of hiding it:

Screenshot 2026-02-17 at 12 04 14 p m

ClemRz:
When I search for "Jean-Bernard" and select the network, the balloon opens with the information of the "T04 (T4)" instead:

Screenshot 2026-02-17 at 12 06 43 p m

Screenshot 2026-02-17 at 12 07 10 p m

urien:
C'est un super service. Bravo Paul.
Effectivement il faut trouver un moyen pour que le picto apparaisse quand on choisi une organisation. La proposition de Clément me convient

ClemRz (commented):
.

ClemRz (changes requested):
## Code Review — PR #1177: feat: search on map

Nice work on this one. The core feature is solid and well-structured. Here are some things I spotted:

---

### 1. N+1 API calls for networks (performance concern)

The network search fetches a list of caves, then fires an individual `GET /caves/:id` for each result to check entrance count and get coordinates. With `size: 3`, that's up to 4 HTTP requests just for networks (1 search + 3 detail fetches). This is a waterfall pattern that adds latency.

```js
Promise.all(
  caves.map(cave =>
    fetch(`${getCaveUrl}${cave.id}`)
      .then(res => res.json())
      .then(detail => {
        const entrances = detail.entrances || [];
        if (entrances.length < 2) return null;
        // ...
      })
  )
)
```

With `size: 3` it's tolerable, but if the limit ever increases this will scale poorly. Worth a comment explaining the tradeoff, or ideally the backend could return entrance coordinates directly in the search results.

---

### 2. Empty CSS rule block

```css
.MuiListItemText-primary {
}
```

This is a no-op styled rule — should be removed.

---

### 3. Race condition on rapid typing

The debounced search fires 4 parallel API calls. If the user types quickly, a previous batch of requests could resolve after a newer one, showing stale results. The timer is cleared on re-render, but the `fetch` calls from the previous timer invocation are not aborted.

Consider using an `AbortController`:

```js
useEffect(() => {
  if (query.length < AUTOCOMPLETE_MIN_CHARACTERS) { /* ... */ }

  const controller = new AbortController();
  setLoading(true);

  const timer = setTimeout(async () => {
    try {
      const [locationData, ...rest] = await Promise.all([
        fetch(url, { signal: controller.signal }).then(r => r.json()),
        // ...
      ]);
    } catch (error) {
      if (error.name !== 'AbortError') console.error(error);
    }
  }, AUTOCOMPLETE_DEBOUNCE_DELAY);

  return () => {
    clearTimeout(timer);
    controller.abort();
  };
}, [query, locale, map]);
```

---

### 4. Marker matching by coordinate proximity is fragile

```js
const isTargetMarker = layer =>
  layer instanceof L.Marker &&
  Math.abs(layer.getLatLng().lat - lat) < 0.001 &&
  Math.abs(layer.getLatLng().lng - lng) < 0.001;
```

A tolerance of 0.001° (~111m) could match the wrong marker in dense cave areas. If the API returns a unique entrance ID, matching on a marker data attribute would be more reliable.

---

### 5. `width: calc(100%)` is redundant

In `MapContainer.jsx`:
```js
${theme.breakpoints.up('md')} {
  width: calc(100%);  // This is just `width: 100%`
```

Looks like a leftover from editing.

---

### 6. `50svh` unit in OrganizationLogo is unusual

```jsx
<FormControl component="fieldset" style={{ width: '50svh' }}>
```

The original was `50vh`. Using a viewport-height unit for a **width** is semantically odd. Was this meant to be `50svw` (small viewport width), or just left as `50vh`? Worth double-checking.

---

### 7. Dead translation key `Search location...`

The placeholder changed from `Search location...` to `Search on map...`, but the old key still exists in `en.json`. If nothing else references it, it should be cleaned up.

---

### 8. Duplicated `club.svg` icon

The new `iconsV3/club.svg` is byte-for-byte identical to the existing `images/club.svg`. Was there a reason to duplicate it rather than reusing the existing one (e.g. adjusting the `CustomIcon` path or symlinking)?

---

### Minor / positive notes

- The `dvh` fallback pattern (keeping `vh` then overriding with `dvh`) is done correctly and consistently. 👍
- Touch event handling with `passive: true` is the right call for mobile scroll performance.
- The `layeradd` listener with a 5s safety timeout is a good defensive pattern.
- The `hasResults` dependency for the event listener effect is clean.

Overall this is in good shape. The main actionable items are the race condition fix (AbortController), the empty CSS rule, `calc(100%)` cleanup, and verifying the `50svh` intent.

Paul-AUB:
> When I search for an organization and click on one, the map pans to the right place but the balloon does not open and the marker is not shown. Maybe we should consider checking automatically the organizations marker (top-right menu on the map) when an org. is selected?

Yes I know @ClemRz , but seems a bit weird to change the map selector silently ? The problem is slightly the same for networks, and we can't display both entrances and networks :/ I can manage the specific case for organization by checking as requested @urien, but I'm neutral about the fact that's a good UX. 

> Unrelated bug: the tip of the balloon's arrow is anchored to the bottom of the marker which has the undesirable effect of hiding it:

Positioning is the same for all entities, which "fits well" for caves because it makes it totally disapear, but not for organizations. Maybe it would need to be reconsidered for all ? (I'm working on a perf improvement for entrance display, cf https://github.com/GrottoCenter/grottocenter-front/issues/1073#issuecomment-3902115471, I propose this behavior is addressed within this feature.

image


> When I search for "Jean-Bernard" and select the network, the balloon opens with the information of the "T04 (T4)" instead

Yes indeed, that's relative to my PR comment:
*For networks, I used the first entry coordinates (and considered a network = a cave with more than one entry ; definition of network vs cave is a bit... confused it seems - at least for me)*.

I can't get through API exact coordinates of a network. For network positioning when activated on map selector, a barycenter is computed from current view, but I can't do that as I *need to change current view to focus on a point*. So I decided to use the first entrance. Benefit is you can click on the entrance tootip to access to the network page directly. If you have another behavior in mind, I can change.

I will address other PR comments soon

Paul-AUB:
Hum, seems fine ?

> 7. Dead translation key Search location...
> The placeholder changed from Search location... to Search on map..., but the old key still exists in en.json. If nothing else > references it, it should be cleaned up.

https://github.com/GrottoCenter/grottocenter-front/pull/1177/changes#diff-8d9a812b4ed7e342061074bfe79526a8908c9591e3baecc5c1722e33c779fddc

Paul-AUB:
>8. Duplicated club.svg icon

> The new iconsV3/club.svg is byte-for-byte identical to the existing images/club.svg. Was there a reason to duplicate it rather than reusing the existing one (e.g. adjusting the CustomIcon path or symlinking)?

There's 6 other old icons, I won't do an exeception for this one. I propose to manage globally this issue within https://github.com/GrottoCenter/grottocenter-front/issues/329

Paul-AUB:
Fixed most comments, still have to change the behavior for the organizations.

ClemRz:
Thank you @Paul-AUB !!

Paul-AUB:
Well, I spend several hours but I can't figure out how to synchronize the orga markers display with the search results.
I think it's out of my skills :s

I propose one of:
- if @ClemRz you know how to do (and want), you can implement it in the branch
- else I can squash an we can merge as it, without the synchro of orga display when clicking on an orga result.

PS: I rebased and fixed the conflicts

Paul-AUB:
Well good news, I finally managed to implement the link between the click on an orga result and the display of orga in map ! See my last commit :)

I let you re-review and validate the behavior in staging

ClemRz (commented):
Overall this is a well-structured PR. The parallel fetching with AbortController, stale closure prevention via refs, and the `vh` → `dvh` migration with fallbacks are all solid patterns. A few things worth flagging:

## GeocodingControl.jsx is doing a lot

This file went from ~100 to ~460 lines. The search logic (fetching + transforming results for 4 entity types), the marker popup listener, and the rendering are all in one component. Consider extracting:
- A `useEntitySearch(query, map, locale)` hook that returns `{ entranceResults, networkResults, organizationResults, locationResults, loading }`
- The result rendering into a `SearchResultItem` component (the `if/if/if/return` chain in the map callback is a code smell)

## Coordinate matching tolerance is risky

```js
Math.abs(layer.getLatLng().lat - lat) < 0.001 &&
Math.abs(layer.getLatLng().lng - lng) < 0.001;
```

0.001° ≈ 111 meters. In a dense cave area with multiple entrances close together, this could match the wrong marker. A tighter tolerance (e.g., `0.00001` ≈ 1m) or exact equality on the original coordinates would be safer.

## N+1 network fetching

The cave detail fetches are sequential per cave but parallel across caves, which is fine for 3 results. But the comment says "3 network results X 3 caves details = 9 API calls" which is misleading — it's actually 1 search + 3 detail fetches = 4 calls for the network section. Minor, but the comment could confuse future readers.

## `selectedMarkersList` is recomputed every render

```js
const selectedMarkersList = Object.entries(selectedMarkers)
  .filter(([, v]) => v)
  .map(([k]) => k);
```

This creates a new array reference each render, which means the `useEffect` depending on `selectedMarkers` (the object) is fine, but the ref `selectedMarkersListRef` is updated every render regardless. Consider wrapping in `useMemo` for clarity, even if the perf impact is negligible.

## Missing keyboard accessibility on search results

The `StyledListItem` items use `cursor: pointer` and `onClick` but lack `role="option"`, `aria-selected`, or keyboard event handlers. Users navigating with a keyboard can't arrow through results or press Enter to select. Worth adding `role="listbox"` on the `ResultsList` and `role="option"` on items, plus `onKeyDown` handling.

## Magic numbers

- `setTimeout(() => { ... }, 150)` and `setTimeout(() => { ... }, 100)` in `handleSelect` — these timing values are undocumented. A brief comment explaining why 150ms vs 100ms would help.
- The 5000ms safety timeout on the `layeradd` listener is good but also worth a comment on why 5s was chosen.

## Minor nits
- **OrganizationLogo.jsx**: Changed `width: '50vh'` to `width: '50svw'` — was this intentional? `50vh` (viewport height) → `50svw` (small viewport width) is a unit *and* axis change. If the original intent was width-based, `50vw` or `50svw` makes sense, but worth confirming this wasn't a typo.
- The **PolygonMap.jsx** changes are purely formatting (closing `>` on same line). These are fine but unrelated to the feature — could be a separate commit for cleaner history.
- `handleUpdateHeat` has `useCallback(... , [])` with empty deps, which works because it reads from `selectedMarkersListRef`. This is correct but a comment explaining the empty deps + ref pattern would help reviewers unfamiliar with it.

## What looks good
- AbortController for canceling stale search requests — proper pattern
- Refs to avoid stale closures in `zoomend` handler — correct approach
- `dvh` with `vh` fallback for mobile viewport — nice progressive enhancement
- Touch event handling fix for mobile scroll in results list
- Filtering out entities without coordinates before display
- The `onOrganizationSelect` callback to auto-enable the org marker layer when selecting an org from search — thoughtful UX

Paul-AUB:
> GeocodingControl.jsx is doing a lot

I would propose to eventually refactor in another PR

> N+1 network fetching

Fixed

> selectedMarkersList is recomputed every render

Fixed

> Missing keyboard accessibility on search results

For this I've already start working on a modernisation of the map interface, that would come in that batch ;) but in another PR

> Coordinate matching tolerance is risky

I think it's "ok", anyway if another entrance is close, it would be from the same network (which one we want to target anyway).

> Magic numbers

Fixed



If it's good for you, I can squash before merge. 

NB: is it possible to activate "squash and merge" option in github from the PR ? Would gain some time to be done directly.

ClemRz (approved):
Thank you @Paul-AUB! Great job! Please remember to rebase your branch and then squash before the merge ;)

PR #1184, opened 2026-02-21
feat: optimize map performances

### LINKED ISSUE #892: [MAP] Improve performance by only calling API on real region change
We could do the same as the mobile App : 
- we store the extends used to load the current set of entrances (markers) to display.
- on map action (zoom, move, ...) we check if the new extends is larger than the saved one
- only if the new extends is larger then we do an API call
- otherwise we should only hide/show the entrance markers from the map using the locally saved list of entrances.

### LINKED ISSUE #1073: [MAP] Améliorer la fluidité dans les zones a forte densité
Il est un peu difficile de faire une recherche sur la carte quand le nombre de cavités est très important sur une portion de carte limité

Ajuster les paramètres pour augmenter la fluidité s'il y a beaucoup de trous. Par exemple en gardant des cluster sur la carte avec un zoom plus important et passer à l'affichage des points avec un zoom plus important qu'actuellement

Paul-AUB:
# Optimize map performances

## 🤔 What

- Improve the display of zones with lot of markers, and markers display strategy globally
- Improve the display by not re-rendering the map at each move when not necessary
- Load all entrancesCoordinates at once (~700Ko) to populate heatmaps. **Really** more efficicent than big calls (anyway, the first worldwild call already fetched all coordinates when full unzoomed...)
- Give priority to map moving over markers display to avoid "freezeing"
- Replace cave markers with colored circle for better performances (and evolutivity, for potential filter/sort capability based on color, etc.)
- The colored circles are displayed differently depending on the *size* of the cave
- Fix remaining orphan tooltip when zooin
- Fix on mobiles devices, both popup and tooltipe where displayed
- Fix prevent clicking (on desktop) on an hex while panning. Also change cursor to pointer when hoovering an hex.
- (Internal) code cleaning

## 🤷‍♂️ Why

See https://github.com/GrottoCenter/grottocenter-front/issues/1073

I think whith heatmap full load, https://github.com/GrottoCenter/grottocenter-front/issues/892 is no more relevant. It doesn't apply for heatmap, only when zoomin with not pan for juste entrances. This precise case is very rare, and in such case the API response qui quite short (not a lot of points).

## 🔍 How

Biggest points where :
- global display strategy 
- load once all heatmap points
- using circles over svg markers.

## 🧪 Testing

Tested manually on mobile and desktop. Really better for high densit zone markers, now the bottleneck is rather the heatmap hex display due to long API calls.

## 📸 Previews

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,4 @@ typings/
 AGENTS.md
 .amazonq
 .kiro
+.claude
--- a/packages/web-app/src/actions/Map.js
+++ b/packages/web-app/src/actions/Map.js
@@ -9,8 +9,8 @@ import {
 import makeErrorMessage from '../helpers/makeErrorMessage';
 import { makeUrl } from './utils';
 
-export const FETCH_MAP_START_LOADING = 'FETCH_MAP_LOADING';
-export const FETCH_MAP_END_LOADING = 'FETCH_MAP_LOADING';
+export const FETCH_MAP_START_LOADING = 'FETCH_MAP_START_LOADING';
+export const FETCH_MAP_END_LOADING = 'FETCH_MAP_END_LOADING';
 export const FETCH_MAP_NETWORKS_SUCCESS = 'FETCH_MAP_NETWORKS_SUCCESS';
 export const FETCH_MAP_NETWORKS_FAILURE = 'FETCH_MAP_NETWORKS_FAILURE';
 export const FETCH_MAP_NETWORKS_COORDINATES_SUCCESS =
@@ -27,10 +27,6 @@ export const FETCH_MAP_ORGANIZATIONS_SUCCESS =
   'FETCH_MAP_ORGANIZATIONS_SUCCESS';
 export const FETCH_MAP_ORGANIZATIONS_FAILURE =
   'FETCH_MAP_ORGANIZATIONS_FAILURE';
-export const CHANGE_LOCATION = 'CHANGE_LOCATION';
-export const CHANGE_ZOOM = 'CHANGE_ZOOM';
-export const FOCUS_ON_LOCATION = 'FOCUS_ON_LOCATION';
-
 export const LOADINGS = {
   NETWORKS: 'networks',
   NETWORKS_COORDINATES: 'networks_coordinates',
@@ -39,51 +35,65 @@ export const LOADINGS = {
   ORGANIZATIONS: 'organizations'
 };
 
-export const fetchNetworksCoordinates = criteria => {
-  const thunkToDebounce = function (dispatch) {
-    dispatch({
-      type: FETCH_MAP_START_LOADING,
-      key: LOADINGS.NETWORKS_COORDINATES
-    });
-    const completedUrl = makeUrl(getMapCavesCoordinatesUrl, criteria);
-    return fetch(completedUrl)
+// Heatmap coordinates are fetched once at startup with world-wide bounds rather than
+// on every moveend. The @asymmetrik/leaflet-d3 hexbin layer filters points client-side,
+// so it only renders hexagons visible in the current viewport regardless of dataset size.
+//
+// Benchmark (2025): ~130k entrances → ~2.6 MB uncompressed, ~700 KB gzipped.
+// One-time cost on page load vs. a bounded API call on every pan/zoom
+
+// Retries the fetch up to maxRetries times with exponential backoff (1 s, 2 s, 4 s…).
+// Rejects only after all attempts are exhausted.
+const fetchWithRetry = (url, maxRetries = 3) => {
+  const attempt = (retriesLeft, delay) =>
+    fetch(url)
       .then(response => {
-        if (response.status >= 400) {
-          throw new Error(response.status);
-        }
+        if (response.status >= 400) throw new Error(response.status);
         return response.text();
       })
-      .then(text => {
-        dispatch({
-          type: FETCH_MAP_NETWORKS_COORDINATES_SUCCESS,
-          data: JSON.parse(text)
-        });
-      })
       .catch(error => {
-        dispatch({
-          type: FETCH_MAP_NETWORKS_COORDINATES_FAILURE,
-          error: makeErrorMessage(
-            error.message,
-            `Fetching networks coordinates`
-          )
-        });
-      })
-      .finally(() => {
-        dispatch({
-          type: FETCH_MAP_END_LOADING,
-          key: LOADINGS.NETWORKS_COORDINATES
-        });
+        if (retriesLeft === 0) throw error;
+        return new Promise(resolve => setTimeout(resolve, delay)).then(() =>
+          attempt(retriesLeft - 1, delay * 2)
+        );
       });
-  };
+  return attempt(maxRetries, 1000);
+};
 
-  thunkToDebounce.meta = {
-    debounce: {
-      time: 500,
-      key: 'FETCH_MAP_NETWORKS_COORDINATES'
-    }
-  };
+const MAX_BOUNDS = {
+  sw_lat: -90,
+  sw_lng: -180,
+  ne_lat: 90,
+  ne_lng: 180
+};
 
-  return thunkToDebounce;
+export const fetchAllNetworksCoordinates = () => dispatch => {
+  dispatch({
+    type: FETCH_MAP_START_LOADING,
+    key: LOADINGS.NETWORKS_COORDINATES
+  });
+  return fetchWithRetry(makeUrl(getMapCavesCoordinatesUrl, MAX_BOUNDS))
+    .then(text => {
+      dispatch({
+        type: FETCH_MAP_NETWORKS_COORDINATES_SUCCESS,
+        data: JSON.parse(text)
+      });
+    })
+    .catch(error => {
+      dispatch({
+        type: FETCH_MAP_NETWORKS_COORDINATES_FAILURE,
+        error: makeErrorMessage(
+          error.message,
+          `Fetching all networks coordinates`
+        )
+      });
+    })
+    .finally(() => {
+      dispatch({
+        type: FETCH_MAP_END_LOADING,
+        key: LOADINGS.NETWORKS_COORDINATES
+      });
+    });
 };
 
 export const fetchNetworks = criteria => {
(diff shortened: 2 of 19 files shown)


Paul-AUB:
I let you try it on staging env. I think it's really more fluid.

@urien Note the change on markers for entrances. I don't know if it's acceptable for you ? Design can be adapted a bit if needed, but using svg markers is really really not efficient on performances point of vue.

Example of an improvement with the circles, defining categories of caves to visually differentie the "real" caves from just a tiny hole with no interest :

image
image

I have a specific question for @urien (but other insights are welcome ;) ): 
What would be the caracteristics of the following in terms of depth and length ? Currenly I set : 
- large cave  (depth > 100 || length > 1000) 
- medium cave (depth >= 20 && depth <= 100) || (length >= 50 && length <= 200))
- small cave (others, smaller)

Maybe for medium it would worth at least **30m depth** or **100m length** ?

urien:
Je vais faire voter les membres en leur expliquant les enjeux

Paul-AUB:
Ok, if some additional technical details are needed, tell me :)
(Basically these changes are inspired from what is done on grottomap, really better optimized)

urien:
Peux tu me donner ton mel sur Linkdin. Je vais t'inviter a rejoindre le des développeurs sur Slack

ClemRz:
Beau travail! Attention, le clustering est repris par la page des massifs donc il faut veiller à bien propager les changements ici aussi, par exemple: https://orange-rock-0d4f87503-1184.westeurope.azurestaticapps.net/ui/massifs/1

ClemRz (commented):
## Code Review — PR #1184: Optimize map performances

Overall this is a solid performance-focused PR with good architectural decisions (one-time heatmap fetch, lazy popups, RAF coalescing, `moveend` over `dragend`, incremental marker diffing). Nice work. A few issues worth addressing before merge:

---

### Bugs

**1. Cave size classification has a gap (constants.js)**

The medium-cave condition leaves a hole: lengths between 200 and 1000 fall through to `SMALL` instead of `MEDIUM` or `LARGE`.

```js
if (depth > 100 || length > 1000) return CAVE_SIZE.LARGE;
if ((depth >= 20 && depth <= 100) || (length >= 50 && length <= 200))
  return CAVE_SIZE.MEDIUM;
return CAVE_SIZE.SMALL;
```

A cave with depth=5 and length=500 is classified as SMALL. That seems unintentional — the medium range for length should probably extend up to 1000 to close the gap:

```js
if (depth > 100 || length > 1000) return CAVE_SIZE.LARGE;
if (depth >= 20 || length >= 50) return CAVE_SIZE.MEDIUM;
return CAVE_SIZE.SMALL;
```

Or at minimum `length <= 1000` instead of `length <= 200`. Worth confirming with @urien since the thresholds are already under discussion in the PR comments.

**2. Redux state for `location`/`zoom` is no longer updated on pan/zoom (Map.jsx)**

`handleUpdate` no longer dispatches `changeLocation(center)` or `changeZoom(newZoom)`. This means `state.map.location` and `state.map.zoom` in Redux will go stale after the initial load. If any other component reads these values (e.g. deep-linking, sharing, or other pages that reference the map state), they'll get the wrong position. If nothing else consumes them, consider removing them from the reducer entirely to avoid confusion. If something does consume them, they need to stay in sync.

**3. `handleUpdate` is recreated every render (Map.jsx)**

`handleUpdate` is defined as a plain function inside the component body, capturing `dispatch`, `navigate`, `urlDebounceRef` via closure. It's passed as `onUpdate` to `MapClusters`, which wraps it in a ref (`onUpdateRef`). This works, but it means `MapClusters` re-renders on every parent render since `onUpdate` is a new reference each time. Consider wrapping it in `useCallback`.

---

### Potential Issues

**4. ~2.6 MB payload on every page load (actions/Map.js)**

Fetching all 130k entrance + network coordinates on mount is a bold trade-off. The comment says ~700 KB gzipped, which is reasonable on broadband but could be painful on mobile/slow connections. Consider:
- Adding error handling with retry logic (currently a single failed fetch = no heatmap for the session)
- A loading indicator while the initial fetch is in progress
- Potentially caching the response (service worker, `Cache-Control` headers on the API side)

**5. `initialZoom` is computed outside of state/ref (MapClusters/index.jsx)**

```js
const map = useMap();
const initialZoom = map.getZoom();
const isInitiallyZoomedIn = initialZoom >= MARKERS_LIMIT;
```

These run on every render, not just the first. `initialZoom` and `isInitiallyZoomedIn` are used to set initial `useState` values (which only take effect on mount), so it's functionally fine, but it's misleading. Consider moving them into a `useRef` or `useMemo` to signal intent:

```js
const initialZoom = useRef(map.getZoom()).current;
```

**6. Orphan `setTimeout` in drag-end handler (useHeatLayer.js)**

```js
useMapEvent('dragend', () => {
  setTimeout(() => {
    isDraggingRef.current = false;
  }, 0);
});
```

This `setTimeout` is never cleared on unmount. It's a zero-delay timer so the window is tiny, but if the component unmounts mid-drag, it'll write to a stale ref. Harmless in practice but worth a cleanup for correctness.

**7. No error boundary / fallback for the one-shot coordinate fetches**

If `fetchAllEntrancesCoordinates` or `fetchAllNetworksCoordinates` fails, the heatmap silently stays empty for the entire session. The old approach would retry on the next pan. Consider dispatching a user-visible error or adding a retry mechanism.

---

### Nits / Cleanup

**8.** `FETCH_MAP_START_LOADING` and `FETCH_MAP_END_LOADING` were both `'FETCH_MAP_LOADING'` on the base branch — great catch fixing that bug. But note the reducer on `develop` already imports and switches on both constants, so the loading state was silently broken before this PR. Might be worth calling that out in the PR description as a separate bug fix.

**9.** The `useMarkers` hook no longer clears markers when `newMarkers` is `null` and there were previously markers with open popups/tooltips. The `remove()` call should handle it, but worth verifying that Leaflet's canvas renderer properly cleans up popup DOM when a `circleMarker` is removed.

**10.** I flagged in a previous comment that the massif page also uses the clustering components. The staging link for `/ui/massifs/1` should be tested to make sure the `circleMarkerStyle` / `icon` branching in `useMarkers` doesn't break pages that still pass `icon` without `circleMarkerStyle`. Ideally the massif map should behave the same way as the main map (circle markers, same perf optimizations, etc.) — if that's not covered in this PR, it should at least not regress. Worth noting that the massif page only receives data intersecting the massif polygon, so the one-shot world-wide fetch strategy doesn't apply there — but the marker rendering changes (circle markers, lazy popups, incremental diffing) should still propagate.

---

### What's good

- Lazy popup rendering (`bindPopup(() => ...)`) is a big win — avoids `renderToString` for thousands of markers that may never be clicked.
- The incremental marker diffing via `Map<id, L.Marker>` is much cleaner than the old remove-all/add-all approach.
- RAF coalescing for hex data updates is a nice touch.
- Replacing `dragend` with `moveend` for data fetching correctly handles mobile inertia.
- The `eslint-disable react-hooks/exhaustive-deps` suppressions being removed in `Markers.jsx` is a welcome cleanup.
- URL update debouncing (1s) is a good call to avoid router churn during panning.

Paul-AUB:
1. Fixed
2. Removed
3. Fixed
4. Managed within 7. option (map is still usable, would be strange to fully lock it inside a loader
5. Fixed
6. Fixed

I also update the massif map with the main improvements (including one time fetch, but on the bounds of the massif ; I didn't implemented a retry mechanism as there's really less data here).

urien (approved):
Pour moi c'est ok, j'ai demandé aux membres de l'association 7 ont répondu et 6 sont pour ce que tu proposes.

ClemRz:
What is the primary reason to call `/api/v1/geoloc/entrancesCoordinates` with fixed bounding box -90<lat<90 -180<lng<180?

- If I open https://grottocenter.org/ui/map/43.52589967307701%2C2.5354385375976567%2C12  only 55 clusters are downloaded: the server responds in ~210ms + download of ~50ms.

v.s.

- If I open https://orange-rock-0d4f87503-1184.westeurope.azurestaticapps.net/ui/map/43.52589967307701%2C2.5354385375976567%2C12 100,000 clusters are downloaded (not the full DB, results get truncated at SQL level: `... LIMIT 100000`) and the server responds in ~1,470ms + download of ~950ms.

This endpoint was initially designed to deliver a response adapted to the viewport, as WMS usually do.

Also it's important to consider that a low number of long-latency requests is more stressful on the API server than high concurrency of short-latency requests.

Paul-AUB:
I may missunderstood something, but the idea was to ask for whole world points once (which is what is done anyway if you just go to https://grottocenter.org/ui/map and start to zoom).

So if you go through a specific link to an already zoomed map location, and never zoom out, yes indeed the new implementation is worse. But in other cases (e.g. if you don't activate geolocation on your device), it's strictly better. 

So... it's a tradeoff and really depends about what we want to optimize. Or maybe you have in mind something more elaborate to take advantage of the 2 situations ?

urien:
@ClemRz, @bsoufflet et @vmarseguerra  pouvez vous jeter un oeil sur cette PR ?

ClemRz (changes requested):
This is awesome! Coupled with the recent optimization of the `/api/v1/geoloc/entrancesCoordinates` endpoint already delivered and the DB optimization to come, the UX will be so much better, TY!

This PR makes a lot of good architectural moves: one-shot heatmap fetch, lazy popups, incremental marker diffing, RAF coalescing, `moveend` over `dragend`. The code quality is solid overall.

## Issues worth fixing

1. `handleUpdate` in `Map.jsx` is wrapped in `useCallback` with `[dispatch, navigate]` deps, which is correct. However, the dep array is missing `encodeMapTarget` — if that's ever not a stable module-level function, this could go stale. Worth confirming it's a pure import.

2. In `MapMassif.jsx`, the one-shot heat fetch effect has `[massifBounds, massifId]` as deps. Since `massifBounds` is memoized on `geoJson`, and `massifId` is a prop, this could re-fire if the parent re-renders with a new `geoJson` reference that produces the same bounds. The `useMemo` on `L.geoJSON(geoJson).getBounds()` helps, but Leaflet's `LatLngBounds` doesn't have referential equality for identical bounds — so this is safe only because `geoJson` itself is stable. If it ever comes from Redux or a fetch, this could double-fire. Low risk but worth a comment.

3. `location` and `zoom` moved from Redux to local `useState` in `Map.jsx`. The `changeLocation` and `changeZoom` imports are removed. If the reducer still has those action handlers and nothing else dispatches them, they're dead code now. Worth a follow-up to clean up the reducer, or at least a TODO.

Paul-AUB:
About 1., no change is necessary. Explanation : encodeMapTarget is defined at module level on line 23 as a plain const arrow function ; it's stable and never reassigned. It's safe to omit from the useCallback dep array since module-level constants are not reactive values in React's model.

Fixed 2. and 3.

Also fixed an ugly behavior when resizing the map, in particular when folding the menu (in a separated commit).

ClemRz (approved):

PR #1203, opened 2026-03-04
fix(massif): handle empty coordinates in GeoJSON import

ClemRz:
# fix(massif): handle empty coordinates in GeoJSON import

## 🤔 What

- Fix crash when importing GeoJSON files containing features with empty `coordinates` arrays
- Extract GeoJSON-to-MultiPolygon parsing logic into a reusable `geojsonParser` helper
- Add property-based tests (fast-check) covering valid, empty, degenerate, and mixed geometries

## 🤷‍♂️ Why

Importing a valid GeoJSON file (e.g. "New River Gorge National Park and Preserve.geojson") that contains features with empty `coordinates: []` caused a `TypeError: Cannot read properties of undefined (reading 'map')` in `PolygonMap.jsx`. The empty polygon was pushed into the MultiPolygon without validation, and downstream code assumed `polygon[0]` would always exist.

## 🔍 How

- Created `src/helpers/geojsonParser.js` with `parseGeoJsonToMultiPolygon()` that skips geometries with empty/missing coordinates, null geometry, or unsupported types (Point, LineString, etc.)
- Refactored `ShapefileImport.jsx` to use the new helper for both GeoJSON and ZIP (shapefile) parsing paths, removing duplicated inline logic and the now-unused direct `transformToWGS84` import
- Added `src/helpers/geojsonParser.test.js` with 8 property-based tests using fast-check that generate random GeoJSON structures and verify parser invariants

## 🔗 Related API PR

N/A

## 🧪 Testing

1. Upload a GeoJSON file with features containing empty `coordinates: []` — should no longer crash: [New River Gorge National Park and Preserve.geojson.txt](https://github.com/user-attachments/files/25748844/New.River.Gorge.National.Park.and.Preserve.geojson.txt) (remove the `.txt` from the file name)
2. Upload a normal GeoJSON or shapefile ZIP — should work as before
3. Run `npx react-scripts test --watchAll=false --testPathPattern='geojsonParser.test'` — all 8 tests pass

## 📸 Previews

N/A — no visual changes.

### DIFF
--- a/packages/web-app/src/components/appli/EntitiesForm/Massif/ShapefileImport.jsx
+++ b/packages/web-app/src/components/appli/EntitiesForm/Massif/ShapefileImport.jsx
@@ -15,8 +15,8 @@ import {
 import { Upload } from '@mui/icons-material';
 import simplify from 'simplify-js';
 import shp from 'shpjs';
-import { transformToWGS84 } from '../../../../helpers/coordinateTransform';
 import { countMultiPolygonVertices } from '../../../../helpers/vertexCount';
+import { parseGeoJsonToMultiPolygon } from '../../../../helpers/geojsonParser';
 
 // Simplify polygon using simplify-js library
 const simplifyPolygon = (multiPolygon, toleranceValue) => ({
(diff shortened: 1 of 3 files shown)


Paul-AUB (commented):
No "real" error, but some potential improvements to make it more "futur proof".
Feel free to fix if you think it is needed.

## Issues

### 1. `extractPolygons` exported as default — misleading API · **Medium**

[geojsonParser.js:83](../helpers/geojsonParser.js#L83)

```js
export default extractPolygons;
```

`extractPolygons` is an internal helper. Exporting it as the default export signals it's the primary public API, when `parseGeoJsonToMultiPolygon` is. This will confuse consumers who do `import geojsonParser from './geojsonParser'`. Either:
- Remove the default export entirely (it's not imported anywhere), or
- Make `parseGeoJsonToMultiPolygon` the default export.

---

### 2. `Feature` with `null` geometry crashes in `FeatureCollection` path · **High**

[geojsonParser.js:56](../helpers/geojsonParser.js#L56)

```js
geojson.features.forEach(feature => {
  multiPolygon.coordinates.push(
    ...extractPolygons(feature.geometry, sourceCRS)
  );
});
```

`feature.geometry` can legally be `null` in GeoJSON (a feature with no geometry). `extractPolygons` does guard `if (!geometry ...)` so this is safe **now**, but only because of the guard inside `extractPolygons`. The test suite covers `null` geometry inside a `FeatureCollection` via `degenerateGeometryArb` — good — but the caller still spreads `extractPolygons(null, ...)` which returns `[]`, so it works. No code change needed, but worth a comment explaining the null case is intentionally handled by `extractPolygons`.

---

### 3. `transformToWGS84` call on `MultiPolygon` filters **before** transforming · **Medium**

[geojsonParser.js:36–43](../helpers/geojsonParser.js#L36)

```js
const validPolygons = geometry.coordinates.filter(
  poly => poly[0] && poly[0].length > 0
);
if (validPolygons.length > 0) {
  const transformed = sourceCRS
    ? transformToWGS84(validPolygons, sourceCRS, 3)
    : validPolygons;
  return transformed;
}
```

After filtering, the indices of the polygons in `validPolygons` no longer correspond to the original `geometry.coordinates`. `transformToWGS84` operates on the filtered array — that's correct. But if `transformToWGS84` ever needs the original array shape (e.g. preserves holes by index), the mismatch could silently produce wrong output. This is low-risk with the current `coordinateTransform` implementation but fragile. A comment clarifying this is intentional would help.

---

## Positive notes

- Refactoring inline GeoJSON parsing into `parseGeoJsonToMultiPolygon` is a clear improvement — removes ~60 lines of duplicated code from `ShapefileImport.jsx`.
- The property-based tests with `fast-check` are excellent: they cover degenerate inputs, mixed collections, and structural contracts.
- The `null` return (vs empty MultiPolygon) makes the "no results" case explicit and unambiguous at the call site.

Paul-AUB (approved):
LGTM !

PR #1230, opened 2026-03-18
fix: advance search sliders for caves

Paul-AUB:
### Description

In the advanced search for entrances/cavities, the **depth** and **length** filters were silently broken: changing the slider had no effect on search results.

image

### Root cause

In `EntrancesSearch.jsx`, the `onChange` callbacks for both sliders were wrapping the emitted value with `Math.trunc()`:

```js
// Before (broken)
onChange={e => updateFilter('cave.depth', Math.trunc(e))}
onChange={e => updateFilter('cave.length', Math.trunc(e))}
```

However, `SearchSlider` emits a **range array** `[min, max]` (e.g. `[50, 1000]`), not a single number. Calling `Math.trunc()` on an array coerces it to `NaN`:

```js
Math.trunc([50, 1000]) // → NaN
```

As a result, `filterState['cave.depth']` and `filterState['cave.length']` were always `NaN`, and the backend received invalid filter values regardless of what the user selected.

The other sliders in the same form (ratings) were not affected because they pass `e` directly without wrapping.

### Fix

Remove the `Math.trunc()` wrapper and pass the array value through unchanged:

```js
// After (fixed)
onChange={e => updateFilter('cave.depth', e)}
onChange={e => updateFilter('cave.length', e)}
```

### Affected file

`packages/web-app/src/components/appli/AdvancedSearch/EntrancesSearch.jsx`

## Screenshot of fixed feature

image

### DIFF
--- a/packages/web-app/src/components/appli/AdvancedSearch/EntrancesSearch.jsx
+++ b/packages/web-app/src/components/appli/AdvancedSearch/EntrancesSearch.jsx
@@ -194,13 +194,13 @@ const EntrancesSearch = () => {
           label="Depth"
           helperText="In meters"
           marks={depthMarks}
-          onChange={e => updateFilter('cave.depth', Math.trunc(e))}
+          onChange={e => updateFilter('cave.depth', e)}
         />
         <SearchSlider
           label="Length"
           helperText="In meters"
           marks={lengthMarks}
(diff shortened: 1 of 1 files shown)


urien:
Il faut l'avis d'un développeur !
Remercie ton copain pour cette remarque très bienvenue

ClemRz (approved):

Issue #1235, opened 2026-03-19
Document Advanced Search 500 — Field Name Mismatch


ClemRz:
## Summary

The advanced search for documents returns a 500 error when filtering by entrance or massif name. The front-end sends singular field names (`entrance.name`, `massif.name`) but the Typesense schema expects plural names (`entrances.name`, `massifs.name`).

## Error

```
POST /api/v1/advanced-search → 500
RequestMalformed: Could not find a filter field named `entrance.name` in the schema.
```

## Root Cause

The back-end Typesense collection schema for `documents` defines these fields (in `api/dbSync/entities/document.js`):

| Schema field      | Type       |
|-------------------|------------|
| `cave.name`       | `string`   |
| `entrances.name`  | `string[]` |
| `massifs.name`    | `string[]` |

The front-end sends these filter keys in the request body:

| Sent by front-end | Expected by Typesense |
|--------------------|-----------------------|
| `cave.name` ✅     | `cave.name`           |
| `entrance.name` ❌ | `entrances.name`      |
| `massif.name` ❌   | `massifs.name`        |

`cave.name` works because it's singular in both places (a document has one cave). `entrance.name` and `massif.name` fail because the back-end schema uses plural names — a document can relate to multiple entrances and multiple massifs.

The `buildFilter()` function in `SearchService.js` passes filter keys through to Typesense's `filter_by` parameter as-is, with no field name remapping.

## What to Fix (Front-End)

In the front-end repo (`grottocenter-front`), find where the document advanced search filter object is built. The filter field names for entrance and massif need to be changed from singular to plural:

- `entrance.name` → `entrances.name`
- `massif.name` → `massifs.name`

Look for the component or service that constructs the filter payload for `POST /api/v1/advanced-search` when `entity` is `"documents"`. The field names are likely defined in a form configuration, a constants file, or directly in the search component.

## Reproduction

Search documents on grottocenter.org with any value in the entrance name or massif name filter fields. The API returns a 500 instead of results.

PR #1236, opened 2026-03-19
fix(Search): rename entrance.name and massif.name to plural in document advanced search

ClemRz:
# fix(Search): rename entrance.name and massif.name to plural in document advanced search

## 🤔 What

- Rename `entrance.name` → `entrances.name` in the document advanced search filter state and JSX
- Rename `massif.name` → `massifs.name` in the document advanced search filter state and JSX
- Fix the documents entity table `entrance` column: rename to `Entrances`, use `entrances` array field with `keyArray('name')` renderer, mark as not sortable (array fields can't be sorted in Typesense)
- Close issue #1235

## 🤷♂️ Why

The Typesense collection schema for `documents` defines `entrances.name` and `massifs.name` as plural fields (arrays). The front-end was sending the singular variants in two places:
1. The advanced search filter — causing a 500 error (`RequestMalformed: Could not find a filter field named entrance.name in the schema`) whenever a user filtered documents by entrance or massif name.
2. The entity table column config — the `entrance.name` field didn't match the actual API response shape (`entrances[].name`), so the entrance name never rendered; and since `entrances` is an array field, sorting on it is not supported by Typesense.

## 🔍 How

4 occurrences updated in `DocumentSearch.jsx`:
- `initialFilterState` keys renamed from singular to plural
- `onChange` / `value` props on the Entrance and Massif `SearchText` fields updated accordingly

1 occurrence updated in `entitiesConfig.jsx`:
- `entrance.name` column replaced with `entrances` using `keyArray('name')` renderer to correctly display the array, labeled `Entrances`, and marked as not sortable (consistent with the `massifs` column)

## 🔗 Related API PR

N/A — the schema is already correct on the back-end side.

## 🧪 Testing

1. Go to the advanced search page and select **Documents**
2. Enter any value in the **Entrance** or **Massif** name fields
3. Submit — should return results instead of a 500 error
4. The **Entrances** column in the results table should now display the entrance name(s)

### DIFF
--- a/packages/web-app/src/components/appli/AdvancedSearch/DocumentSearch.jsx
+++ b/packages/web-app/src/components/appli/AdvancedSearch/DocumentSearch.jsx
@@ -47,8 +47,8 @@ const initialFilterState = {
   issue: '',
   'parent.title': '',
   'cave.name': '',
-  'entrance.name': '',
-  'massif.name': ''
+  'entrances.name': '',
+  'massifs.name': ''
 };
 
 const SubjectEntry = ({ subject }) => {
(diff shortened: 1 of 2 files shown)


urien (approved):
Bravo @ClemRz c'est une fonctionnalité très utile... Et qui fonctionne désormais

Paul-AUB (approved):

PR #1238, opened 2026-03-20
fix(auth): logout user on 401 response

### LINKED ISSUE #778: [SECURITY] Remove Logout user on 401 unauthorized response
Cette fonctionnalité a été mise en place de manière à apporter une solution alors que les utilisateurs pouvaient se trouver dans une situation où ils pouvaient penser être connecté alors qu'ils ne le sont plus.

Il faudra mettre en oeuvre une autre fonctionnalité qui nécessitera plus de développement mais qui sera plus satisfaisante
Par exemple utiliser  utiliser checkAndGetStatus dans toutes les actions.

ClemRz:
# fix(auth): logout user on 401 response

Closes #778

## 🤔 What

- Automatically log out the user when any API request returns a 401 response
- Extract Redux store creation into a dedicated `store.js` file
- Make `checkAndGetStatus` async for consistent promise handling

## 🤷♂️ Why

When a user's JWT token expires or is revoked, the API returns a 401. The app was not logging the user out in this case because:

1. The 401 check in `CountUnreadNotifications` compared `error.message === '401'`, but `checkAndGetStatus` was throwing with `body.message` (e.g. `"Unauthorized"`) instead of the status code
2. Even after fixing that, CORS policy blocks the response entirely for cross-origin 401s, so `fetch` throws `TypeError: Failed to fetch` before `checkAndGetStatus` even runs — meaning the status code was never readable from the catch block

## 🔍 How

- Moved the logout logic into `checkAndGetStatus` in `utils.js`, where the raw `response.status` is directly accessible before any CORS/fetch error can obscure it
- The store is accessed via a lazy `require('../store')` to avoid pulling in the full reducer tree at module load time (which broke Jest tests due to `intlBootstrap` not being defined in the test environment)
- Extracted store creation from `ApplicationShell.jsx` into `src/store.js` so it can be required lazily
- Made `checkAndGetStatus` `async` for consistent promise handling — two callers (`GetNotifications`, `GetMenuNotifications`) were missing `await`, causing unhandled promise rejections
- Removed the now-redundant logout logic from `CountUnreadNotifications`

## 🔗 Related API PR

N/A

## 🧪 Testing

1. Log in
2. Revoke or expire the JWT token on the backend
3. Navigate to a page that triggers an authenticated API call
4. Verify the user is automatically logged out

## 📸 Previews

N/A

### DIFF
--- a/packages/web-app/src/actions/Notifications/CountUnreadNotifications.js
+++ b/packages/web-app/src/actions/Notifications/CountUnreadNotifications.js
@@ -2,7 +2,6 @@ import fetch from 'isomorphic-fetch';
 import { countUnreadNotificationsUrl } from '../../conf/apiRoutes';
 import makeErrorMessage from '../../helpers/makeErrorMessage';
 import { checkAndGetStatus } from '../utils';
-import { logout } from '../Login';
 
 export const COUNT_UNREAD_NOTIFICATIONS = 'COUNT_UNREAD_NOTIFICATIONS';
 export const COUNT_UNREAD_NOTIFICATIONS_SUCCESS =
@@ -40,14 +39,6 @@ export function countUnreadNotifications() {
         dispatch(countUnreadNotificationsActionSuccess(data.count));
       })
       .catch(error => {
-        // When logged, countUnreadNotifications() is the first authenticated request made to the API
-        // So it is the first request that test if the JWT is really valid
-        // In case of invalid token we logout the user
-        if (error.message === '401') {
-          dispatch(logout());
-          return;
-        }
-
         dispatch(
           countUnreadNotificationsActionFailure(
             makeErrorMessage(error.message, `Counting unread notifications`),
--- a/packages/web-app/src/actions/Notifications/GetMenuNotifications.js
+++ b/packages/web-app/src/actions/Notifications/GetMenuNotifications.js
@@ -37,7 +37,7 @@ export function fetchMenuNotifications(criterias) {
     };
 
     try {
-      const response = checkAndGetStatus(
+      const response = await checkAndGetStatus(
         await fetch(makeUrl(fetchNotificationsUrl, criterias), requestOptions)
       );
 
--- a/packages/web-app/src/actions/Notifications/GetNotifications.js
+++ b/packages/web-app/src/actions/Notifications/GetNotifications.js
@@ -32,7 +32,7 @@ export function fetchNotifications(criterias) {
     };
 
     try {
-      const response = checkAndGetStatus(
+      const response = await checkAndGetStatus(
         await fetch(makeUrl(fetchNotificationsUrl, criterias), requestOptions)
       );
 
--- a/packages/web-app/src/actions/utils.js
+++ b/packages/web-app/src/actions/utils.js
@@ -8,6 +8,7 @@ import {
   defaultTo,
   equals
 } from 'ramda';
+import { logout } from './Login';
 
 // Remove the next line when other exports are created.
 export const makeUrl = (url, criterias) => {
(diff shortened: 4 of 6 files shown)


Paul-AUB (changes requested):
# Code Review — `fix/auth-logout-on-401`

## Issues

### 1. Regression: API error message is lost

**File:** [actions/utils.js:32](packages/web-app/src/actions/utils.js#L32)

```js
// Before
const errorMessage = new Error(body.message || response.status);

// After
const errorMessage = new Error(response.status);
errorMessage.body = await response.json();
```

The error message is now always the raw HTTP status code (e.g. `"404"`), whereas before it used `body.message` first (the human-readable message returned by the API). All callers passing `error.message` to `makeErrorMessage(...)` will now display a numeric code instead of the API's message.

I don't know if not displaying the message from API is in purpose or not. **If yes, ignore this comment.**

**Suggested fix:**
```js
const errorMessage = new Error(response.status);
try {
  errorMessage.body = await response.json();
  if (errorMessage.body?.message) {
    errorMessage.message = errorMessage.body.message;
  }
} catch (_) {
  // non-JSON body
}
```

---

### 2. Error is still thrown after logout on 401

**File:** [actions/utils.js:27-38](packages/web-app/src/actions/utils.js#L27)

```js
if (response.status === 401) {
  const gcStore = require('../store').default;
  gcStore.dispatch(logout());
}
// ... throw errorMessage; <-- still executed
```

After dispatching `logout()`, the error is still thrown. Calling actions will enter their `.catch()` handler and may dispatch a "401" error notification to the user — who is already being logged out. This is unnecessary noise.

**Suggested fix:** return after dispatching to short-circuit the error on 401:
```js
if (response.status === 401) {
  const gcStore = require('../store').default;
  gcStore.dispatch(logout());
  return response; // or throw a silent error
}
```

Note: check whether any callers have specific 401 handling in their `.catch()` that could be affected.

---

## Minor remarks

### 3. Dynamic `require` to avoid circular dependency

**File:** [actions/utils.js:29](packages/web-app/src/actions/utils.js#L29)

```js
// eslint-disable-next-line global-require
const gcStore = require('../store').default;
```

The eslint-disable comment signals this is a known workaround. The circular dependency exists because:
`store.js` → `GCReducer` → (reducers import action constants) → (actions import `utils.js`) → `utils.js` would import `store.js`

Acceptable short-term. A cleaner long-term approach would be a Redux middleware that intercepts HTTP errors and has native access to the store, without any circular import.

### 4. `status <= 300` includes 300 (Multiple Choices)

**File:** [actions/utils.js:24](packages/web-app/src/actions/utils.js#L24)

```js
if (response.status >= 200 && response.status <= 300) {
```

Status 300 is a redirect, not a success. Should be `status < 300`. Pre-existing bug, not introduced by this branch, but worth fixing.

---

## What's good

- **Centralization**: removing the ad-hoc workaround from `CountUnreadNotifications.js` is the right call. All API calls now benefit from automatic logout.
- **`await` added** in `GetMenuNotifications.js` and `GetNotifications.js`: required since `checkAndGetStatus` became async.
- The 80+ other files using `.then(checkAndGetStatus)` do not need changes — `.then()` automatically unwraps returned Promises.
- **`store.js` extraction**: clean separation of concerns. The `typeof window !== 'undefined'` guard for Redux DevTools is an improvement over the direct `window` access that was in `ApplicationShell.jsx`.
- **`store.js` is now reusable**: can be imported anywhere without recreating a store instance.

---

## Recommended actions

| Priority | Issue                                                     | File               |
|----------|-----------------------------------------------------------|--------------------|
| High     | API error message lost (`error.message` = raw HTTP code)  | `actions/utils.js` |
| High     | 401 error toast shown to user during logout               | `actions/utils.js` |
| Low      | `status <= 300` should be `status < 300` (pre-existing)   | `actions/utils.js` |

ClemRz:
Thanks for the thorough review @Paul-AUB! All three actionable items addressed:

1. **API error message restored** — `errorMessage.message` is now set from `body.message` when available, so callers get human-readable messages again.
2. **Early return on 401** — After dispatching `logout()`, we `return response` instead of falling through to `throw`. No more error toasts during logout.
3. **`< 300` instead of `<= 300`** — Fixed the pre-existing off-by-one so status 300 is no longer treated as success.

Issue number 3 (dynamic `require`) — agreed it's acceptable short-term; noted the middleware suggestion for later.

Paul-AUB (approved):

PR #1244, opened 2026-03-31
fix(search): validate sort fields against API whitelist

ClemRz:
# fix(search): validate sort fields against API whitelist

## 🤔 What

- Corrected `sortable` flags in `entitiesConfig.jsx` to match the API's actual sortable fields per entity type
- Added an `ALLOWED_SORT_FIELDS` whitelist in `EntityTable.jsx` as a defense-in-depth guard that prevents invalid sort fields from reaching the API
- Enabled sorting on columns that the API supports but the UI wasn't exposing (e.g., `documents.importSource`, `documents.importId`, `documents.dateInscription`)

## 🤷‍♂️ Why

The UI was marking some columns as sortable even though their field names weren't accepted by the Typesense search engine on the backend, causing 500 errors when users clicked those column headers to sort. Conversely, some API-sortable fields weren't exposed in the UI.

## 🔍 How

Two-layer fix:

1. **`entitiesConfig.jsx`**: Audited every column's `sortable` flag against the API's sortable fields list. Corrected mismatches in both directions (disabled non-sortable, enabled missing sortable).

2. **`EntityTable.jsx`**: Added a `ALLOWED_SORT_FIELDS` constant (a map of entity type → Set of valid field names). `handleRequestSort()` now checks this whitelist before calling `onSortChange` — if the field isn't allowed, the sort is silently ignored. This prevents future regressions if column configs drift from the API schema.

## 🔗 Related API PR

N/A

## 🧪 Testing

1. Open Advanced Search, search for organizations
2. Verify "Cavers", "Is a partner" columns show sort indicators (they're now valid)
3. Search for documents, verify "Added at", "Import source", "Import Id" are sortable
4. Search for massifs, verify "Entrances" is sortable
5. Verify all existing sort functionality still works (entrances by depth, name, etc.)

## 📸 Previews

N/A — no visual changes beyond sort indicators appearing/disappearing on corrected columns.

### DIFF
--- a/packages/web-app/src/components/common/EntityTable/EntityTable.jsx
+++ b/packages/web-app/src/components/common/EntityTable/EntityTable.jsx
@@ -36,6 +36,15 @@ import Translate from '../Translate';
 const DEFAULT_PAGE_SIZE_OPTIONS = [20, 100, 200];
 const MAX_DOCUMENTS_TO_EXPORT_IN_CSV = 10000; // This limit is also enforced server side
 
+// Derive the whitelist of sortable fields from entitiesConfig so there is a single
+// source of truth.  col[3] is the sortable flag, col[5] ?? col[1] is the API field name.
+const ALLOWED_SORT_FIELDS = Object.fromEntries(
+  Object.entries(entitiesConfig).map(([type, config]) => [
+    type,
+    new Set(config.columns.filter(col => col[3]).map(col => col[5] ?? col[1]))
+  ])
+);
+
 const applyColumnVisibility = (columns, storedVisibility) => {
   try {
     const visibleColumns = JSON.parse(storedVisibility);
@@ -264,6 +273,13 @@ const EntityTable = ({
   };
 
   const handleRequestSort = (event, property) => {
+    const allowedFields = ALLOWED_SORT_FIELDS[entityType];
+    if (!allowedFields || !allowedFields.has(property)) {
+      // eslint-disable-next-line no-console
+      console.warn(`Sort blocked: field "${property}" is not allowed for entity type "${entityType}"`);
+      return;
+    }
+
     if (orderBy === property) {
       let newOrder = 'asc';
       if (order === 'asc') newOrder = 'desc';
--- a/packages/web-app/src/components/common/EntityTable/entitiesConfig.jsx
+++ b/packages/web-app/src/components/common/EntityTable/entitiesConfig.jsx
@@ -131,6 +131,7 @@ const documents = {
     [false, 'reviewer', 'Reviewer', false, cellsRender.person],
     [false, 'validator', 'Validator', false, cellsRender.person],
     [false, 'creatorComment', 'Creator comment', false],
+    [false, 'language', 'Language', true],
     [true, 'type', 'Type', true, cellsRender.translate],
     [true, 'title', 'Title', true, cellsRender.ellipsis],
     [true, 'description', 'Description', false, cellsRender.ellipsis],
@@ -146,8 +147,8 @@ const documents = {
     [false, 'license', 'License', true],
     [false, 'subjects', 'Subjects', false, cellsRender.keyArray('code')],
     [true, 'iso3166', 'Country / Region', false, cellsRender.keyArray('iso'), 'iso3166.iso'],
-    [false, 'importSource', 'Import source', false],
-    [false, 'importId', 'Import Id', false],
+    [false, 'importSource', 'Import source', true],
+    [false, 'importId', 'Import Id', true],
     [false, 'cave.name', 'Cave', true],
     [false, 'entrances', 'Entrances', false, cellsRender.keyArray('name')],
     [false, 'massifs', 'Massifs', false, cellsRender.keyArray('name')]


urien:
Salut @ClemRz 
J'ai vérifié les points indiqué et j'ai revu l'ensemble des champs qui sont triables pour toutes les entités
Dans les champs liés aux documents il y a un diysfonctionnement
quand je trie la licence
image
la liste des champs accessibles change
image
Quand je demande a trier id d'import ou source d'import
image
Les champs accessibles changent
image

Dans le champ Langue il semble qu'il n'y ai aucune valeur pour les entrées
image
De même pour les massifs
image
Pour les organisations
image
Pour les entrées
image

*bilan*
Ce qui pris en charge par ce ticket fonctionne mais d'autres bugs apparaissent à l'occasion des tests. Voir si ces bugs seront pris en charge dans le cadre de ce ticket ou si un autre ticket doit être réalisé

urien:
Hello  @ClemRz , 
Je crois que j'ai compris le pb que je rencontrais : quand on sélectionne tous les champs, certains champs ne sont pas visibles, sauf a diminuer fortement la taille des caractères. Du coup a l'écran les champs visibles dans la partie droite de la page changent quand on demande a faire un tri.
> est ce qu'il serait possible de mettre un curseur horizontal afin de l'ensemble des champs soient visibles quand on en sélectionne beaucoup ?
Pour les documents la langue ne fait plus partie des champs proposés
Pour les entrées, les massifs et les organisations le champ langue est bien proposé mais il n'y a aucune valeur pour ce champ

Paul-AUB (changes requested):
# Code Review — `fix(search): validate sort fields against API whitelist`

Commit: `3f69c90f65bfb15350b509b96d4ee4380486a631`

## What's good

- The guard in `handleRequestSort` is minimal and clear (3 lines).
- Using `Set` for O(1) lookups is appropriate.
- Making `importSource` and `importId` sortable in `entitiesConfig.jsx` is a legitimate fix.

---

## Issues

### 1. Bug: `language` missing from `ALLOWED_SORT_FIELDS.documents`

The most direct inconsistency in this commit: `language` is added as sortable in `entitiesConfig.jsx`...

```js
// entitiesConfig.jsx
[false, 'language', 'Language', true],  // 4th element = true = sortable
```

...but `language` is absent from the `documents` Set in `ALLOWED_SORT_FIELDS`:

```js
documents: new Set([
  'importId', 'importSource', ..., 'cave.name'
  // ❌ 'language' missing
]),
```

Result: the UI renders a sort arrow for `language`, but clicking it silently does nothing.

### 2. Dual source of truth — fragile by design

`entitiesConfig.jsx` already has a "sortable" flag (4th element of each column definition) that controls whether sort arrows are displayed. `ALLOWED_SORT_FIELDS` duplicates that information. Every future addition or change to a sortable column must now be made in two places — as proven by the bug above.

**Suggested fix:** derive the whitelist dynamically from `entitiesConfig` instead of hardcoding it:

```js
const ALLOWED_SORT_FIELDS = Object.fromEntries(
  Object.entries(entitiesConfig).map(([type, config]) => [
    type,
    new Set(config.columns.filter(col => col[3]).map(col => col[5] ?? col[1]))
  ])
);
```

This eliminates the duplication and the bug.

### 3. Silent failure

```js
if (allowedFields && !allowedFields.has(property)) return; // no feedback
```

The user clicks a column header and nothing happens — no feedback, no `console.warn`. Hard to debug.

### 4. Permissive fallback for unknown entity types

```js
if (allowedFields && !allowedFields.has(property)) return;
//  ^^^^^^^^^^^^^^^^ if entityType is not in the map → everything is allowed
```

If a new entity type is added without a corresponding entry in `ALLOWED_SORT_FIELDS`, the whitelist is silently bypassed. A proper whitelist should deny by default. This weakens the intended defense-in-depth.

---

## Summary

| # | Severity | Issue |
|---|---|---|
| 1 | **Bug** | `language` missing from `ALLOWED_SORT_FIELDS.documents` |
| 2 | **Design** | Dual source of truth between `entitiesConfig` and the whitelist |
| 3 | Minor | Silent failure with no log |
| 4 | Minor | Permissive for unlisted entity types |

ClemRz:
Thanks @Paul-AUB — all 4 items addressed in `82d19960`:

1. **`language` missing from documents** — fixed (now derived automatically, see #2).
2. **Dual source of truth** — `ALLOWED_SORT_FIELDS` is now derived from `entitiesConfig` columns at module level:
   ```js
   const ALLOWED_SORT_FIELDS = Object.fromEntries(
     Object.entries(entitiesConfig).map(([type, config]) => [
       type,
       new Set(config.columns.filter(col => col[3]).map(col => col[5] ?? col[1]))
     ])
   );
   ```
   Single source of truth — no more manual sync.
3. **Silent failure** — added `console.warn` with the blocked field and entity type.
4. **Permissive fallback** — guard now denies by default when `entityType` has no entry in the derived map (`!allowedFields || !allowedFields.has(property)`).

Also added the `language` column to the `documents` entity config (hidden by default, sortable) to match the API fix from GrottoCenter/grottocenter-api#1528.

Lint ✅ Tests ✅ (88/88)

Paul-AUB (approved):

PR #1246, opened 2026-03-31
feat: display data quality in entrance page

### LINKED ISSUE #845: [Entrance view] add data quality informations
Désormais la qualité des données de chaque cavité est évaluée et affichée dans une page accessible depuis les pages des massifs et des pays. Il serait très utile que cette information figure aussi sur la page de chaque cavité

Ajout du même bouton que dans la page des entrances d'un pays / massif (au niveau du nom de la cavité ?)

Paul-AUB:
# Display data quality in entrance page)

## 🤔 What

Display the data quality in each entrance page:

image

image

image

## 🤷‍♂️ Why

Uses data from https://github.com/GrottoCenter/grottocenter-api/pull/1509

## 🔍 How

_Describe how we're implementing the change.  
Feel free to dive into technicalities!_

## 🔗 Related API PR

https://github.com/GrottoCenter/grottocenter-api/pull/1509

## 🧪 Testing

Visual testing (desktop + mobile)


ClemRz (changes requested):
1. **Storybook translations stale** — `.storybook/en.json` still has the old key `"General data on the cave"` and is missing `"Details by category"` and `"General data"`. This will cause missing-translation warnings in any Storybook story that renders `DataQualityComputeTable` or `Properties`.

2. **Duplicated `dataQuality` prop type** — The shape is defined both in `EntrancePropTypes` (`entrance.type.js`) and again in `Properties.propTypes`. Since `Properties` already imports `EntrancePropTypes`, the local copy could reference the shared type to avoid drift.

3. **`Comment` → `Comments` label change in `DataQualityComputeTable`** — The row name was changed from `'Comment'` to `'Comments'`, which is used as a translation key via `formatMessage({ id: row.name })`. The `"Comments"` key exists in `en.json` so it works, but it's a subtle inconsistency with the API field name (`comment`). Verify all 15 language files have a `"Comments"` key.

4. **Accessibility: expand/collapse toggle not keyboard-reachable** — The "Details by category" toggle uses a `Box` with `onClick` as the interactive element, and the inner `IconButton` has `tabIndex={-1}`. This makes the toggle unreachable via keyboard or screen reader. Consider using `ButtonBase` with `aria-expanded`, or remove `tabIndex={-1}` from the `IconButton`.

5. **`score == null` handles `0` correctly** — Loose equality lets `0` scores render, which appears intentional. Just confirming.

6. **`lastComputedAt` unused** — Defined in prop types but never displayed. Either omit it or add a comment noting it's for future use.

Paul-AUB:
1, 2, 4, 6 fixed
3, 5 in purpose

ClemRz (approved):

PR #1260, opened 2026-04-03
Feature/map improvements

### LINKED ISSUE #1242: [MAP] Améliorations
Les utilisateurs découvrent les nouvelles fonctionnalités associées à la carte et des demandes sont apparues

1- Les services associés à la carte ne semblent pas directement accessibles aux utilisateurs, qu'il faut les guider pour qu'ils les trouvent
> Est ce qu'il serait possible, sur les grands écrans que le menu soit ouvert par défaut ?

Image

2- les boutons de taille différente sont appréciés mais il y a une demande qui a été formulée concernant les critères qui sont utilisés pour classer les entrées
> Est ce qu'il serait possible de donner accès a cette information ?

Image
Peut être pas dans le menu car cela va impacter la lisibilité, peut être comme pour les critères de qualité ?

Paul-AUB:
# Map improvements — multiple layers, guided tour, cave size popover

closes #1242 

## Summary

Three UX improvements to the main map (`MapClusters`):

1. **Multiple heat layers simultaneously** — users can now enable entrances, networks, and massifs heatmaps at the same time (previously a single radio selection). Each layer gets its own Leaflet pane with a fixed z-index and a slight SVG pixel offset so hexagons from different types don't perfectly overlap. Opacity is automatically reduced when more than one layer is active.

image
image

2. **Guided tour** — an auto-launching step-by-step tour highlights the Data control and Map layers controls on first visit. The tour uses `localStorage` (permanent "don't show again") and `sessionStorage` (per-session suppression). A reusable `AppTour` / `TourTooltip` system under `src/components/common/Tour/` can be extended to other pages.

image
image

3. **Cave size thresholds popover** — an info icon next to the "Filter by size" section opens a popover explaining the depth/length thresholds that define small, medium, and large caves.

image

image

### New files
- `src/components/common/Tour/AppTour.jsx` — generic `TourProvider` wrapper
- `src/components/common/Tour/TourTooltip.jsx` — MUI-based step tooltip (title, body, prev/next/finish, "don't show again" checkbox)
- `src/components/common/Maps/MapClusters/MapTour.jsx` — map-specific tour steps
- `src/components/common/Tour/README.md` — how to add tours to new pages
- `src/components/common/Maps/MapClusters/README.md` — map cluster architecture notes

### Modified files
- `DataControl.jsx` — radio → checkboxes for heat layers; cave size info popover; `data-tour` attribute on toggle button
- `useHeatLayer.js` — manages three independent hexbin layers (one per type) with per-type panes, z-indexes, SVG offsets, and shared `updateLayers()` API
- `index.jsx` — multi-select heat state (`Set`); tour auto-launch logic; `isMassifPolygonMode` state; `setVisibleMarkersStable` optimization
- `constants.js` — `CAVE_SIZE_THRESHOLDS`, `HEX_MULTI_OPACITY`, `HEX_MAX_RADIUS` exported; `getHeatOffZoom` moved here
- All lang files — new keys for tour labels and cave size threshold messages

## How to test

- Open the map. The guided tour should launch automatically on first visit.
- Check "don't show again" on the last step — the tour should not reappear after a full page reload.
- To reset: run in the browser console:
  ```js
  localStorage.removeItem('mapTourSeen_v1');
  sessionStorage.removeItem('mapTourSeenThisSession_v1');
  ```
- Enable two or more heat layers at the same time and verify they render side by side with reduced opacity and slight offsets.
- Zoom in past the markers threshold — heat layers should disappear and point markers should appear for each active type.
- Zoom in past level 8 — massif heatmap should be replaced by GeoJSON polygons (if "massifs" is checked).
- Open the "Filter by size" section (zoom in first) and click the ⓘ icon — the threshold popover should appear.

## Notes for reviewers

- `REACT_APP_DISABLE_MAP_TOUR=false` is committed to `.env`. Set `REACT_APP_DISABLE_MAP_TOUR=true` in `.env.local` to suppress the auto-launch during development without touching storage.
- `heatmapTypes.NONE` has been removed from the enum — it was never reachable with the new multi-select model.


urien:
La visite guidée est vraiment super,
J'ai tout essayé et cela fonctionne bien sauf 1 truc,
Quand on va sur un massif le nom n'apparait pas et il n'est pas possible de faire apparaitre le nom on de cliquer sur un pop up pour accéder à la page
image

urien:
Désormais le pop up fonctionne mais quand je clique sur le nom du massif il y a une erreur
https://orange-rock-0d4f87503-1260.westeurope.azurestaticapps.net/ui/massifs/2684
En fait cela ne fonctionne pas toujours. e n'arrive plus a y acceder
image

Par contre les liens fonctionnent pour les entrées, réseaux et organisations

Paul-AUB:
Oui je viens de voir que les tests e2e ont détecté l'erreur. Je corrige ça ce soir

Paul-AUB:
Les deux problèmes suivants sont corrigés:
- affichage de la page d'un massif
- info sur la carte principale au survol et clic d'un massif

Tu devrais pouvoir retester :)

ClemRz (changes requested):
## Issues

### 1. `issue.md` should not be committed

There's a new `issue.md` file at the repo root that duplicates the PR description. This looks like a working document that shouldn't be in the codebase. Remove it before merging.

### 2. MassifPolygons pane was added then reverted — PR description is stale

The PR description says "MassifPolygons.jsx — dedicated Leaflet pane (z-index: 350) so massif polygons always render below entrance/network markers" and the notes section mentions "The massif polygon pane (z-index: 350) sits below Leaflet's default overlay pane (400)." But the last fixup commit (`fixup massifs popup on map`) reverts that change entirely. The final `MassifPolygons.jsx` has no custom pane. The PR description should be updated to reflect the actual final state, or the pane should be re-added if it was intentionally part of the design.

### 3. MapClusters README has stale storage key names

The `MapClusters/README.md` references `mapTourSeen` and `mapTourSeenThisSession` (without the `_v1` suffix), but the actual code uses versioned keys like `mapTourSeen_v1` and `mapTourSeenThisSession_v1`. The README's reset instructions won't work as written.

### 4. Duplicate heat selection state in `DataControl` and `index.jsx`

Both `DataControl` and `HydratedMap` (in `index.jsx`) independently maintain a `selectedHeats` `Set`. `DataControl` has its own `useState(new Set([heatmapTypes.ENTRANCES]))` and syncs changes upward via the `updateHeatmap` callback + a `useEffect` that diffs against `prevHeatsRef`. Meanwhile `HydratedMap` also has `selectedHeats` state that it updates in `handleUpdateHeat`. This dual-state approach is fragile — if they ever drift out of sync, the UI checkboxes won't match the actual layer visibility. Consider lifting the state fully to `HydratedMap` and passing `selectedHeats` down as a prop, or removing the local state from `DataControl` entirely.

### 5. `framer-motion` added as a production dependency

`framer-motion` (^11.0.0) was added to `dependencies`. This is a substantial bundle-size addition. Is it actually used directly, or is it only a peer dependency of `@reactour/tour`? If it's only a transitive peer dep, it might be worth noting that in the PR. If it's not used at all, it shouldn't be added.

### 6. Untranslated Spanish strings

The following keys in `es.json` are still in English. Proposed translations:

```json
"cave size large threshold": "profundidad ≥ {depth}m o longitud ≥ {length}m",
"cave size medium threshold": "profundidad ≥ {depth}m o longitud ≥ {length}m",
"cave size small threshold": "profundidad < {depth}m y longitud < {length}m",
"Open guided tour": "Descubrir el mapa",
"Tour - Back": "Atrás",
"Tour - Data control": "Control de datos",
"Tour - Data control description": "Activa o desactiva las capas de datos: entradas, redes, macizos y organizaciones.",
"Tour - Do not show again": "No volver a mostrar",
"Tour - Finish": "Finalizar",
"Tour - Map layers": "Capas del mapa",
"Tour - Map layers description": "Cambia el mapa base y activa las capas superpuestas.",
"Tour - Next": "Siguiente",
"Tour - Skip": "Omitir"
```

### 7. `updateLayers` has an empty dependency array but closes over `formatMessage`

In `useHeatLayer.js`, the `useEffect` that creates all three hex layers calls `formatMessage` inside the hover handler, but the dependency array is `[]`. If the locale changes at runtime, the tooltip text will remain in the original language. This is a minor issue since locale changes typically trigger a full re-render, but it's worth being aware of.

### 8. `updateLayers` callback also has `[]` deps but is safe — just document why

The `updateLayers` `useCallback` has `[]` dependencies, which is correct since it only references refs. But the eslint-disable comment is on the `useEffect` above it, not on `updateLayers` itself. Consider adding a brief comment explaining why the empty deps are intentional.

## Minor / Nits

- The `setVisibleMarkersStable` comparison uses strict element-by-element ordering (`prev.every((v, i) => v === next[i])`). This is correct for the current usage but could be surprising if the order of items in the array ever varies. A `Set`-based comparison might be more robust.

- In `DataControl`, the `useEffect` that syncs `selectedHeats` → `updateHeatmap` fires on every render where `selectedHeats` changes, iterating all three types to find the diff. This works but is slightly over-engineered — since `handleHeatToggle` only ever toggles one type at a time, you could pass the changed type directly instead of diffing.

- The `MapTour` steps `useMemo` depends on `[formatMessage]` which is correct, but the Tour README example shows `// eslint-disable-next-line react-hooks/exhaustive-deps` with `[]` — the example should match the actual pattern used.

Paul-AUB:
All fixed !

ClemRz (approved):
Great job! TY!

urien (approved):
Du côté utilisateur tout fonctionne.
Merci @Paul-AUB

PR #1264, opened 2026-04-08
feat(search): improve advance search table UX

### LINKED ISSUE #1219: [Recherche avancée] Améliorations proposées
La recherche avancée a été complétement revue suite à la mise en place d'un nouveau moteur de recherche. Merci @vmarseguerra pour ce très beau travail.
A l'usage des propositions d'éméliorations sont proposées

demandes formulées par  @BTHOM 

1 - La colonne "Id" pourrait elle être triable, ("sortable"), à l'instar des autres colonnes du tableau de présentation des résultats ?
2 - Les dates de "création" et de "dernière modification" pourraient elle faire partie des colonnes affichables et triables ?
3 - Le pavé de recherche "Propriétés du réseau" pourrait il être remonté en seconde position, juste après le pavé "Localisation" ?

Paul-AUB:
# feat(search): improve advance search table UX

## 🤔 What

- Add a **"Go to page"** input field in the table pagination footer for quick direct navigation
- Add **first/last page buttons** to pagination
- Fix sort API field mapping: use the correct API field name (e.g. `numericId`) instead of the display field name when sending sort parameters
- Add `dateLastModif` (Last modified) and `numericId`-sortable `id` column for the entrances entity
- Add i18n keys: `Go to page`, `Last modified`; remove unused `Last reviewed`

## 🤷‍♂️ Why

The advance search table was missing convenient pagination controls for large result sets. The sort field mapping bug caused incorrect sort parameters to be sent to the API when a column had a separate API field name (e.g. `numericId`). Adding more sortable/visible fields improves discoverability.

## 🔍 How

- Introduced `SORT_FIELD_MAP` (replaces `ALLOWED_SORT_FIELDS`), mapping each column's display field → API sort field, so `handleSortRequest` now resolves the correct `apiField` before calling `onSortChange`.
- Added a `JumpToPage` controlled component rendering a small `TextField` + label; shown on non-mobile viewports inside the pagination row.
- Enabled `showFirstButton` / `showLastButton` on `StyledTablePagination`.
- Extended `entrances` config in `entitiesConfig.jsx` with `dateLastModif` and marked `id` as sortable via `numericId`.

## 🔗 Related API PR

See https://github.com/GrottoCenter/grottocenter-api/pull/1533

## 🧪 Testing

1. Open any advance search result table with multiple pages.
2. Verify the "Go to page" input jumps to the correct page on Enter/blur.
3. Verify first/last page buttons work.
4. Sort by a column that has a separate API field (e.g. entrance Id) and confirm the request uses `numericId:asc` not `id:asc`.
5. Check the "Last modified" column appears and sorts correctly for entrances.

## 📸 Previews

image

image


urien (approved):
Merci @Paul-AUB tout fonctionne parfaitement et répond a la demande formulée
@BTHOM voila la réponse a ta demande. Peux tu vérifier que cela te convient ?

ClemRz (changes requested):
## 1. `JumpToPage` doesn't recalculate `inputValue` when `count`/`rowsPerPage` change

`totalPages` is computed on every render (not memoized), so that's fine. But `inputValue` is only synced to `page` via the `useEffect`, not to `count` or `rowsPerPage`. If the user types "5", then the result set shrinks to 2 pages, the input will still show "5" until `page` itself changes. Consider adding `count` and `rowsPerPage` to the effect or clamping in the effect:

```jsx
useEffect(() => {
  const maxPage = Math.max(1, Math.ceil(count / rowsPerPage));
  setInputValue(String(Math.min(page + 1, maxPage)));
}, [page, count, rowsPerPage]);
```

## 2. MUI deprecation: `inputProps` → `slotProps.htmlInput`

With MUI 7, `inputProps` on `TextField` is deprecated in favor of `slotProps.htmlInput`. Since the project is on MUI 7, this should use the new API:

```jsx
<TextField
  slotProps={{
    htmlInput: {
      inputMode: 'numeric',
      pattern: '[0-9]*',
      style: { textAlign: 'center', padding: '2px 6px' },
      'aria-label': formatMessage({ id: 'Go to page' })
    }
  }}
  ...
/>
```

## 3. `SORT_FIELD_MAP` includes non-sortable entity types

Minor, but `placeholder`, `notifications`, `csvImportEntrances`, `csvImportDocuments`, and `duplicate` all have zero sortable columns, so their entries in the map are empty objects. Not a bug, just noise. Could filter them out or just leave it — it's harmless.

## 4. The column tuple format is getting unwieldy

The column config is a positional array where index 0 = visible, 1 = field, 2 = label, 3 = sortable, 4 = renderer, 5 = apiField. With the new `numericId` mapping, some entries now have `undefined` placeholders (`[false, 'id', 'Id', true, undefined, 'numericId']`). This is fragile and hard to read. Not a blocker for this PR, but worth considering a migration to named objects in a follow-up:

```jsx
{ visible: false, field: 'id', label: 'Id', sortable: true, apiField: 'numericId' }
```

Paul-AUB:
# Fix notes

1. `JumpToPage`: `inputValue` was only synced to `page`; now clamps to `totalPages` when `count`/`rowsPerPage` change
2. `inputProps` → `slotProps.htmlInput` (deprecated in MUI 7)
3. `SORT_FIELD_MAP`: entity types with no sortable columns are now excluded
4. Column tuple positional arrays replaced with named objects (`{ visible, field, label, sortable, render?, apiField? }`)

ClemRz (changes requested):
## Bug: `entityColumnsModifier` caller still uses old array format

`packages/web-app/src/pages/Contributions/index.jsx` (line 47-48) still accesses columns with the old array-index convention:

```jsx
entityColumnsModifier={columns => {
  const c = columns.find(e => e[1] === 'dateInscription');
  c[0] = true;
}}
```

Columns are now objects with `{ visible, field, label, ... }`, so this should be:

```jsx
entityColumnsModifier={columns => {
  const c = columns.find(e => e.field === 'dateInscription');
  c.visible = true;
}}
```

As-is, `c` will be `undefined` (no column has an index `[1]` property equal to `'dateInscription'`), and the next line will throw a TypeError at runtime. This is a **breaking bug** on the Contributions page.

---

## Stale PropTypes

Two PropTypes still declare `PropTypes.arrayOf(PropTypes.array)` even though columns are now arrays of objects:

- `EntityTableHead.propTypes.columns` (line ~140)
- `VisibleColumnsMenu.propTypes.columns` (line ~199)

Both should be `PropTypes.arrayOf(PropTypes.shape({}))` (or a more specific shape) to match the new object format. Not a runtime break, but it will produce console warnings in development.

---

## `VisibleColumnsMenu` mutates column objects in place

```jsx
column.visible = !column.visible; // eslint-disable-line no-param-reassign
const newColumns = [...columns];
```

The mutation happens before the shallow copy, so the original object in the `entitiesConfig` module-level constant gets mutated. The spread `[...columns]` creates a new array but the elements are the same object references. This was already the case with the old array format (`column[0] = !column[0]`), so it's a pre-existing issue, not a regression — but worth noting since the refactor was a good opportunity to fix it:

```jsx
const newColumns = columns.map(c =>
  c === column ? { ...c, visible: !c.visible } : c
);
```

---

## JumpToPage: `useEffect` dependency list suppresses a warning but is incomplete

```jsx
useEffect(() => {
  setInputValue(String(Math.min(page + 1, totalPages)));
}, [page, count, rowsPerPage]); // eslint-disable-line react-hooks/exhaustive-deps
```

`totalPages` is derived from `count` and `rowsPerPage`, so the deps are technically sufficient for correctness. But the eslint-disable is there because `totalPages` is referenced inside the effect but not listed. Since `totalPages` is a derived value, either add it to the array (it's stable when its inputs are stable) or inline the computation. Minor nit.

---

## JumpToPage: no debounce or guard on rapid Enter presses

Each Enter press calls `onPageChange` immediately, which dispatches a network request. Rapid key presses will fire multiple requests. Low severity for a "go to page" input, but worth considering a small debounce or disabling the input while loading.

Paul-AUB:
# Fix notes

## 1. `entityColumnsModifier` crash on Contributions page

**`pages/Contributions/index.jsx`**

Columns are now objects but the callback still used array-index access — `find(e => e[1] === 'dateInscription')` always returned `undefined`, throwing a `TypeError` at runtime.

```js
// before
const c = columns.find(e => e[1] === 'dateInscription');
c[0] = true;
// after
const c = columns.find(e => e.field === 'dateInscription');
c.visible = true;
```

## 2. Stale PropTypes warnings

**`EntityTable.jsx` — `EntityTableHead` & `VisibleColumnsMenu`**

Both still declared `PropTypes.arrayOf(PropTypes.array)` instead of `PropTypes.arrayOf(PropTypes.shape({}))`.

## 3. `VisibleColumnsMenu` mutated column objects in place

**`EntityTable.jsx`**

Direct mutation before the spread meant the `entitiesConfig` module-level objects were permanently altered.

```js
// before
column.visible = !column.visible;
const newColumns = [...columns];
// after
const newColumns = columns.map(c =>
  c === column ? { ...c, visible: !c.visible } : c
);
```

## 4. `JumpToPage` — incomplete `useEffect` deps

**`EntityTable.jsx`**

Effect referenced `totalPages` but listed `count` and `rowsPerPage` as proxies, requiring an `eslint-disable`. Simplified to `[page, totalPages]`.

## 5. `JumpToPage` — no guard on rapid Enter presses

**`EntityTable.jsx`**

Rapid Enter presses fired multiple `onPageChange` calls. Added a `pending` flag: set to `true` on commit, reset when `page` actually changes. The input is also `disabled` while pending.

ClemRz (approved):

PR #1266, opened 2026-04-09
fix(network): improve eye tooltip in entrances list

Paul-AUB:
# fix(network): improve eye tooltip in entrances list

## 🤔 What

When only one entrance is visible on the map (all others are hidden), the eye icon tooltip on that last visible entrance now reads **"Show all on map"** instead of "Hide from map".

## 🤷‍♂️ Why

When a single entrance remains selected, clicking its eye doesn't hide it — it deselects the last filter and reverts to showing all entrances. The previous "Hide from map" label was misleading. "Show all on map" accurately describes the action.

## 🔍 How

Added a condition in `EntrancesList.jsx`: when `selectedEntrancesId.length === 1 && isSelected`, use the `"Show all on map"` message key instead of `"Hide from map"`. All 15 lang files updated with proper translations for each language.

## 🧪 Testing

On a network page with multiple entrances:

1. Click the eye on one entrance to hide it — tooltip should show "Hide from map"
2. Keep clicking until only one entrance remains visible — its tooltip should now show "Show all on map"
3. Click it — all entrances should reappear on the map

## 📸 Previews

_N/A_


urien (approved):
Pour moi tout fonctionne bien, c'est très agréable à utiliser comme outil

ClemRz (approved):

PR #1269, opened 2026-04-20
1168 signup email verification interface 

dawoldo:
# Signup Email Verification Interface

## 🤔 What

This PR adds the necessary frontend infrastructure to handle the new email verification lifecycle for Grottocenter accounts.

- **Redesigned Signup Success**: The signup process now informs users to check their email for an activation link instead of prompting for immediate login.
- **Verification Landing Page**: Created a new route at `/ui/verify-email` that extracts activation tokens from query parameters and communicates with the backend to activate the account.
- **Unverified State Handling**: 
    - Login attempts on unverified accounts now trigger a specialized warning view within the Login Dialog.
    - Forgot-password requests for unverified accounts similarly redirect users to the verification warning.
- **Resend Utility**: Integrated a privacy-focused "Resend verification email" feature that uses `POST` with the email in the request body and always reports success to prevent email enumeration.
- **Internationalization**: Added complete translation sets for all new UI elements in English, French, and Spanish.

## 🤷‍♂️ Why

The Grottocenter API now requires an email verification step before an account is considered "active." These changes ensure the frontend is synchronized with this requirement, providing clear user feedback and a path to resolution for unverified users.

## 🔍 How

- **State Management**: 
    - Introduced `VerifyEmailReducer` and `ResendVerificationEmailReducer` to manage the lifecycle of verification requests.
    - Updated `LoginReducer` to handle a new `isNotVerifiedMessageDisplayed` state.
    - **Actions**: Implemented specialized thunks in `actions/VerifyEmail.js` and `actions/ResendVerificationEmail.js` following the `VERIFY_EMAIL` and `RESEND_VERIFICATION` naming patterns.
- **Error Handling**: Unified error display via the global `ErrorHandler.js`.

## 🔗 Related API PR

[](https://github.com/GrottoCenter/grottocenter-api/pull/1534)

## 🧪 Testing

1. **New User Signup**: Complete the signup form and verify the new instruction message appears.
2. **Unverified Login**: Attempt to log in with an unverified account. Confirm the warning icon and "Resend" button appear.
3. **Resend Flow**: Verify that clicking "Resend" shows a loading spinner and eventually a success snackbar, regardless of whether the email technically exists.
4. **Verification Page**: Manually navigate to `/ui/verify-email?token=<TOKEN>` and verify the success/failure states are rendered.

## 📸 Previews

N/A


ClemRz (changes requested):
# PR 1269 Review — Issues

## Requirement 2: Email Verification Page

- **AC 4** — The error message string says "invalid or expired" but the backend has no expiration concept — it only returns not-found. The translation keys in en/fr/es all say "or expired" which is misleading. Should say just "invalid."
- **AC 5** — No distinct "already verified" state. The `VerifyEmailPage` component only has three branches: `loading`, `success`, and error. The `message` prop is passed through but never used to differentiate "newly verified" from "already verified." Both cases render the same success message. The spec requires a distinct message.
- **AC 6** — Missing token validation. When the `token` query parameter is absent, the component never dispatches the action and falls through to the error branch only because the initial reducer state has `success: false` and `error: null`. This works by accident, not by design. Should explicitly detect a missing token and show a clear "invalid link" message.

## Requirement 5: Forgot-Password Error Handling for Unverified Accounts

- **AC 1** — The `postForgotPassword` action does check for `NotVerified` and transitions to the "not verified" view, but the message shown is the generic "Your account is not verified yet / Please check your email to activate it." It doesn't say "verify your email before resetting your password." The user can't tell whether their login or their password reset failed since both show the same screen.

## Requirement 6: API Route Definitions

- **AC 2** — Endpoint name mismatch. The spec says `resendVerificationUrl` for `POST /api/v1/resend-verification`. The PR defines `resendVerificationEmailUrl` pointing to `/api/v1/resend-verification-email`. The backend spec defines the endpoint as `POST /api/v1/resend-verification`, not `resend-verification-email`. This will cause a 404 at runtime unless the backend also uses the `-email` suffix.

## Requirement 8: Internationalization

- **AC 3** — No distinct translation key for the forgot-password unverified scenario — it reuses the same generic message as login.
- **AC 4** — Missing translation keys:
  - `"Email Verification"` used as page title in `VerifyEmailPage` but not in en.json.
  - No "already verified" distinct message key.
  - `"You can request a new verification email for:"` used in `Login.jsx` but not in en.json.

## Requirement 9: Routing

- **AC 1** — The route is `/ui/verify-email`, but the backend sends verification emails with links in the format `{frontendBaseUrl}/verify-email?token={activationCode}` (no `/ui/` prefix). This will cause a 404 when users click the link in their email. Needs alignment with the backend.

## Additional Issues

1. **HTTP method mismatch on verify-email.** The action uses `method: 'POST'`, but the spec and backend define the endpoint as `GET /api/v1/verify-email?token={token}`. Will fail at runtime if the backend only accepts GET.
2. **`Content-Type` header missing on resend POST.** `ResendVerificationEmail.js` sends `JSON.stringify({ email })` but doesn't set `Content-Type: application/json`. The backend will likely fail to parse the body.
3. **Double body read in forgot-password error handling.** In `Login.js`, the catch block does `await response.json()` and the inner fallback does `await response.text()`. A response body can only be consumed once — the second read will throw.
4. **Container placement.** `VerifyEmail.jsx` lives in `components/appli/` but acts as a Redux-connected container. Project conventions place containers in `src/containers/`.

ClemRz (changes requested):
Please sort the translation files using scripts/translations/sort.js

dawoldo:
> **AC 5** — No distinct "already verified" state. The `VerifyEmailPage` component only has three branches: `loading`, `success`, and error. The `message` prop is passed through but never used to differentiate "newly verified" from "already verified." Both cases render the same success message. The spec requires a distinct message.

This is currently a technical limitation of the API. When a user is successfully verified, the activation code is removed from the database. Consequently, if a user clicks the link again, the backend returns a 404/Error because the code no longer exists (which is the same outcome for invalid activation codes, since we can't associate an entry in the db to the code that was inputted).
So from the frontend's perspective, there is no way to distinguish between a 'code that was already used' and a 'code that never existed.'

dawoldo:
> ## Requirement 6: API Route Definitions
> 
>     * **AC 2** — Endpoint name mismatch. The spec says `resendVerificationUrl` for `POST /api/v1/resend-verification`. The PR defines `resendVerificationEmailUrl` pointing to `/api/v1/resend-verification-email`. The backend spec defines the endpoint as `POST /api/v1/resend-verification`, not `resend-verification-email`. This will cause a 404 at runtime unless the backend also uses the `-email` suffix.

Backend spec defines the route as ```/api/v1/resend-verification-email```. 

> Requirement 9: Routing
>
>    AC 1 — The route is /ui/verify-email, but the backend sends verification emails with links in the format {frontendBaseUrl}/verify-email?token={activationCode} (no /ui/ prefix). This will cause a 404 when users click the link in their email. Needs alignment with the backend.

Just like every other route of the app it uses the generic ```API_BASE_PATH``` string to build the route url. The prefix to all routes of the backend is ```/api/v1/``` not ui (?) which is included in the ```API_BASE_PATH``` constant.

ClemRz (changes requested):
### ForgotPassword.js — Double body read

The error handler in `ForgotPassword.js` (`src/actions/ForgotPassword.js`) calls `await response.json()`, and in the inner catch falls through to `await response.text()`. The response body stream is already consumed by `.json()`, so `.text()` will throw. The `Login.js` version was fixed (reads `.text()` once, then `JSON.parse`), but `ForgotPassword.js` still has the old pattern.

### ForgotPassword.js — Dispatches `fetchForgotPasswordSuccess()` on NotVerified

When the forgot-password endpoint returns 401/NotVerified, the handler dispatches `fetchForgotPasswordSuccess()` to clear loading state. But this sets `success: true` in the ForgotPasswordReducer, which is semantically wrong — the request didn't succeed. A dedicated reset action or a neutral state transition would be cleaner.

ClementRonzon-loanpro (approved):

ClemRz (approved):

ClemRz:
Closing this since it was delivered through PR 1274. The current one has an issue with commits being squashed.

PR #1273, opened 2026-04-22
feat(coordinates): multi-projection coordinate system

### LINKED ISSUE #265: [ENTRANCE Create] possibility to convert coordinates available in another coordinate system
Ajouter un bouton qui ouvre le convertisseur de coordonnées existant. La validation des coordonnées obtenues dans cette fenetre permettant de remplir les coordonnées dans la fenetre de création d'une entrée

A mettre en relation avec https://github.com/GrottoCenter/grottocenter-front/issues/878

Actuellement sur la page de création ou d'édition d'une cavité, les informations sur la localisation sont fournies de cette façon
image
On pourrait imaginer faire quelque chose comme cela
image

### LINKED ISSUE #878: [ENTRANCE ORGANISATION VIEW] Affichage des coordonnées dans le système de coordonnées de son choix
Bonjour,
L'affichage actuel des coordonnées WGS84 en degré décimal:

    Mascate: 23.494686, 58.796286
    Pozo Colorado: -23.494686, -58.796286

La communauté Wikipédia a adopté une notation différente
- Mascate :           23° 35′ 26″ nord, 58° 24′ 35″ est      
- Pozo Colorado : 23° 29′ 44″ sud,  58° 47′ 35″ ouest    

D'autres utilisateurs peuvent préférer un autre système de coordonnées

- L'indication du système de coordonnées actuel est remplacé par un bouton sur lequel le système est indiqué

Image

- le fait de cliquer sur ce bouton ouvre un menu qui permet de choisir un autre système de coordonnées
-  Les coordonnées sont alors affichées dans le système de coordonnées retenu
- Le système de coordonnées retenu est stocké et utilisé à chaque fois que l'utilisateur affiche des coordonnées

### LINKED ISSUE #1145: Carte dans les pages de création et de modification d'entrées et d'organisation
Dans le module de “correction – modification”, et seulement dans ce module (en visite, la carte est très confortable), la fenêtre proposée est trop petite, cela on peut le contourner en mode plein écran, mais le problème est que cette petite fenêtre est grandement masquée par le bouton “Convertir” (mal placé), bouton qui empêche d’accéder aux différentes cartes proposées qui elles sont très intéressantes.

Donc, agrandir cette fenêtre de modification OU/ET enlever le bouton “Convertir”.
Par contre, très bien de faire glisser l’ensemble de la carte sous le pointeur qui reste fixe. C’est plutôt efficace !

Cette demande a été formulée par Patrick LE ROUX
Elle me semble très pertinente

### LINKED ISSUE #1178: [MAP] Update the converter to improve the UX
Le convertisseur sur la carte est vraiment "archaïque", et il faut un moment pour comprendre exactement où sont les champs à remplir et ceux à lire.

Image

---

Image

(j'ai pris un des pire cas de figure, j'avoue)

Il prend bcp de place, et sur mobile c'est presque inutilisable:

Image

- Avoir un rendu moderne
- Debounce les coordonnées pour se passer du bouton de conversion
- Identifier facilement les champs d'input et les champs d'output
- Garantir une bonne utilisabilité sur mobile, avec une occupation de l'écran maitrisée
- Pouvoir copier les coordonnées d'output en 1 clic
- (internal) Au passage réduire le nombre de composants utilisés, cleanup, etc.
- (internal) Utiliser des adresses en https pour les liens tant qu'à faire

Exemple de rendu possible (WIP):

Image

Image

Paul-AUB:
# feat(coordinates): multi-projection coordinate system

closes #878 (main feature)
closes #265 (allow to select each time any system to enter the coordinates)
closes #1178 (converter is no more usefull, user can choose each time how he wants display the data, and can also choose how to enter the data).
closes #1145 (converter deleted, and edit page is no more a modale, which allow more space)

## 🤔 What

Adds a full multi-projection coordinate system to the app:

- **`CoordinateDisplay`** — new common component that shows coordinates in the user's preferred system (WGS84 DD, DMS, or any server-side projection), with one-click copy and links to OSM / Google Maps / Grottocenter map.
- **`CoordinateFormSection`** — new form section for coordinate input supporting real-time conversion from any projection back to WGS84, including UTM zone/hemisphere fields and a live WGS84 preview.
- **`coordinateConvert.js`** — pure helper module: DMS ↔ decimal, proj4 projection conversions, UTM zone computation, clipboard formatting.
- **`useCoordinatePreference`** — hook that persists the user's preferred CRS to `localStorage` and syncs all mounted components instantly.
- **`useProjections`** — hook that lazy-fetches the server projection list from Redux.
- **`CRSMenu`** — shared dropdown component used by `CoordinateDisplay`, `CoordinateFormSection` and the map context menu.
- **Map right-click context menu** — right-clicking the map now shows the clicked point's coordinates in the preferred CRS, with a copy button and CRS switcher.
- **Entrance edit page** — editing an entrance now navigates to a dedicated `/ui/entrances/:id/edit` page instead of opening a dialog.
- Removes the old `Converter` component (544 lines) and `StatusMessage` component.

## 🤷‍♂️ Why

Coordinates were previously always displayed as raw WGS84 decimal degrees. Many cavers work in national grid systems (Lambert, UTM, RGF93…). This change lets users pick their preferred system once and have it applied consistently across the app — display pages, forms, and the map.

## 🔍 How

- `proj4` handles all CRS transformations. The server exposes available projections; `useProjections` fetches and caches them in Redux.
- `useCoordinatePreference` stores the active CRS code in `localStorage` and propagates changes to all listeners via a module-level `Set`, avoiding a Redux dependency for pure UI preference.
- `CoordinateFormSection` writes only WGS84 values to the react-hook-form fields; all CRS conversion is local state. A `selfSetRef` prevents feedback loops when the map drags the marker (external WGS84 change) vs. the user typing in a projection field.
- `CRSMenu` is extracted as a shared component so the same grouped/sorted projection list renders identically in the display component, the form, and the map context menu.

## 🧪 Testing

1. Open any entrance page — coordinates should display in WGS84 DD by default.
2. Click the **⚙ Tune** icon → select DMS or a national projection → value updates immediately and persists on reload.
3. Open entrance edit → change CRS in the form → type projected coordinates → verify WGS84 preview updates and the form submits valid lat/lng.
4. Drag the map marker → projected fields should resync to the new position.
5. Right-click the map → verify coordinate popup shows the clicked point and the copy button works.

## 📸 Previews
image

Preference for coordinates system is savec in CRS (memory in navigator)

image

image

image

Right click on the map to get point coordinates :

image

Create/edit page is always WGS84 by default (no memory), but can be changed ponctually :

image

image

Edit Entrance is now a full page rather than a modale (more space, more responsive) :

image

Unified notification system with better color platette:

image

image


Paul-AUB:
Note : with the new contextual menu in the map, it *could* be simpler to implement then https://github.com/GrottoCenter/grottocenter-front/issues/1148

urien:
Super ton travail. J'ai envoyé un message aux membres de l'association en leur demandant de faire des retours avant mercredi prochain.
De mon côté je vais approfondir le sujet

ClemRz (changes requested):
### 1. EntranceEdit reads stale data on navigation

**File:** `pages/EntityEdit/Entrance/EntranceEdit.jsx` (line 18)

```js
const { loading, data: entrance, error } = useSelector(
  state => state.entrance
);
```

When the user navigates from entrance A's edit page to entrance B's edit page, `state.entrance.data` will still hold entrance A's data while entrance B is loading. The component will briefly render the form with stale data from entrance A before the new fetch completes. Consider resetting the entrance state on mount or checking `entrance.id !== entranceId`.

---

### 2. Missing translation key: `"Loading the entrance data..."`

**File:** `pages/EntityEdit/Entrance/EntranceEdit.jsx`

`formatMessage({ id: 'Loading the entrance data...' })` is used but this key doesn't exist in `en.json`. It will silently fall back to the key itself (since the custom `onError` suppresses `MISSING_TRANSLATION`), but it should be added to the master translation file.

---

### 3. `useProjections` infinite fetch loop if API returns empty array

**File:** `hooks/useProjections.js`

```js
useEffect(() => {
  if (projections.length === 0 && !isLoading) dispatch(fetchProjections());
}, [dispatch, projections.length, isLoading]);
```

If the API returns an empty array, this will re-fetch on every render. Consider tracking a `fetched` flag or checking for an explicit "not yet loaded" state rather than relying on `length === 0`.

---

### 4. `useCallback` inside JSX violates rules of hooks

**File:** `components/appli/EntitiesForm/utils/CoordinateFormSection.jsx` (~line 370)

```jsx
<CRSMenu
  ...
  onSelect={useCallback(
    code => { ... },
    [handleCRSChange, degreesOption, dmsOption, projections]
  )}
/>
```

Calling `useCallback` inline inside JSX is a React rules-of-hooks violation. Hooks must be called at the top level of the component, not inside JSX expressions. This will work in practice because the JSX always renders, but it's fragile and linters should flag it. Extract this to a named callback at the top of the component.

---

### 5. `parseDMS` comment/behavior mismatch for plain decimals

**File:** `helpers/coordinateConvert.js`

The JSDoc comment says it accepts `"-48.5234"` but the function only matches DMS and DDM patterns, then returns `NaN` for anything else. A plain decimal string like `"-48.5234"` won't match either regex and will return `NaN`. If this is intentional (callers handle decimal separately), the comment should be updated. If not, add a `parseFloat` fallback.

---

## Issues

### 6. No `aria-label` on context menu buttons in MapClusters

**File:** `components/common/Maps/MapClusters/index.jsx` (~lines 290-300)

The `CoordinateDisplay` component correctly adds `aria-label` to its icon buttons, but the equivalent copy/CRS buttons in the map context menu only have `Tooltip` without `aria-label`. Add `aria-label` for screen reader accessibility.

---

### 7. `EntranceEdit` error state renders outside Layout

**File:** `pages/EntityEdit/Entrance/EntranceEdit.jsx`

When `error` is truthy, the component returns a bare `<Translate>` without the `Layout` wrapper, which will look inconsistent with the rest of the app. Wrap it in the same `Layout` component.

---

### 8. `CRSMenu` uses `defaultProps` (deprecated in React 19)

**File:** `components/common/CRSMenu/index.jsx`

```js
CRSMenu.defaultProps = {
  anchorEl: null,
  projections: []
};
```

React 19 deprecates `defaultProps` on function components. Use default parameter values in the destructuring instead:

```js
const CRSMenu = ({ anchorEl = null, onClose, preferred, projections = [], onSelect }) => {
```

---

### 9. Map page — `zoom` is state but never updated

**File:** `pages/Map.jsx`

```js
const [zoom] = useState(() => {
  const target = decodeMapTarget(params.target);
  return target ? target.zoom : defaultZoom;
});
```

Since `zoom` is never updated after initialization, it could just be a `useMemo` or a plain `const` with `useRef`. Using `useState` without the setter is a minor code smell — it signals to readers that the value might change.

---

### 10. `AppSnackbar` uses `forwardRef` (unnecessary in React 19)

**File:** `pages/ApplicationShell.jsx`

```js
const AppSnackbar = forwardRef(({ id: _id, message, variant, ...rest }, ref) => {
```

React 19 passes `ref` as a regular prop, so `forwardRef` is no longer needed. Not a bug, just a modernization opportunity.

---

## Suggestions

### 11. `coordinateConvert.js` — document that `proj4` calls can throw

`convertWGS84ToProjection` and `convertProjectionToWGS84` call `proj4()` which can throw on invalid definitions. The callers do wrap in try/catch, which is good. But the functions themselves don't document that they throw. A JSDoc `@throws` annotation would help future callers.

---

### 12. Stale `Converter` translation keys in `en.json`

The old `Converter` component is deleted, but translation keys like `"Converter"`, `"Convert"`, `"Coords converter"`, and `"Based on the library {proj4jsLink}..."` still exist in `en.json`. These are now dead keys. Consider cleaning them up (low priority, won't cause runtime issues).

ClemRz:
I think that these are great UX enhancements, congratulations!

A couple of wishes:
- The CRS dropdown has so many options that we probably need a top free-text field to filter the options based on the input
- Open on map > Grottocenter is a great feature! I was expecting it to open the balloon of the entrance I clicked. How feasible would it be?

Paul-AUB:
All comments addressed/fixed

Paul-AUB:
> * The CRS dropdown has so many options that we probably need a top free-text field to filter the options based on the input
> 
> * Open on map > Grottocenter is a great feature! I was expecting it to open the balloon of the entrance I clicked. How feasible would it be?

NB : for that I did a bit more development, in particular the capability to open popup from map url (e.g. https://orange-rock-0d4f87503-1273.westeurope.azurestaticapps.net/ui/map/45.304414,5.562944,16?popup=entrance:148366).

This may worth an additional review of the code @ClemRz

ClemRz (changes requested):
## Previous review items — status

| # | Item | Status |
|---|------|--------|
| 1 | Stale data on navigation (`EntranceEdit`) | ✅ Fixed — `isStale` guard added |
| 2 | Missing translation key `"Loading the entrance data..."` | ❌ Still missing from `en.json` |
| 3 | `useProjections` infinite fetch loop on empty array | ✅ Fixed — now checks `=== null` instead of `.length === 0` |
| 4 | `useCallback` inside JSX | ✅ Fixed — extracted to `handleCRSSelect` at top level |
| 5 | `parseDMS` comment/behavior mismatch | ✅ Fixed — comment now says "Returns NaN for plain decimals" |
| 6 | No `aria-label` on context menu buttons | ✅ Fixed — both copy and CRS buttons now have `aria-label` |
| 7 | Error state renders outside Layout | ✅ Fixed — error is now inside `Layout`'s `content` prop |
| 8 | `defaultProps` deprecated in React 19 | ✅ Fixed — moved to default parameter values |
| 9 | `zoom` is state but never updated | ✅ Fixed — changed to `useRef` |
| 10 | `forwardRef` unnecessary in React 19 | ✅ Fixed — `ref` is now a regular prop |
| 11 | Document that `proj4` calls can throw | ✅ Fixed — `@throws` annotations added |
| 12 | Stale `Converter` translation keys | ✅ Fixed — `"Convert"`, `"Converter"`, `"Coords converter"` removed from `en.json` |

**One item remains open: #2.** `formatMessage({ id: 'Loading the entrance data...' })` is used in `EntranceEdit.jsx` (line 30) but the key doesn't exist in `en.json`. It works by accident because the custom `onError` handler suppresses `MISSING_TRANSLATION` and react-intl falls back to the key string itself — but it should be added to the master translation file for consistency and so translators can find it.

---

## New issues found in this pass

### N1. `EntranceEdit` has no auth guard

`/ui/entrances/:entranceId/edit` is a public route. An unauthenticated user can navigate to it directly and will see the full edit form. The form submission will fail with a 401, but the UX is confusing — the user fills out fields only to get an error on submit.

The other edit pages (Massif, Organization, Person) have the same pattern, so this is consistent with the existing codebase. But since this is a brand-new route, it's a good opportunity to wrap it in `AuthChecker` or at least add a comment explaining the design choice.

**Severity:** low (consistent with existing patterns, but worth noting)

### N2. Pre-existing reducer typo interacts with `useProjections`

The `Projections` reducer (not modified in this PR) writes `loadings` instead of `loading` in both `FETCH_PROJECTIONS_SUCCESS` and `FETCH_PROJECTIONS_FAILURE`:

```js
case FETCH_PROJECTIONS_SUCCESS:
  return { ...state, loadings: false, projections: action.data };
//                    ^^^^^^^^ should be "loading"
```

This means `state.projections.loading` is set to `true` on `FETCH_PROJECTIONS_LOADING` but never set back to `false`. The new `useProjections` hook reads `state.projections?.loading` — if the fetch succeeds, `loading` stays `true` forever while `projections` becomes non-null. The hook still works because it checks `rawProjections === null` first, and after success `rawProjections` is non-null so the fetch guard short-circuits. But if the fetch *fails*, `loading` stays `true` and `projections` stays `null`, so the hook will never retry.

Not introduced by this PR, but the new hook makes it load-bearing. Worth fixing the reducer typo in a follow-up (or in this PR since it's a one-line change).

**Severity:** medium (silent failure on API error — projections never load and no retry)

### N3. `PopupTargetHandler` — `fetchPopupData` uses raw API URLs without error status checks

```js
const data = await fetch(`${getEntranceUrl}${id}`, { signal }).then(r => r.json());
```

If the API returns a 404 or 500, `.json()` will either throw on non-JSON body or parse an error object that gets treated as valid data. The component will then try to read `data.latitude` / `data.longitude` from the error response and open a popup at `(undefined, undefined)`.

Add a response status check:

```js
const data = await fetch(`${getEntranceUrl}${id}`, { signal })
  .then(r => { if (!r.ok) throw new Error(r.status); return r.json(); });
```

This applies to all four `fetch` calls in `fetchPopupData`.

**Severity:** medium (bad UX on invalid popup target — silent failure or broken popup)

### N4. `popupTarget` parsing accepts `id: 0` as invalid

```js
const id = parseInt(idStr, 10);
if (!type || !id) return null;  // !0 is true → id 0 is rejected
```

`!id` is falsy for `0` and `NaN`. If entity IDs can never be 0 this is fine, but `Number.isNaN(id)` would be a more precise guard. Minor, since IDs are likely always positive integers from the API.

**Severity:** negligible

### N5. `decodeMapTarget` is called twice on mount

```js
const [location, setLocation] = useState(() => {
  const target = decodeMapTarget(params.target);
  return target ? { lat: target.lat, lng: target.lng } : defaultCoord;
});
const { current: zoom } = useRef(
  decodeMapTarget(params.target)?.zoom ?? defaultZoom
);
```

`decodeMapTarget` is pure and cheap, so this isn't a bug — just a minor redundancy. Could destructure once:

```js
const initialTarget = useMemo(() => decodeMapTarget(params.target), []);
const [location, setLocation] = useState(() =>
  initialTarget ? { lat: initialTarget.lat, lng: initialTarget.lng } : defaultCoord
);
const { current: zoom } = useRef(initialTarget?.zoom ?? defaultZoom);
```

**Severity:** negligible (style)

### N6. `AppSnackbar` missing `PropTypes` and `displayName`

The `AppSnackbar` component is defined as an arrow function assigned to a `const`, but has no `propTypes` or `displayName`. Since it's passed to notistack's `Components` map, React DevTools will show it as `Anonymous`. Adding a `displayName` or converting to a named function expression would help debugging.

**Severity:** negligible (DX)

### N7. `CoordinateFormSection` — `useEffect` deps lint suppression

```js
}, [watchedLat, watchedLng]);
```

The suppressed deps are `selectedCRS` and `prefillFromWGS84`. The comment above explains the intent (only re-sync on *external* WGS84 changes, not on CRS changes), which is correct. But if `prefillFromWGS84` ever gains dependencies beyond `[]`, this will silently break. A brief inline comment explaining *which* deps are intentionally omitted and why would help future maintainers.

**Severity:** negligible (already has a good comment above, just could be more explicit)

Paul-AUB:
# Review fixes — feature/coordinates

## Changes applied

### Missing i18n key

`packages/web-app/public/lang/en.json`
Added `"Loading the entrance data..."` key. Was working by accident (react-intl falls back to the key string when `MISSING_TRANSLATION` is suppressed), but translators couldn't find it.

### N1 — No auth guard on edit routes

`packages/web-app/src/components/appli/PrivateRoute/index.jsx` (new)
`packages/web-app/src/App.jsx`
Created a `PrivateRoute` layout component (wraps `AuthChecker` with `<Outlet />`). All five edit routes are now grouped under it in `App.jsx`: entrances, organizations, massifs, persons, documents. Unauthenticated users see the standard "You must be authenticated" error with a login button instead of the form.

### N2 — Projections reducer typo

`packages/web-app/src/reducers/Projections.js`
Fixed `loadings: false` → `loading: false` in both `FETCH_PROJECTIONS_SUCCESS` and `FETCH_PROJECTIONS_FAILURE` cases. The typo meant `state.projections.loading` was set to `true` on fetch start but never reset to `false`, causing the hook to stay in a permanent loading state on API failure.

### N3 — fetchPopupData missing response status checks

`packages/web-app/src/components/common/Maps/MapClusters/PopupTargetHandler.jsx`
Added `if (!r.ok) throw new Error(r.status)` before `r.json()` in all four fetch calls (entrance, network, massif, organization). Previously a 404/500 would parse the error body as valid data and open a broken popup at `(undefined, undefined)`.

### N4 — popupTarget id guard rejects 0

`packages/web-app/src/pages/Map.jsx`
Changed `!id` → `Number.isNaN(id)`. The `!id` check was falsy for `0` and `NaN`; `Number.isNaN` is precise and only rejects `NaN`.

### N5 — decodeMapTarget called twice on mount

`packages/web-app/src/pages/Map.jsx`
Computed `decodeMapTarget(params.target)` once via `useMemo` and shared the result between `useState` and `useRef`, removing the redundant second call.

### N6 — AppSnackbar missing propTypes and displayName

`packages/web-app/src/pages/ApplicationShell.jsx`
Added `AppSnackbar.displayName = 'AppSnackbar'` and `AppSnackbar.propTypes`. The component was anonymous in React DevTools because it was an arrow function assigned to a `const`.

### N7 — Vague eslint-disable comment in CoordinateFormSection

`packages/web-app/src/components/appli/EntitiesForm/utils/CoordinateFormSection.jsx`
Expanded the comment above `// eslint-disable-next-line react-hooks/exhaustive-deps` to explicitly name which deps are omitted (`selectedCRS`, `prefillFromWGS84`) and why (re-sync should only trigger on external WGS84 changes, not CRS switches).

ClemRz (approved):
This is awesome! I think that the map link with the popup argument will be helpful for other purpose, I remember someone asking for it at some point, do you remember @urien ?

urien:
Salut @ClemRz il y avait une demande pour pouvoir récupérer les coordonnées d'un point quelconque sur la carte. La fonctionnalité est bien plus riche

Issue #1206, opened 2026-03-10
Massif sensitive setting user interface


ClemRz:
# Requirements Document

## Introduction

This feature adds frontend support for massif-level sensitivity in GrottoCenter. The backend API (grottocenter-api) provides an `isSensitive` boolean on the TMassif model, cascading sensitivity to entrances within a massif polygon when enabled. The API enforces an Administrator-only permission model: only Administrators can enable or disable massif sensitivity, and the `isSensitive` field is omitted from API responses for non-Administrator users. The frontend must surface this information to Administrators, provide appropriate controls, handle the field's absence for other users, and give clear feedback about cascading effects and legal basis.

## Glossary

- **Massif_Detail_Page**: The page component at `src/components/appli/Massif/Massif.jsx` that displays a single massif's information, accessed via `/ui/massifs/:massifId`.
- **Massif_Form**: The form component at `src/components/appli/EntitiesForm/Massif/index.jsx` used for creating and editing massifs.
- **Sensitivity_Toggle**: A Switch control in the Massif_Form that allows toggling the `isSensitive` field on a massif.
- **Confirmation_Dialog**: A modal dialog presented to the Administrator before enabling sensitivity on a massif, warning about cascading effects on entrances.
- **Removal_Confirmation_Dialog**: A modal dialog presented to the Administrator before disabling sensitivity on a massif, warning that entrance-level sensitivity flags will not be removed automatically.
- **Sensitivity_Indicator**: A visual element (MUI Alert with warning severity) displayed on the Massif_Detail_Page to communicate that a massif is sensitive. Only visible to Administrators.
- **Permissions_Hook**: The `usePermissions` hook at `src/hooks/usePermissions.js` that exposes `isAdmin`, `isModerator`, `isLeader`, and `isAuth` booleans derived from the JWT token.
- **Massif_Reducer**: The Redux reducer at `src/reducers/MassifReducer.js` that manages the massif state including the fetched massif object.
- **Update_Massif_Action**: The Redux thunk at `src/actions/Massif/UpdateMassif.js` that sends a PUT request to update a massif.
- **Affected_Entrances_Count**: The integer returned by the API in the update response indicating how many entrances were marked sensitive when massif sensitivity was enabled.
- **Affected_Entrances_Preview**: The integer returned by the API before the update is confirmed, indicating how many entrances within the massif polygon will be affected if sensitivity is enabled.
[...]

PR #1277, opened 2026-04-29
Feat/1206 massif sensitive user interface

dawoldo:
# feat(1206): Implement massif-level sensitivity management UI

## 🤔 What

This PR implements the user interface and state management for managing massif-level sensitivity. It allows authorized users to mark or unmark an entire massif as sensitive, which in turn affects all associated entrances.

Key changes include:
- **New Component**: `MassifSensitivityControl` provides a toggle switch and confirmation dialog for managing sensitivity.
- **API Integration**: Added support for `preview-sensitive`, `mark-sensitive`, and `unmark-sensitive` endpoints.
- **Redux State**: Implemented actions and reducers for handling sensitivity state and asynchronous API calls.
- **Localization**: Added full translation support (EN, FR, ES) for sensitivity-related strings, including confirmation messages and status notifications.
- **Type Definitions**: Updated `Massif` type to include the `isSensitive` property.
- **Error Handling**: Implemented defensive handling for forbidden (403) errors on the sensitivity switch.

[Links issue: #1206](https://github.com/GrottoCenter/grottocenter-front/issues/1206#issue-4053717516)

## 🤷‍♂️ Why
This feature allows administrators to protect an entire geographic area in one action. It also provides a clear preview of the impact before applying changes.

## 🔍 How
- **State Management**: Created dedicated Redux actions (`MarkSensitiveMassif`, `PreviewSensitiveMassif`, `UnmarkSensitiveMassif`) and reducers to manage the lifecycle of sensitivity updates independently of the main massif update flow.

- **UI/UX**: 
    - The `MassifSensitivityControl` is integrated into the Massif edit form.
    - When toggling sensitivity, a confirmation dialog is triggered.
    - This dialog calls the `preview` endpoint to inform the user exactly how many entrances will be affected.
    
- **Architecture**: Decoupled the sensitivity toggle from the standard form submission to ensure that these sensitive operations are handled with explicit confirmation and immediate feedback.

- **Defensive Design**: Added checks for permissions and handled cases where the UI might be accessible but the user lacks the specific rights to perform the action.

## 🔗 Related API PR

This frontend implementation requires the corresponding changes in the [Grottocenter API repository](https://github.com/GrottoCenter/grottocenter-api/issues/1450#issue-3928987374).

## 🧪 Testing
1. **Authentication**: Log in with an administrator account.
2. **Navigation**: Go to any Massif page and enter the "Edit" mode.
3. **Toggle Sensitivity**:
    - Switch the "Sensitive" toggle.
    - Verify that the confirmation dialog appears and correctly displays the number of affected entrances (preview).
    - Confirm the change and verify the success notification.
4. **Localization**: Change the language to French or Spanish and verify that all sensitivity-related labels and messages are correctly translated.
5. **Permissions**: Test with a user who does not have administrative rights to verify that they cannot perform these actions (and handle 403 errors if the toggle is interacted with).


urien:
Le serveur de test ne semble pas se connecter aux données alors que https://grottocenter.org fonctionne normalement

ClemRz (changes requested):
### 1. `.env` changed to localhost — must not be merged

The PR changes `REACT_APP_API_URL` from `https://api.grottocenter.org` to `http://localhost:1337`. This is a local development artifact that would break the production build and every other developer's setup. This must be reverted before merge. For local development please modify `.env.local` instead.

**File:** `packages/web-app/.env`

### 2. `UnmarkSensitiveMassif` action: inconsistent response handling

In `MarkSensitiveMassif.js`, the success handler extracts `data.massif`:

```js
.then(data => dispatch(markMassifSensitiveSuccess(data.massif, data.count)))
```

But in `UnmarkSensitiveMassif.js`, it passes the entire `data` object:

```js
.then(data => dispatch(unmarkMassifSensitiveSuccess(data)))
```

The `MassifReducer` expects `action.massif` for `UNMARK_MASSIF_SENSITIVE_SUCCESS`:

```js
case UNMARK_MASSIF_SENSITIVE_SUCCESS:
  return { ...state, massif: action.massif, isFetching: false };
```

But the action creator sets `massif: data` (the full response). If the API returns `{ massif: {...} }`, then `state.massif` will be `{ massif: {...} }` instead of the massif object itself. This will likely break the detail page rendering. The unmark action should extract `data.massif` like the mark action does.

**File:** `packages/web-app/src/actions/Massif/UnmarkSensitiveMassif.js`

### 3. `handleConfirm` catches errors but actions swallow them

In `MarkSensitiveMassif.js` and `UnmarkSensitiveMassif.js`, the `.catch()` block dispatches the failure action and then implicitly returns `undefined` (the promise resolves, it doesn't reject). So in `handleConfirm`, the `catch` block will never fire for API errors — the `await dispatch(...)` will resolve to `undefined` rather than throwing.

This means:

- The 403 error handling in the component (Requirement 6) will never trigger
- The optimistic state revert on error will never execute
- The error notification will never show

The `PreviewSensitiveMassif` action correctly re-throws (`throw error`), but the mark/unmark actions don't. They need to re-throw after dispatching the failure action, similar to how the preview action does it.

**Files:** `MarkSensitiveMassif.js`, `UnmarkSensitiveMassif.js`

### 4. Sensitivity toggle fires immediately on switch change, not on form submit

The spec says (Req 3, criterion 1): "WHEN an Administrator toggles the Sensitivity_Toggle from `false` to `true` **and submits the Massif_Form**". The PR fires the preview/confirm flow immediately when the switch is toggled (`onChange={handleToggle}`), bypassing the form submission entirely. The component is also placed outside the `<form>` element.

This is a UX design choice that may be intentional (decouple sensitivity from the normal form flow), and the PR description explicitly calls this out. But it deviates from the spec. If this is the desired behavior, the spec should be updated.

### 5. `UPDATE_MASSIF_SUCCESS` added to `MassifReducer` — verify no duplicate on `develop`

The PR adds `UPDATE_MASSIF_SUCCESS` to the `MassifReducer`'s case list as a new import. Verify this doesn't cause a duplicate case or conflict with what's already present on `develop`.

**File:** `packages/web-app/src/reducers/MassifReducer.js`

### 6. Three separate reducers for what could be one

`MarkSensitiveMassifReducer`, `UnmarkSensitiveMassifReducer`, and `PreviewSensitiveMassifReducer` are nearly identical boilerplate. These three could be consolidated into a single `MassifSensitivityReducer` with a combined state shape. This would reduce the 3 new entries in `GCReducer` to 1 and cut ~60 lines of duplication.

### 7. Missing `Content-Type` header on POST requests

The `MarkSensitiveMassif` and `UnmarkSensitiveMassif` actions use `method: 'POST'` but don't set `Content-Type: 'application/json'` and don't send a body. Compare with `UpdateMassif.js` which sets `body: JSON.stringify(body)` and relies on the auth header. If the API expects JSON content type, these requests may fail. If no body is needed, `POST` without a body is fine, but confirm the API doesn't require it.

### 8. Extra blank lines in `MassifSensitivityControl.jsx`

Two consecutive blank lines after the `useEffect` block (around line 37–38 in the new file). Prettier should catch this.

### 9. Hardcoded border color

```jsx
<Box sx={{ mt: 3, mb: 3, p: 2, border: '1px solid #ccc', borderRadius: 1 }}>
```

The `#ccc` hardcoded color doesn't use the MUI theme. It should use `theme.palette.divider` or similar to respect dark mode and theme consistency:

```jsx
<Box sx={{ mt: 3, mb: 3, p: 2, border: 1, borderColor: 'divider', borderRadius: 1 }}>
```

### 10. Translation files not sorted

New keys in `es.json` and `fr.json` are appended at the end and not sorted alphabetically. Run `yarn translations:sort` on these files.

### 11. Spanish translations use informal register

The existing `es.json` translations in the project use the formal "usted" form. Several new strings in this PR use the informal "tú" form instead (e.g., "¿Desea continuar?" is correct, but other existing keys in the file use "usted" conjugations). Review all new Spanish strings for consistent use of the formal register to match the rest of the application.

dawoldo:
> 7. Missing Content-Type header on POST requests

No Request Body Required: The endpoints ```POST /massifs/{id}/mark-sensitive``` and ```POST /massifs/{id}/unmark-sensitive``` only define a path parameter (id). There is no ```requestBody``` defined in the specification for these actions. Their current implementation (which sends a POST request without a body or ```Content-Type header```) is consistent with other similar actions in the codebase, such as ```SubscribeToMassif.js```

ClemRz (changes requested):
## 1. Missing PropTypes on `MassifSensitivityControl`

**Severity:** Low
**File:** `packages/web-app/src/components/appli/EntitiesForm/Massif/MassifSensitivityControl.jsx`

The component accepts a `massif` prop but doesn't define `propTypes`. Every other component in the codebase defines them. Add:

```jsx
import { MassifTypes } from '../../../../types/massif.type';

MassifSensitivityControl.propTypes = {
  massif: MassifTypes
};
```

---

## 2. Sensitivity alert only visible to admins on the detail page

**Severity:** Medium (design question)
**File:** `packages/web-app/src/components/appli/Massif/Massif.jsx`

The sensitivity warning on the massif detail page is gated behind `permissions.isAdmin`:

```jsx
{permissions.isAdmin && massif.isSensitive && (
  <Alert severity="warning" title={...} />
)}
```

The equivalent entrance-level component (`SensitiveCaveWarning`) is shown to **all users** without any permission check. If the goal is to inform visitors that the massif is sensitive, the admin guard should be removed. If it's intentionally admin-only, that's fine — but it should be a deliberate choice, not an oversight.

---

## 3. `UPDATE_MASSIF_SUCCESS` in `MassifReducer` is unrelated to sensitivity

**Severity:** Low
**File:** `packages/web-app/src/reducers/MassifReducer.js`

This PR adds `UPDATE_MASSIF_SUCCESS` as a new case in `MassifReducer` so the detail page refreshes after an edit. That's a valid cross-reducer listener, but it has nothing to do with the sensitivity feature. It should either live in a separate commit/PR or be explicitly called out in the PR description so reviewers understand it's a bonus fix, not part of the sensitivity work.

---

## 4. `massifSensitivity` Redux state is unused by the component

**Severity:** Low (dead state)
**Files:** `packages/web-app/src/reducers/MassifSensitivityReducer.js`, `packages/web-app/src/reducers/GCReducer.js`

`MassifSensitivityControl` never reads from `state.massifSensitivity`. All loading, error, and count state is managed locally via `useState`. The `MassifReducer` already handles the success cases by updating `state.massif` (which is what the detail page consumes).

Unless another component needs to observe sensitivity operation status, `MassifSensitivityReducer` and its entry in `GCReducer` are dead code. Consider removing them to keep the store lean.

---

## 5. Optimistic switch flip before confirmation causes visual flicker

**Severity:** Minor (UX)
**File:** `packages/web-app/src/components/appli/EntitiesForm/Massif/MassifSensitivityControl.jsx`

`handleToggle` calls `setOptimisticSensitive(!isSensitive)` immediately, which flips the switch visually *before* the confirmation dialog opens. If the user cancels, `handleCancel` reverts it. This creates a brief flicker: switch toggles → dialog appears → user cancels → switch toggles back.

Consider deferring the optimistic update to `handleConfirm` (after the user actually confirms), or at least waiting until the dialog is open.

---

## 6. `isAuthError` handling inconsistent with codebase convention

**Severity:** Minor
**Files:** `packages/web-app/src/actions/Massif/MarkSensitiveMassif.js`, `packages/web-app/src/actions/Massif/UnmarkSensitiveMassif.js`

The catch blocks use:

```js
if (error.isAuthError) return Promise.reject(error);
```

The established convention in other authenticated actions (`UpdateMassif`, `GetMassif`, `DeleteMassif`, `RestoreMassif`) is:

```js
```

`checkAuthStatus` already handles 401 by logging out and reloading the page, so the rejected promise reaching the component is harmless — but it's inconsistent. Aligning with `return;` keeps the pattern uniform.

ClemRz (changes requested):
### 1. `FormActionRow` receives unknown props — Cancel button silently lost

**Severity:** Bug
**File:** `packages/web-app/src/components/appli/EntitiesForm/Massif/index.jsx`

The PR changed:

```jsx
<FormActionRow
  isDirty={isDirty}
  isNew={isNewMassif}
  isSubmitting={isSubmitting}
  onReset={handleReset}
/>
```

But `FormActionRow` only accepts `{ isNew, isSubmitting, isCenter, onCancel }`. The `isDirty` and `onReset` props are silently ignored. This means the Massif edit form no longer renders a Cancel button — the previous code passed `onCancel={onCancel}` (from the component's props), and now nothing maps to `onCancel`. This is a regression in the edit form UX unrelated to the sensitivity feature.

Either revert to passing `onCancel` as before, or if the intent was to change the cancel behavior, `FormActionRow` needs to be updated to accept the new props.

---

### 2. Optimistic switch flip still happens before confirmation

**Severity:** Minor (UX)
**File:** `MassifSensitivityControl.jsx`, line 72

`handleConfirm` calls `setOptimisticSensitive(!isSensitive)` at the top, before the async action completes. If the action fails, it reverts in the catch block. This creates a brief visual flicker: switch flips → API call → error → switch flips back.

This was raised in the second review (item 5). Moving `setOptimisticSensitive(!isSensitive)` to after the `await dispatch(...)` succeeds would eliminate the flicker without changing the happy-path behavior.

---

### 3. Dispatched failure actions are dead code

**Severity:** Low (code hygiene)
**Files:** `MarkSensitiveMassif.js`, `UnmarkSensitiveMassif.js`, `PreviewSensitiveMassif.js`

`MARK_MASSIF_SENSITIVE_FAILURE`, `UNMARK_MASSIF_SENSITIVE_FAILURE`, and `PREVIEW_MASSIF_SENSITIVE_FAILURE` are dispatched but no reducer listens for them. The component handles errors locally via try/catch. These dispatches are harmless but add noise. Either remove them or add a comment explaining they're kept for Redux DevTools observability.

urien:
Merci Zouhair,

Pour moi le fonctionnement est conforme a ce qui est attendu.
il y a 2 points a reprendre
-  est ce que tu pourrais mettre en place la traduction dans les 15 langues désormais ? Paul vient de le faire pour tous les textes existants et ce serait bien de conserver cette exigence désormais
- lorsqu'on marque un massif comme sensible c'est visible en tant qu'admin mais si je ne suis pas connecté le texte n'est pas visible. Je pense que c'est indispensable car dans une issue prochaine tu vas rajouter la disposition applicable qui sera également visible par tous, connecté ou pas

image
 
image

ClemRz (changes requested):
### 1. `PREVIEW_MASSIF_SENSITIVE_SUCCESS` dispatch is dead code

**Severity:** Low (code hygiene)
**File:** `packages/web-app/src/actions/Massif/PreviewSensitiveMassif.js`

The mark/unmark failure dispatches were removed, but the preview action still dispatches `previewMassifSensitiveSuccess(data.count)` — and no reducer listens for `PREVIEW_MASSIF_SENSITIVE_SUCCESS`. The component uses the returned `data.count` value directly. The success action creator and its dispatch serve no purpose.

```js
// Lines 7-10: dead code
const previewMassifSensitiveSuccess = count => ({
  count,
  type: PREVIEW_MASSIF_SENSITIVE_SUCCESS
});
```

```js
// Line 28: dispatch has no effect
dispatch(previewMassifSensitiveSuccess(data.count));
return data.count; // ← this is what the component actually uses
```

Consider removing the success action creator and dispatch, or keep it explicitly for Redux DevTools traceability with a comment.

---

### 2. Translations only cover 3 of 15 languages

**Severity:** Medium (raised by urien in comments)
**Files:** Only `en.json`, `es.json`, `fr.json` have the new sensitivity keys.

urien commented (2026-05-05) that all 15 languages should now have translations, since Paul recently completed full coverage for existing strings. The remaining 12 language files need the new keys added (at minimum with English fallback values).

---

### 3. Sensitivity alert potentially not visible to unauthenticated users

**Severity:** Medium (raised by urien in comments)
**File:** `packages/web-app/src/components/appli/Massif/Massif.jsx`

urien reports that the sensitivity warning is not visible when not logged in. The frontend code has no permission guard on the alert — it renders whenever `massif.isSensitive` is truthy. This suggests the API may not return the `isSensitive` field for unauthenticated requests, or the field is stripped from the response. This needs investigation on the API side. If the API intentionally omits it, the frontend can't fix this alone.

ClemRz (changes requested):
Good progress — most issues from previous reviews have been addressed!

### Missing translation key in `fr.json` and `es.json`

**Severity:** Bug
**Files:** `packages/web-app/public/lang/fr.json`, `packages/web-app/public/lang/es.json`

The key `"This massif is marked as sensitive based on applicable legislation."` (used in the `Massif.jsx` detail page alert) is present in all 13 other language files but **missing** from `fr.json` and `es.json`. This means French and Spanish users will see the raw English key as the alert text on the massif detail page.

This is likely the issue urien flagged on 2026-05-06 ("La traduction n'est pas bonne pour le message d'alerte").

ClemRz (approved):

PR #1302, opened 2026-05-02
Fix massif creation and edition form

Paul-AUB:
# Massif creation form: remove description fields

Linked to https://github.com/GrottoCenter/grottocenter-api/pull/1555

## What changed

The API no longer requires a description or title when creating a massif. The creation form has been updated accordingly:

- Removed `descriptionTitle` and `descriptionBody` fields from `MassifFields.jsx`
- Removed `descriptionTitle`, `descriptionBody` from `defaultMassifValues` in `index.jsx`
- The `postMassif` payload now only sends `name`, `descriptionAndNameLanguage`, and `geogPolygon`

## Why it makes sense

Descriptions are already handled as a separate concern in the app. On any massif page, the `Descriptions` ScrollableContent component lets authenticated users add, edit, and reorder descriptions independently — after the massif is created.

Duplicating that capability in the creation form would mean maintaining two separate code paths for the same data, with different validation logic, different UX patterns, and different API calls. It also creates an asymmetry: description editing is never part of the massif edit form, only creation — which is confusing.

The right flow is:
1. Create the massif (name + language + area)
2. Add descriptions from the massif page, using the existing `Descriptions` component

This keeps the creation form minimal and focused, and reuses the existing description infrastructure for everything description-related.

---

## Responsive layout

Made the massif creation/edit form responsive, with a focus on the map area (`PolygonMap.jsx`, `PolygonLayersList.jsx`).

### DIFF
--- a/packages/web-app/src/components/appli/EntitiesForm/Massif/MassifFields.jsx
+++ b/packages/web-app/src/components/appli/EntitiesForm/Massif/MassifFields.jsx
@@ -10,8 +10,9 @@ import { FormRow, FormSectionLabel } from '../utils/FormContainers';
 
 const PolygonMap = React.lazy(() => import('./PolygonMap'));
 
-const MassifFields = ({ control, errors, geoJson, isNew }) => {
+const MassifFields = ({ control, errors, geoJson }) => {
   const { formatMessage } = useIntl();
+
   return (
     <>
       <FormSectionLabel label={formatMessage({ id: 'Basic Information' })} />
@@ -31,29 +32,6 @@ const MassifFields = ({ control, errors, geoJson, isNew }) => {
         />
       </FormRow>
 
-      {isNew && (
-        <>
-          <FormSectionLabel
-            label={formatMessage({ id: 'Description of the massif' })}
-          />
-          <InputText
-            formKey="massif.descriptionTitle"
-            labelName="Title"
-            control={control}
-            isError={!!errors?.massif?.descriptionTitle}
-            isRequired
-          />
-          <InputText
-            formKey="massif.descriptionBody"
-            labelName="Description"
-            control={control}
-            isError={!!errors?.massif?.description}
-            isRequired
-            minRows={6}
-          />
-        </>
-      )}
-
       <FormSectionLabel label={formatMessage({ id: 'Massif area' })} />
       <FormHelperText>
         {formatMessage({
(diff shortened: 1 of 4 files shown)


ClemRz (changes requested):
Clean, well-scoped PR. The two concerns (removing description fields + responsive layout) are logically related and the rationale in the PR body is solid.

## Commit messages (blocking)

Both commits violate the project's commitlint rules. The config requires a scope in pascal-case or camel-case (`scope-empty: [2, 'never']`), but neither commit has one:

- `feat: don't ask for description at massif creation` → e.g. `feat(Massif): remove description fields from creation form`
- `fix: responsive massif edit map page` → e.g. `fix(Massif): make polygon map responsive`

These will fail the pre-commit hook. They need to be amended before merge.

## Code

**Description removal (MassifFields + index.jsx)** — Looks correct. The fields are cleanly removed from default values, form JSX, `postMassif` payload, and PropTypes. The `isNew` prop removal from `MassifFields` is appropriate since it was only used to conditionally show the description block. Other entities (Entrance, Organization) still use their own `descriptionTitle`/`descriptionBody` — no collateral damage.

**Responsive layout** — The `invalidateSize()` call, the flex direction switch, and the `PolygonLayersList` responsive sizing all look good. The mixed approach (JS `isMobile` for `MapContainer`'s `style` prop vs MUI responsive `sx` syntax elsewhere) is justified since `MapContainer` doesn't support `sx`.

**Minor note on `invalidateSize` timing**: the CSS transition from the responsive breakpoint may not have completed by the time the effect fires. If the map occasionally renders with stale tile gaps after a resize, wrapping the call in a short `setTimeout` (~200ms) would fix it. Not a blocker.

No translation key cleanup or reducer changes needed — the removed fields used generic keys still referenced elsewhere, and `postMassif` passes the body straight through to the API.

urien:
C'est une très bonne solution @Paul-AUB . 
Oeux tu aussi t'assurer que le fonctionnement est identique pour l'édition d'un massif
Je préviens Zouhair qui travaille sur la gestion des massifs

urien (approved):
En tant qu'utilisateur le fonctionnement est parfait : fluide, simple. Merci @Paul-AUB

Paul-AUB:
> Peux tu aussi t'assurer que le fonctionnement est identique pour l'édition d'un massif

L'édition du massif n'avait déjà pas de champ Description (c'est pour ça que j'ai enlevé ce champ à la création pour rester homogène). J'ai testé sur https://grottocenter.org/ui/massifs/3662, pas de régression à l'édition

Paul-AUB:
- Commit messages fixed
- Added setTimeout for `invalidateSize`

ClemRz (approved):

PR #1306, opened 2026-05-06
Documents creation UI/UX

### LINKED ISSUE #42: [ENTRANCE PAGE] Add a document
Depuis une cavité existante créer la possibilité de lancer le processus de création d'un document tel qu'il a été mis au point . A la fin du processus le document est créé et il est associé à la cavité

Comme pour tous les documents celui-ci devra être validé pour apparaitre associé à la cavité. Il faudrait donc afficher, à la fin du processus, la page de la cavité avec un message indiquant que le document n'apparaitra que quand il sera validé

### LINKED ISSUE #806: [DOCUMENT CREATE] Language autocompletion
Quand on créé un document il faut saisir la langue du document et la langue du résumé, c'est un peu fastidieux car en règle générale c'est la langue qu'on utilise pour Grottocenter

Auto-compléter ces langues avec la langue utilisée pour Grottocenter

Paul-AUB:
# feat(documents): improve documents creation UI/UX

## 🤔 What

closes https://github.com/GrottoCenter/grottocenter-front/issues/42

Complete rework of the document creation form:

- **Type selector**: card-based UI with Image, Topographic Drawing and Video as featured types; other types in a collapsible grid
- **Adaptive layouts**: simplified layout (Title / Language / Description / Files) for Image, Video, Sound, Map, Dataset, Interactive Resource, Physical Object, Service, Software, Topographic Data; full layout for Article, Book, Issue, Collection, Text, etc.
- **Advanced metadata accordion**: Publication date, Identifier, Editor, Library, ISO regions, BBS subjects and Comment collapsed by default — available for all types except Event and Authorization to Publish
- **Parent document & Pages** moved into the main Identification area (right after the description) for Article and Issue
- **Authorization / licensing UX** consolidated inside `AddFileForm`:
  - Radio buttons for licensing type (hidden until a file is uploaded)
  - Auto-fills the logged-in user as author when "I am the author" is selected
  - License selector shown only after a radio is selected
  - Authorization document selector required when "There is an authorization" is chosen
- **File type restrictions**: per-type accept config for Image, Topographic Drawing, Video (`video/*,image/gif`), Sound (`audio/*`), Topographic Data, Authorization to Publish
- **Required files**: validation enforces at least one file for Image, Topographic Drawing, Video, Sound, Map, Dataset, Physical Object
- **Linked entrance**: the form can be pre-linked to a cave entrance (via URL parameter), showing an info banner and automatically associating the created document
- **Form reset on type change**: switching type resets all fields except Title and Language
- **Translation**: "Moving Image" → language-appropriate "video" word in all 15 languages; new "Advanced metadata" key added

## 🤷‍♂️ Why

The previous form was a single long page with all fields always visible regardless of document type, making it overwhelming for the most common use cases (uploading an image or a video). The authorization/licensing section was split across multiple components with inconsistent state management.

## 🔍 How

- `DocumentTypeSelect`: card grid with `FEATURED_TYPES` constant; `resetContext()` on type change preserving title and language
- `AddFileForm`: new `showAuthorization` prop drives both UI and side-effects (license fetch, author auto-set); `validateAndBuildFileEntries` extracted to `FileHelpers.js`
- `FormContent`: three layout branches (simple / event / auth-to-publish / full) + a single `Accordion` for advanced metadata rendered outside all branches
- `Provider.js`: `FILE_REQUIRED_TYPES` and `SIMPLE_MEDIA_TYPES` sets extended; `checkFormValidation` covers authorization document requirement
- `fetchLicense` guarded against redundant calls (`!licenses && !licensesLoading`)
- Translation scripts: BOM stripping added to `loadJson`; `sync-with-en.js` no longer re-translates keys already matching English

## 🔗 Related API PR

_None._

## 🧪 Testing

1. Select each document type and verify the correct layout and file restrictions apply
2. Upload a file for Image/Video/Sound — verify the authorization radios appear and the submit button stays disabled until one is selected
3. Select "I am the author" — verify your username is auto-filled and the license selector is hidden
4. Select "There is an authorization" — verify an authorization document must be selected before submitting
5. Switch document type mid-form — verify all fields except Title and Language are reset
6. Verify "Advanced metadata" accordion is hidden for Event and Authorization to Publish types

## 📸 Previews

_To be added._


ClemRz (changes requested):
### Critical

**1. `AuthDocSelect` reads from wrong state path**

```jsx
const { data, isLoading } = useSelector(state => state.documents);
const authDocs = data.authorizationDocuments ?? [];
```

Looking at the reducer, `state.documents.data` has shape `{ documents: [], authorizationDocuments: [] }`. The `getDocuments` action dispatched here uses `documentType: 'Authorization To Publish'`, which triggers `FETCH_AUTHORIZATION_DOCUMENTS_SUCCESS`. However, it also dispatches `FETCH_DOCUMENTS` first, which sets `isLoading: true` and clears `error`. The issue is that `isLoading` from `state.documents` reflects *any* documents fetch, not just the authorization one. If another component triggers a regular `getDocuments` call concurrently, the loading state will be shared/conflicting.

More critically: the `useEffect` dispatches `getDocuments(...)` on every mount with no guard. If the authorization documents are already loaded, it will re-fetch every time the component mounts (e.g., when toggling the radio). Consider adding a guard similar to the license fetch pattern (`!authDocs.length && !isLoading`).

**2. `resetContext` + immediate `updateAttribute` calls may race**

In `DocumentTypeSelect.handleSelect`:
```jsx
const handleSelect = newDocType => {
  const { title, mainLanguage, mainLanguageName } = document;
  resetContext();
  updateAttribute('type', newDocType);
  updateAttribute('title', title);
  updateAttribute('mainLanguage', mainLanguage);
  updateAttribute('mainLanguageName', mainLanguageName);
};
```

`resetContext` calls `setDocument(defaultDocAttributes)` and each `updateAttribute` calls `setDocument(prevState => ({...prevState, [key]: value}))`. Since React batches state updates within the same event handler, the `prevState` in each `updateAttribute` call will be the state *before* `resetContext` ran (React processes them in order, but each sees the previous queued update). This should actually work correctly with React 18+ automatic batching — but it's fragile. A single `setDocument` call with the merged object would be safer and more explicit.

**3. Missing `createdDocument` in reducer reset**

In `CreateDocumentReducer.js`, the `POST_DOCUMENT` (loading) case doesn't reset `createdDocument`:
```jsx
case POST_DOCUMENT:
  return { ...state, isLoading: true, errorMessages: [] };
```

If a user submits, gets a `createdDocument`, then submits again, the stale `createdDocument` from the first submission persists during loading. The `useEffect` in `DocumentSubmission` that links the document checks `documentState.createdDocument` — it could fire with the old document if the second submission fails. Add `createdDocument: undefined` to the `POST_DOCUMENT` case.

---

### Major

**4. `useEffect` dependency lint suppression in `AddFileForm` hides a real bug**

```jsx
useEffect(() => {
  if (!showAuthorization || !licenses) return;
  const licenseName = documentLicenseName ?? DEFAULT_LICENSE;
  const selected = licenses.find(l => l.name === licenseName);
  if (selected) setLicense(selected);
}, [showAuthorization, licenses, documentLicenseName]);
```

`setLicense` is `newLicense => updateAttribute('license', newLicense)` — an inline arrow created on every render of `FormContent`. This means the effect captures a stale closure. In practice it works because `updateAttribute` uses `setDocument(prev => ...)`, but if `setLicense` ever does more (e.g., side effects), this will silently break. Consider stabilizing `setLicense` with `useCallback` in the parent or passing `updateAttribute` directly.

**5. Author auto-fill effect runs on every `option` change without cleanup**

```jsx
useEffect(() => {
  if (!showAuthorization) return;
  if (option === AUTHORIZATION_FROM_AUTHOR && authTokenDecoded) {
    updateAttribute('authors', [{ id: authTokenDecoded.id, nickname: authTokenDecoded.nickname }]);
  } else if (option && option !== AUTHORIZATION_FROM_AUTHOR) {
    updateAttribute('authors', []);
  }
}, [option, showAuthorization, authTokenDecoded, updateAttribute]);
```

If the user manually adds authors, then switches to a different radio option, this effect clears them. That's probably intentional, but switching *back* to `AUTHORIZATION_FROM_AUTHOR` will overwrite any manually-added co-authors with just the logged-in user. Consider whether this is the desired behavior or if it should only auto-fill when authors is empty.

**6. `FormActionRow` `disabled` prop not wired to `ActionButton` correctly**

```jsx
<ActionButton
  label={...}
  loading={isSubmitting}
  disabled={disabled}
  ...
  type="submit"
/>
```

If `ActionButton` doesn't forward the `disabled` prop (it's a custom component), the submit button may not actually be disabled. Verify that `ActionButton` respects this prop. If it only uses `loading` to disable, the form could be submitted in an invalid state.

**7. `linkedEntrance` linking fires even on document update**

```jsx
if (isNewDocument && linkedEntrance && documentState.createdDocument) {
  dispatch(linkDocumentToEntrance({...}));
}
```

The `isNewDocument` guard is good, but the effect depends on `documentState.latestHttpCode` which persists across renders. If the user navigates away and back, or if the reducer isn't reset, this could re-fire. Consider adding a ref flag (`hasLinked`) to ensure it only fires once.

---

### Minor

**8. File name validation rejects files with multiple dots**

```jsx
if (nameArray.length !== 2) { // file.name.split('.')
  errors.push(...);
  return false;
}
```

This rejects files like `photo.2024.jpg` or `my.cave.map.pdf`. This was pre-existing behavior, but since this PR extracts it into a testable helper, it's a good time to fix it. Consider splitting on the *last* dot instead: `const ext = name.slice(name.lastIndexOf('.') + 1)`.

**9. Unused `IS_MODIFIED` and `IS_INTACT` imports**

`AddFileForm/index.jsx` no longer uses `IS_MODIFIED` or `IS_INTACT` (the `undoRemove` and `updateFileName` functions were removed), but `IS_DELETED` and `IS_NEW` are still imported from `FileHelpers`. The old `FilesList` component that used them is deleted. Verify these constants aren't dead exports now.

**10. `DocumentTypeSelect` — `DocumentType.propTypes` reference removed but no replacement**

The old file had `DocumentType.propTypes = idNameTypeExtended(...)` which seems like it was for a sub-component. The new file doesn't define propTypes for the main `DocumentTypeSelect` component itself (no props needed since `helperText` was removed). This is fine, just noting the cleanup is complete.

**11. Hardcoded `'000'` as unknown language sentinel**

```jsx
const userLanguageId = languages.find(l => l.part1 === locale)?.id ?? '000';
```

The `'000'` value appears in multiple places as a magic string. Consider extracting it as a named constant (e.g., `UNKNOWN_LANGUAGE_ID`) for clarity.

**12. `Entry/Documents/index.jsx` — trailing empty fragment**

```jsx
    </>
  );
```

The component wraps content in `<>...</>` but the `<ScrollableContent>` is the only child now (the old structure was a single element). The fragment is unnecessary.

**13. Translation key `'Moving Image'` → video rename**

The PR description mentions renaming "Moving Image" to a language-appropriate "video" word. Verify the `DocumentTypes.MOVING_IMAGE` constant (`'Moving Image'`) still matches what the API returns, since the translation key in `en.json` is what gets displayed, not the constant value.

---

### Suggestions (non-blocking)

- **Accessibility**: The `DropZone` has `role="button"` and `tabIndex={0}` which is good. Consider adding `aria-label` describing the drop zone purpose for screen readers.
- **`PublicationDatePicker`**: The new `ToggleButtonGroup` approach is cleaner. Consider adding `aria-label="Date precision"` to the group for accessibility.
- **`EntityPicker`**: The `onTouchEnd={e => e.currentTarget.blur()}` is a mobile UX hack — consider documenting why it's needed (presumably to remove focus ring after tap).
- **Test coverage**: This is a large behavioral change with no new tests. Consider at minimum unit tests for `validateAndBuildFileEntries` and `checkFormValidation`.

urien (approved):
En tant qu'utilisateur cette solution m'enthousiasme. Je vais partager avec les membres de l'association.

ClemRz (changes requested):
### Issues (Must Fix)

- **[pages/EntityCreation/entityConfig.jsx:10-22]** Three `descriptionKey` values used in `formatMessage({ id: ... })` don't exist in `en.json`:
  - `'Add a cave entrance with its location and characteristics'` — not in en.json
  - `'Add an image, topographic drawing, book, dataset, bibliography reference, etc.'` — not in en.json (existing key is `'Add an image, topographic drawing or bibliography reference'`)
  - `'Add a massif with its geographical boundaries'` — not in en.json (existing key is `'Define a cave system with its geographical boundaries'`)

  These will render as raw key strings at runtime. Either add the new keys to `en.json` (and all other language files), or reuse the existing keys.

- **[AddFileForm/index.jsx:73-76 — `AuthDocSelect`]** The MUI `<Select>` uses reference equality when `value` is an object. Here `value` comes from `document.authorizationDocument` (Redux/context state) while `MenuItem` values are objects from `authDocs` (a different array). After a re-render or state rehydration, the references won't match and the select will appear empty even though a value was previously chosen. Consider using a scalar value (e.g., `value={value?.id ?? -1}`) and resolving the full object in `onChange`:
  ```jsx
  <Select
    value={isLoading ? -1 : (value?.id ?? -1)}
    onChange={e => {
      const doc = authDocs.find(d => d.id === e.target.value);
      onChange(doc);
    }}>
    ...
    {authDocs.map(doc => (
      <MenuItem key={doc.id} value={doc.id}>
        {doc.title}
      </MenuItem>
    ))}
  </Select>
  ```

- **[Document/index.jsx:105-108 — `onSubmitAnotherArticle`]** After calling `resetContext()`, the subsequent `updateAttribute` calls set values on the *reset* state. However, `resetContext` now accepts an `overrides` object (great fix from the previous review). Consider using it here too for consistency and atomicity:
  ```jsx
  const onSubmitAnotherArticle = () => {
    dispatch(resetDocumentApiErrors());
    resetContext({
      language: document.language,
      type: document.type,
      editor: document.editor,
      library: document.library,
      parent: document.parent,
      datePublication: document.datePublication
    });
  };
  ```

- **[Provider.js:73-74]** The validation for `ISSUE` requires `document.parent` but doesn't require it for `ARTICLE`. Looking at `FormContent.jsx`, the `DocumentAutoComplete` for parent is rendered with `required` for both `isArticle` and `isIssue`. Consider adding `isArticle` to the parent validation check in `checkFormValidation` so the form can't be submitted without a parent for articles either — or remove `required` from the UI for articles if it's truly optional.

- **[FormContent.jsx:109-115]** For the simple media layout, `LanguageSelect` defaults to `userLanguageId` when `document.mainLanguage === '000'`, but this default is only visual — `updateAttribute('mainLanguage', ...)` is never called with the default value. If the user doesn't interact with the language selector, the document will be submitted with `mainLanguage: '000'` (unknown). Consider calling `updateAttribute` with the default on mount, or handling `'000'` as the user's locale on the server side. The same pattern appears in the other layout branches.

- **[AddFileForm/index.jsx:130-135]** The `useEffect` that auto-fills authors depends on `updateAttribute` in its dependency array. Since `updateAttribute` is wrapped in `useCallback` with `[setDocument]` as deps, and `setDocument` is stable, this is fine. But the effect also runs when `option` changes from one non-author value to another non-author value (e.g., from `LICENSE_IN_FILE` to `DOCUMENT_AUTHORIZE_TO_PUBLISH`), clearing authors each time. If a user adds authors, then switches between the two non-author radio options, their author list gets wiped. Consider only clearing authors when switching *to* `AUTHORIZATION_FROM_AUTHOR` or when `option` becomes null.

- **[hooks/useDocumentTypes.js]** The `useDocumentTypes` hook returns plain functions (not hooks) and doesn't use any React hooks internally. It could be a simple object export rather than a hook, which would avoid the overhead of creating a new object on every render. Not blocking, but worth noting for a future cleanup.

- **[Entry/Documents/index.jsx:56-60]** The "New" button navigates to `/ui/entity/add/document?entranceId=${entranceId}`. If the user is not authenticated, they'll see the auth gate in `EntityCreation/index.jsx`. However, after logging in (via the dialog), they won't be automatically redirected to the document creation page with the `entranceId` param. The `NewEntityButton` component handles this by showing the login dialog, but here the navigation happens unconditionally. Consider using the same pattern as `NewEntityButton` (check auth first, show login dialog if not authenticated).

### Nitpicks (Optional)

- **[FormContent.jsx:262-270]** The BBS helper text still references `chapter_and_geo_1_2008.pdf` (the 2008 version). Per the PR discussion, @urien mentioned the BBS documentation has moved to the GrottoCenter wiki and now uses the 2025 version. This might be intentional to keep scope limited, but worth tracking.

- **[DocumentTypeSelect.jsx:195]** The `handleSelect` function passes `dt.name` as the type. The `document.type` field stores the string name (e.g., `'Image'`), which is then compared against `DocumentTypes.IMAGE` (also `'Image'`). This works, but it means the type stored in context is a display string rather than an ID. If the API ever returns localized type names, this could break. Low risk given the current architecture, just noting.

- **[FileSelectorInput/index.jsx]** The `extensions` prop is rendered sorted (`.sort()`) which is nice, but `.sort()` mutates the array in place. Since `extensions` comes from `DOCUMENT_TYPE_ACCEPT` (a module-level constant), this mutates the constant's array on first render. Consider `[...extensions].sort()` — oh wait, it already does that. Disregard.

- **[sync-with-en.js]** The previous version had BOM stripping in `loadJson`. The new version removed it (just uses `JSON.parse(fs.readFileSync(...))`). If any language files still have BOMs, this could cause parse errors. The `update-en.js` script still strips BOMs (`.replace(/^\uFEFF/, '')`). Consider keeping the BOM strip in `sync-with-en.js` as well for robustness.

ClemRz (commented):

- **[hooks/useAuthNavigate.js:14]** The `onBeforeNavigate` callback is included in the `useEffect` dependency array. If a caller ever passes an inline arrow (e.g., `useAuthNavigate('/path', { onBeforeNavigate: () => doSomething() })`), the effect will re-run on every render. Currently both call sites are safe (one passes nothing, one passes a `useCallback`-wrapped ref), but the hook's contract is fragile. Consider wrapping `onBeforeNavigate` in a ref to make the hook resilient to unstable callbacks:
  ```js
  const onBeforeNavigateRef = useRef(onBeforeNavigate);
  onBeforeNavigateRef.current = onBeforeNavigate;

  useEffect(() => {
    if (isAuth && waitingForAuth.current) {
      waitingForAuth.current = false;
      onBeforeNavigateRef.current?.();
      navigate(to);
    }
  }, [isAuth, navigate, to]);
  ```

- **[FormContent.jsx:79-80]** The `useEffect` that initializes `mainLanguage` from `userLanguageId` is suppressing the exhaustive-deps lint rule and only depends on `[userLanguageId]`. If the component remounts (e.g., after "Submit another document" resets the context), `document.mainLanguage` will be `'000'` again but `userLanguageId` won't have changed, so the effect won't re-fire. Consider adding `document.mainLanguage` to the dependency array (with a guard to avoid overwriting user selections):
  ```js
  useEffect(() => {
    if (document.mainLanguage === '000' && userLanguageId !== '000')
      updateAttribute('mainLanguage', userLanguageId);
  }, [document.mainLanguage, userLanguageId, updateAttribute]);
  ```

- **[AddFileForm/index.jsx:127-133]** The `useEffect` that auto-fills authors depends on `updateAttribute` in its dependency array. While `updateAttribute` is stable (wrapped in `useCallback`), the effect also fires on mount when `option` is already `AUTHORIZATION_FROM_AUTHOR` (e.g., after a parent document pre-sets it). This means switching back to the form after a navigation could overwrite manually-added co-authors. If this is intentional, a brief comment explaining the design choice would help future readers.

- **[Entry/Documents/index.jsx:56-62]** The "New" button navigates directly to `/ui/entity/add/document?entranceId=...` without an auth check. The `EntityCreation` route-level auth gate handles the unauthenticated case, but after the user logs in via the dialog shown there, they stay on the same URL and the form appears — which works. However, this is a different pattern from `NewEntityButton` (which uses `useAuthNavigate`). Consider using `useAuthNavigate` here too for consistency, so the login dialog appears in-place rather than navigating to a page that shows an error first.

- **[actions/Document/CreateDocument.js]** This action sends `getState().login.authorizationHeader` but doesn't use `checkAuthStatus(dispatch)` from `actions/utils.js`. Per project conventions, authenticated actions should use this utility to handle 401 responses (dispatch logout + reload). This is pre-existing and not a regression from this PR, but since the file was modified, it's a good opportunity to align it with the convention. The custom 401 handling in the `switch` statement would be replaced by the standard pattern.

### Nitpicks (Optional)

- **[hooks/useDocumentTypes.js]** The file is still named `useDocumentTypes.js` (hook naming convention) but now exports a plain object `documentTypeHelpers` rather than a hook. Consider renaming to `documentTypeHelpers.js` to avoid confusion — the `use` prefix conventionally signals a React hook.

- **[Provider.js:120]** The `useMemo` dependency array includes `setLinkedEntrance` (a state setter from `useState`). React guarantees state setters are stable, so it's harmless but unnecessary noise in the deps array.

- **[FormContent.jsx:262-270]** The BBS helper text now links to the GrottoCenter wiki (good), but the old `ssslib.ch` PDF link and 2008 reference are gone. Per the PR discussion, @urien confirmed the wiki is the correct source. Just noting this is intentional per the conversation.

- **[translations.yml:42-52]** The inline `node -e` script in the workflow uses a BOM character literal (``) in the `replace()` call. This works but is invisible and fragile — if the file is re-saved by an editor that strips invisible characters, the BOM handling breaks silently. Consider using the unicode escape `\uFEFF` instead (though this would require switching to a temp script file given the CLI usage rules).

ClemRz (changes requested):
### Issues (Must Fix)

- **[actions/Document/CreateDocument.js:94-95]** The `.catch` block re-throws non-auth errors (`if (!err.isAuthError) throw err`). However, the `throw new Error(...)` on line 85 (for 4xx/5xx responses) will be caught here and re-thrown, producing an unhandled promise rejection. The failure action has already been dispatched before the throw, so the re-throw just creates console noise. The project convention (visible in every other action) is `if (error.isAuthError) return;` — which silently swallows auth errors and lets the failure dispatch handle everything else. Consider aligning with the convention:
  ```js
  .catch(err => {
    if (err.isAuthError) return;
  });
  ```

- **[AddFileForm/index.jsx:222-224]** The translation key `'You are the author of this document (license CC-BY-SA applies).'` does not exist in `en.json`. The existing key is `'You are the author of this document'`. This will render as the raw key string at runtime. Either add the new key to all language files, or use the existing key and append the license mention as a separate element.

- **[FormContent.jsx:77-78 — simple media layout has no `required` on LanguageSelect]** For the simple media layout, `LanguageSelect` is rendered without `required`. Since the `useEffect` on line 73 sets `mainLanguage` from the user's locale on mount, this works in practice. However, if the user explicitly selects the disabled `'000'` placeholder (which shouldn't be possible since it's `disabled`), or if `userLanguageId` is `'000'`, the document could be submitted with `mainLanguage: '000'`. Consider adding `required` to the `LanguageSelect` in all layout branches for consistency with the full document layout.

- **[hooks/documentTypeHelpers.js — `STILL_IMAGE` not in `SIMPLE_MEDIA_TYPES`]** `DocumentTypes.STILL_IMAGE` is defined but not included in `SIMPLE_MEDIA_TYPES`. If the API returns this type, it will fall into the "full document" layout branch (requiring description, etc.). This may be intentional, but given that `IMAGE` is in the simple set, `STILL_IMAGE` seems like a candidate too. Worth confirming whether this type is actually used by the API.

- **[DocumentTypeSelect.jsx:195-197 — `handleSelect` resets form on re-selecting the same type]** If the user clicks the already-selected type card, `handleSelect` fires `resetContext(...)` which clears all fields except title/language/type. Consider adding a guard: `if (newDocType === document.type) return;`

- **[FileSelectorInput/index.jsx — DropZone click handler doesn't prevent action when disabled]** The `onClick` handler on the DropZone checks `!disabled` before triggering the file input, which is good. However, the drag-and-drop handlers (`handleDrop`, `handleDragEnter`, etc.) don't check `disabled`. A user could still drop files onto a disabled drop zone. Consider adding an early return in `handleDrop` when `disabled` is true.

- **[Provider.js:73-74 — ARTICLE parent validation]** `checkFormValidation` now correctly requires `parent` for both `ARTICLE` and `ISSUE` (good fix from previous review). However, the `STILL_IMAGE` type is not in `DESCRIPTION_OPTIONAL_TYPES` — if it's ever selected, description will be required. This is consistent with the "full layout" rendering, just noting for awareness.

- **[Entry/Documents/index.jsx:56-60 — "New" button visible but non-functional for non-edit users]** The "New" button is gated by `permissions.isAuth && isEditAllowed`. The `useAuthNavigate` hook is called unconditionally at the top of the component (line 30), which is correct for hooks rules. However, if `isEditAllowed` is false, the button won't render anyway, so the hook call is harmless but slightly wasteful. Not blocking, just noting.

### Nitpicks (Optional)

- **[DocumentTypeSelect.jsx:46-65 — `TYPE_ICONS` creates JSX elements at module level]** These icon elements are created once at module load time and reused across renders. This works fine in React (elements are immutable descriptors), but it's an unusual pattern. If this ever causes issues with React DevTools or testing, extracting them into a function would be the fix.

- **[FormContent.jsx:96-100 — `addFileFormProps` object recreated every render]** The `addFileFormProps` object is created inline on every render. Since `AddFileForm` doesn't use `React.memo`, this doesn't cause extra re-renders, but if memoization is ever added, this would defeat it. Consider `useMemo` if performance becomes a concern.

- **[PagesEditor.jsx:39 — `!pages.match(pages)` is always false]** This is pre-existing, but since the file was touched: `!pages.match(pages)` matches the string against itself as a regex, which always succeeds, making the `!` always false. The condition effectively reduces to just `newV.match(reg)`. The intent was likely `!newV.match(reg)` or something else. Low priority since behavior is unchanged.

- **[translations.yml:49 — BOM literal in inline script]** The inline `node -e` script uses an invisible BOM character literal in the `.replace()` call. If the workflow file is ever re-saved by an editor that strips invisible characters, BOM handling will silently break. The `sort.js` and `sync-with-en.js` scripts correctly use `\uFEFF` — consider aligning the workflow script (would require moving to a temp file per CLI conventions).

ClemRz (commented):
### Summary

Solid iteration. The previous review items (auth handling in `CreateDocument`, `AuthDocSelect` reference equality, `resetContext` atomicity, `createdDocument` reset, `useAuthNavigate` callback stability, `documentTypeHelpers` rename, same-type reset guard, `FileSelectorInput` disabled guard, BOM handling in scripts) have all been addressed. The code is in much better shape.

A few remaining observations below — nothing blocking.

- **[FormContent.jsx:73-74]** The `useEffect` that initializes `mainLanguage` from `userLanguageId` depends on `[document.mainLanguage, userLanguageId, updateAttribute]`. After "Submit another document" resets the context, `document.mainLanguage` goes back to `'000'` and `userLanguageId` hasn't changed — but since `document.mainLanguage` is in the deps array, the effect will re-fire correctly. Good. However, for the **simple media layout** (line 131), `LanguageSelect` is rendered without `required`. If the user somehow clears the language (unlikely given the Select UX, but possible via browser autofill or extensions), the form could submit with `'000'`. Consider adding `required` to `LanguageSelect` in the simple media branch for consistency with the other layout branches.

- **[Provider.js:87-93]** The `requiresAuthorization` check (`document.type !== DocumentTypes.AUTHORIZATION_TO_PUBLISH`) means that for *all* other document types, if files are present, `selectOptionAuthorizationDocument` must be set. This is correct. However, for the **Event** layout in `FormContent.jsx`, there is no `<AddFileForm>` rendered — so users can never add files to an Event. If the API ever allows files on Events, the validation would block submission without a visible radio group. This is fine for now, just noting the implicit coupling between layout and validation.

- **[AddFileForm/index.jsx:128-133]** The author auto-fill effect fires when `option` changes. If a user selects "Author" (auto-fills), then switches to "License in file" (authors are NOT cleared — good fix from previous review), then switches back to "Author", the effect will overwrite any manually-added co-authors with just the logged-in user. The PR comment thread confirms this is intentional product behavior. A brief inline comment documenting this decision would help future readers (the existing comment on line 126 covers the mount case but not the "switch back" case).

- **[UpdateDocument.js:73-93]** The main `updateDocument` function still doesn't use `checkAuthStatus(dispatch)` — it handles 401 manually in the switch statement. This is pre-existing and not a regression, but since `CreateDocument.js` was aligned with the convention in this PR, it creates an inconsistency between the two sibling actions. Worth a follow-up.

### Nitpicks (Optional)

- **[translations.yml:49]** The inline `node -e` script uses an invisible BOM character literal (``) in the `.replace()` call. The `sort.js` and `sync-with-en.js` scripts correctly use `\uFEFF`. If the workflow file is ever re-saved by an editor that strips invisible characters, BOM handling will silently break. Low risk since CI would catch the parse failure, but worth aligning in a future cleanup.

- **[DocumentTypeSelect.jsx:46-65]** `TYPE_ICONS` creates JSX elements at module level. This works fine (React elements are immutable descriptors), but it's an unusual pattern. If it ever causes issues with testing or HMR, extracting them into a function would be the fix.

- **[PagesEditor.jsx:20-21]** The `useEffect` on mount calls `setIsFormatError(!isDocumentPagesFormatValid(pages))` with an empty deps array. This means if `pages` changes between render and effect execution (unlikely but theoretically possible with concurrent features), the initial error state could be stale. Adding `pages` to deps (with a guard to avoid re-running on every keystroke) would be more robust, but this is very low priority.

- **[Document.jsx:44]** `useLayoutEffect` is used to detect text clamping. This is correct for measuring DOM dimensions, but it will emit a warning during SSR (if SSR is ever added). Since this is a SPA with no SSR plans, this is fine.

---

Overall this is a well-structured refactor with good separation of concerns (type helpers extracted, file validation isolated, auth navigation reusable). The card-based type selector and simplified layouts are a clear UX improvement. Nice work.

ClemRz (approved):

PR #1309, opened 2026-05-07
feat(Entrance): add detach entrance from cave action

ClemRz:
# feat(Entrance): add detach entrance from cave action

## 🤔 What

Adds a "Detach entrance" button to the move entrance page (`/ui/entrances/:id/move`) that allows users to separate an entrance from its current cave with a single click.

- Closes #1308
- The button transparently creates a new cave (named after the entrance) and moves the entrance to it
- If the move step fails, the created cave is rolled back (deleted)
- The button is disabled with a tooltip when the entrance is the only one of its cave
- After a successful detach or move, the entrance data is re-fetched to reflect the updated state

## 🤷‍♂️ Why

Users need the ability to detach an entrance from a multi-entrance cave into its own standalone cave. The backend already supports this via two endpoints (POST cave + PATCH move), but there was no frontend action for it.

## 🔍 How

- New Redux action creator (`DetachEntrance.js`) orchestrates the two-step API process with rollback on failure
- New reducer (`DetachEntranceReducer.js`) manages loading/success/error state
- New component (`DetachEntranceSection.jsx`) renders the button with disabled/tooltip/loading/success/error states
- Integrated below the existing "move to existing cave" form, separated by a divider
- Added `fetchEntrance` dispatch after success in both the detach and existing move actions to refresh stale state
- Translation keys added to all 15 language files with proper translations

## 🔗 Related API PR

None — the backend endpoints (`POST /api/v1/caves` and `PATCH /api/v1/entrances/:id/cave/:caveId`) already exist.

## 🧪 Testing

1. Navigate to `/ui/entrances/:id/move` for an entrance that belongs to a multi-entrance cave
2. Click "Detach entrance" — should show loading, then success message, and the page should refresh with the new cave data
3. For an entrance that is the only one of its cave, the button should be disabled with a tooltip
4. Property-based tests cover the action creator orchestration logic (payload, correct cave ID, rollback, error propagation)
5. Unit tests cover the component states and reducer transitions

## 📸 Previews

Screenshot 2026-05-07 at 3 34 15 p m

Screenshot 2026-05-07 at 3 34 30 p m


urien (approved):
C'est parfait ainsi.
Merci @ClemRz

PR #1313, opened 2026-05-12
feat(account): password rules, language sync, map persistence and chip styling fixes

Paul-AUB:
# feat(account): password rules, language sync, map persistence and chip styling fixes

## 🤔 What

- **Password rules**: new `PasswordRules` component showing real-time validation feedback (length, uppercase, number, special char) on sign-up, change-password, and account pages; new `BoolIcon` helper
- **Language sync**: new `useLanguageSync` hook that applies the user's preferred language from their account to the app on login/profile update; adds `languageMapping` utility and updates `AccountReducer`
- **Map persistence**: the map's last center/zoom is saved to `localStorage` so the position is restored on next visit
- **Admin form fix**: corrected `isRequired` flags on the Person entity form for admin users
- **Chip styling**: MUI chip appearance fixes across `AdvancedSearch`, `Person`, `CoordinateDisplay`, `NotificationMenu`, `RegionsList`, `RecentChangesCard`, `PageTabs`, `ScrollableContent`; cleans up overrides from `grottoTheme`

## 🤷‍♂️ Why

Password rules give users clear guidance on what is expected, reducing failed attempts. Language sync ensures the UI matches the user's stored preference automatically. Map persistence improves UX by keeping the user's last explored area. The chip styling fixes align the UI with the MUI v7 API changes.

## 🔍 How

- `PasswordRules` reads rules from `config.js` (`passwordRules` array) and renders a checklist using `BoolIcon`
- `useLanguageSync` watches the Redux account state and calls `i18n.changeLanguage` + dispatches a language action when the preference changes; mounted in `ApplicationShell`
- `Map.jsx` reads/writes `localStorage` keys `mapLastLat`, `mapLastLng`, `mapLastZoom` on move/zoom events and on initial load
- Chip fixes: replaced deprecated `variant`/`size` prop combinations and removed global theme overrides that conflicted with MUI v7 defaults

## 🔗 Related API PR

_None_

## 🧪 Testing

1. Sign up or change password — the rules checklist should update in real time as you type
2. Set a language preference in your account → log out → log in → the app language should switch automatically
3. Pan/zoom the map, refresh the page — the map should return to the same position
4. Check chip appearance in Advanced Search filters, Person page, Notification menu, Coordinate display

## 📸 Previews

_Add screenshots if available._


ClemRz (approved):

- **[packages/web-app/src/hooks/useLanguageSync.js:43-49]** The third effect (locale → account sync) fires on every `locale` change after mount, including the one triggered by the second effect (account → locale sync). While the guard `languageId !== account?.language` should prevent an unnecessary PATCH in most cases, there's a timing window: if the second effect dispatches `changeLocale`, the third effect runs before `fetchAccount` resolves with the updated account. At that point `account?.language` still holds the old value, so the guard passes and a redundant `updateAccount` is dispatched. Consider adding a ref flag (e.g., `isSyncingFromAccount`) that the second effect sets before dispatching `changeLocale` and the third effect checks before dispatching `updateAccount`.

- **[packages/web-app/public/lang/ (ar, bg, ca, de, el, es, he, id, it, ja, nl, pt, ro)]** The new `password.rule.lowercase`, `password.rule.minLength`, `password.rule.special`, `password.rule.uppercase`, and `password.rules.error` keys are left untranslated (English fallback) in all languages except French. Only `password.rule.digit` is translated in a few languages. Per the project conventions, translations are AI-generated — consider running the translation generation pipeline to fill these in before merging, or at least noting this as a follow-up task.

### Nitpicks (Optional)

- **[packages/web-app/src/components/common/Form/BoolIcon.jsx:18-22]** `defaultProps` is deprecated in React 19 for function components. Consider using JavaScript default parameters instead: `const BoolIcon = ({ value, fontSize = 'small', sx }) => ...`. Same applies to `PasswordRules.jsx` (RuleItem's `defaultProps`).

- **[packages/web-app/src/components/common/CoordinateDisplay/index.jsx]** The `computePrecisionSeverity` logic change (previously `null/undefined` → warning, now → error) is a behavioral change that affects existing entrances with unknown precision. Previously they showed a warning chip; now they'll show an error chip. This seems intentional based on the comments, but worth noting in case it surprises users who see previously-yellow indicators turn red.

- **[packages/web-app/src/components/appli/Person/Person.jsx:86-93]** The chip styling uses `color: '#fff'` as an inline override. Since `color="secondary"` is now set on the Chip, MUI should handle the contrast color automatically. The explicit `color: '#fff'` might conflict with themes where the secondary color is light. Consider removing it and letting MUI handle the text color.

ClemRz (approved):

- **[packages/web-app/src/pages/Account/index.jsx:195-197]** `EditActions.defaultProps` is deprecated in React 19. This commit correctly removed `defaultProps` from `BoolIcon` and `PasswordRules` (addressing the previous review), but introduces a new `defaultProps` usage here. Consider using a default parameter value instead:
  ```js
  const EditActions = ({ isLoading, isDisabled = false, onCancel }) => (
  ```

- **[packages/web-app/src/pages/ChangePasswordForm/index.jsx:180-183]** Same concern — `ChangePasswordForm.defaultProps` is deprecated in React 19. Consider using default parameter values in the destructuring:
  ```js
  const ChangePasswordForm = ({
    currentPassword = undefined,
    onCurrentPasswordChange = undefined,
    password,
    ...
  }) => {
  ```

- **[packages/web-app/src/actions/Person/ChangePassword.js:30-33]** The request body is `JSON.stringify()`'d but no `Content-Type: application/json` header is set (unlike `UpdateAccount.js` which does set it). This is a pre-existing issue, but now that `currentPassword` is being sent in the body it's worth fixing to ensure the API correctly parses the payload:
  ```js
  requestOptions = {
    method: 'PATCH',
    body: JSON.stringify({ password, currentPassword }),
    headers: {
      ...getState().login.authorizationHeader,
      'Content-Type': 'application/json'
    }
  };
  ```

### Nitpicks (Optional)

- **Commit message** has a typo: "feat: ask current password **tu** be able to change password" → should be "**to**".

- **[packages/web-app/src/components/common/Form/PasswordRules.jsx:26]** There's a trailing blank line after the `RuleItem.propTypes` block (before the `PasswordRules` component). Minor formatting inconsistency.

PR #1318, opened 2026-05-13
fix(auth): require current password when changing password

Paul-AUB:
# fix(auth): better error handling for password change

## 🤔 What

- Handle 403 response (wrong current password) with a dedicated translated message: "Current password is incorrect."
- Handle 400 response by displaying the API message directly
- Fall back to a generic error message for any other failure
- Add i18n key `"Current password is incorrect."` translated in all 15 languages

## 🤷‍♂️ Why

The API returns a 403 when the current password is wrong. The previous implementation showed a generic error for all failures, giving no useful feedback to the user.

## 🔍 How

Error resolution is done in the `catch` block (not at render time): `passwordError` state is always `null | string`, keeping the `Alert` render trivial — no magic strings or type checks at display time.

## 🔗 Related API PR

_N/A_

## 🧪 Testing

1. Go to **Mon compte → Email & Sécurité → Changer le mot de passe**
2. Enter a wrong current password → should display "Current password is incorrect."
3. Trigger a 400 from the API → should display the API message verbatim
4. Kill the network → should display the generic fallback message

## 📸 Previews

_N/A_


ClemRz (changes requested):
### Issues (Must Fix)

- **[Multiple lang files]** This PR introduces a UTF-8 BOM (byte order mark, `EF BB BF`) to 7 language files that didn't have one on `develop`: `bg.json`, `ca.json`, `de.json`, `en.json`, `fr.json`, `he.json`, `id.json`. This is an unrelated change that pollutes the diff and creates inconsistency (some files like `ar.json` already had a BOM, others didn't — but the fix should be a separate, deliberate PR if desired). BOMs in JSON files can also cause issues with some parsers. Consider removing these BOM additions from this PR.

- **[src/pages/Account/index.jsx:458]** The 400 branch checks `error?.status === 400 && error?.message` and displays `error.message` verbatim. However, if the API returns a 400 with no `body.message`, `checkAndGetStatus` falls back to `new Error(response.status)`, so `error.message` would be the string `"400"` — which is truthy but not a useful user-facing message. Consider tightening the check to ensure the message is actually meaningful:
  ```suggestion
      else if (error?.status === 400 && error?.message && Number.isNaN(Number(error.message)))
        setPasswordError(error.message);
  ```
  Alternatively, you could check `error?.body?.message` directly, which is only set when the API actually returned a JSON body with a message field.

### Nitpicks (Optional)

- **[src/pages/Account/index.jsx:456-461]** The braceless `if/else if/else` with multi-line bodies is a bit hard to scan. While it passes lint, adding braces would improve readability:
  ```suggestion
      if (error?.status === 403) {
        setPasswordError(formatMessage({ id: 'Current password is incorrect.' }));
      } else if (error?.status === 400 && error?.message) {
        setPasswordError(error.message);
      } else {
        setPasswordError(formatMessage({ id: 'An error occurred. Please try again.' }));
      }
  ```

ClemRz (approved):

PR #1320, opened 2026-05-15
feat(account): add subscriptions tab to account page

Paul-AUB:
# feat(account): add subscriptions tab to account page

## 🤔 What

- Move the subscriptions section from the public person page (`/ui/persons/:id`) to a dedicated **Subscriptions tab** in the account page (`/ui/account`)
- The tab is only visible to users with the **Leader** role
- Redesign `SubscriptionsList` component: data-driven sections (Countries / Massifs / Regions) with proper headings and flex chip layout
- Replace `NotificationsIcon` with `NotificationsActiveIcon` on subscription chips for consistency with entity pages (Massif, Region, Country)
- Fix unsubscribe bug: remove optimistic `fadeOut` animation that caused chips to visually disappear even when the API call failed

## 🤷‍♂️ Why

The subscriptions section on the public person page was hidden behind a `isLeader` check and not easily discoverable. Moving it to the account page makes it part of the user's personal settings, where it belongs.

The `fadeOut` removal fixes a real bug: the chip was hidden via a CSS class before the API responded, and if the call failed, the chip stayed hidden despite the subscription still being active in the store.

## 🔍 How

- `Person.jsx` + `pages/Person/index.jsx`: removed subscriptions props, selector and dispatch — public page no longer fetches or displays subscriptions
- `pages/Account/index.jsx`: added `subscriptions` selector, `fetchSubscriptions` dispatch on mount, and a conditional Subscriptions tab (shown only if `isLeader`) positioned between Activities and Documents
- `SubscriptionsList.jsx`: rewritten as a data-driven component mapping over `[countries, massifs, regions]` — removes repeated JSX, drops unused `HalfWidthBox`/`SubscriptionName` sub-components
- `SubscriptionItem.jsx`: switched to `NotificationsActiveIcon`, removed `fadeOut` dependency, `handleUnsubscribe` now dispatches directly

## 🔗 Related API PR

None.

## 🧪 Testing

1. Log in as a user with the **Leader** role
2. Go to `/ui/account` → verify the **Subscriptions** tab appears between Activities and Documents
3. Click the tab → verify Countries / Massifs / Regions sections render correctly
4. Click a chip → verify navigation to the correct entity page
5. Click the × on a chip → verify the chip disappears only after the API responds
6. Go to `/ui/persons/:id` → verify no subscriptions section is shown anymore

## 📸 Previews

_To be added._


urien (approved):
Le fonctionnement est beaucoup plus logique
J'ai réalisé tous les tests et cela fonctionne parfaitement
Comme utilisateur je ne vois pas le rôle des autres utilisateurs

ClemRz (changes requested):
### Issues (Must Fix)

- **[SubscriptionsList.jsx:37]** The `countries` array lacks a null guard, unlike `massifs` and `regions`. If `subscriptions` is `{}` (from the `?? {}` fallback on line 18), destructuring yields `countries = undefined`, and calling `.map()` on it will throw a TypeError.
  ```suggestion
  items: [...(countries ?? [])]
    .map(c => ({ ...c, name: getLocalizedCountryName(c.id, locale, c.name) }))
    .sort((a, b) => a.name.localeCompare(b.name)),
  ```

- **[SubscriptionsList.jsx:22-29]** The `IDLE` status (initial state before `fetchSubscriptions` dispatches) falls into the error branch because the condition is `!== REDUCER_STATUS.SUCCEEDED`. On the first render cycle before the `useEffect` fires, users will briefly see the error alert "An error occurred when getting the subscriptions." Consider handling `IDLE` explicitly — either return `null` or a spinner:
  ```suggestion
  if (
    subscriptionsStatus === REDUCER_STATUS.IDLE ||
    subscriptionsStatus === REDUCER_STATUS.LOADING
  )
    return <CircularProgress />;

  if (subscriptionsStatus !== REDUCER_STATUS.SUCCEEDED)
    return (
      <Alert
        severity="error"
        content={formatMessage({
          id: 'An error occurred when getting the subscriptions.'
        })}
      />
    );
  ```

- **[pages/Account/index.jsx:833]** `fetchSubscriptions(userId)` is dispatched for all authenticated users, but the subscriptions tab is only visible to leaders. Consider guarding the fetch with `isLeader` to avoid an unnecessary API call for non-leader users:
  ```suggestion
  if (userId) {
    dispatch(fetchPerson(userId));
    if (isLeader) dispatch(fetchSubscriptions(userId));
  }
  ```

- **[pages/Account/index.jsx / Person.jsx]** The old Person page allowed admins to unsubscribe other users (`canUnsubscribe = canEdit || permissions.isAdmin`). Now that subscriptions live exclusively on the Account page with `canUnsubscribe` always `true`, admins lose the ability to manage other users' subscriptions. If this is intentional, no action needed — but worth confirming this capability isn't required.

### Nitpicks (Optional)

- **[SubscriptionName.jsx]** This component is now dead code (no longer imported anywhere). Consider removing it in this PR or a follow-up to keep the codebase clean.

- **[util/fadeOut.js]** Similarly, `fadeOut` is no longer imported anywhere in the codebase. Could be cleaned up.

ClemRz (approved):

PR #1322, opened 2026-05-16
feat(search): replace entrance list pages with EntityTable + advanced search

### LINKED ISSUE #844: [Entrances of a massif / a country] Add a filter
Il est désormais possible d'avoir la liste des cavités d'un massif ou d'un pays avec des indications sur la qualité des données attachées à chaque cavité. Il serait très utile de pouvoir réaliser des filtres sur ces cavités

Ajouter un champ "filter" permettant de filtrer les réponses selon ce qui serait indiqué dans ce champ

### LINKED ISSUE #1245: Cave data quality per region is not sorted through pagination
**Describe the bug**
The region page only sort the current page of result for cave data quality.

**To Reproduce**
Go to  https://grottocenter.org/ui/countries/FR/regions/38/entrances

Page 1 gives:

Image

Pages 2 gives :

Image

**Expected behavior**
I want to have the sort by quality **before** the pages, so that the first pages shows the best quality caves

Paul-AUB:
# feat(search): replace entrance list pages with EntityTable + advanced search

closes #844
closes #1245

## 🤔 What

**Core — EntrancesListPage rewrite (`/countries/:id/entrances`, `/regions/:id/entrances`, `/massifs/:id/entrances`):**
- Replace the old custom paginated list (cards sorted by data quality, using `fetchMassifEntrances` / `fetchCountryEntrances` / `fetchRegionEntrances` actions) with the existing `EntitySearchPage` + `EntrancesSearch` advanced search stack
- Users can now search, filter and sort entrances from any massif, country or region page — instead of a static sorted list
- The locked `initialFilter` + `lockedFilter` pattern pre-filters results to the current entity (massif name, country, region) and hides the corresponding filter control

**Mobile card view for EntityTable:**
- New `MobileEntityList` component: results render as clickable cards with entity icon, title, and key fields in a compact label/value layout
- Empty state with `SearchOffIcon` + translated messages
- "Load more" infinite scroll (accumulates rows locally with an `isAppending` ref; resets on new search)

**View toggle:**
- Icon button to switch between card and table view on both mobile and desktop toolbars
- `viewMode` state in `EntityTable`, initialized from `isMobile`

**Responsive desktop toolbar:**
- Result count truncates with ellipsis (`flex: 1; minWidth: 80`)
- "Export to CSV" and "Change columns" buttons truncate with ellipsis instead of overflowing
- CSV label shortens to plain `CSV` below the `md` breakpoint
- `overflow: hidden` on Toolbar prevents horizontal scroll

**Mobile sort control (`MobileToolbar`):**
- Dropdown listing sortable columns + ASC/DESC toggle
- Sort button text truncates with ellipsis on long column names

**`VisibleColumnsMenu` refactor:**
- Always renders as an outlined primary `Button`
- New props: `icon`, `label`, `menuTitle`, `sx` — desktop passes `ViewColumnIcon` + "Change columns", mobile uses `DatasetIcon` + "Data"

**UI/UX polish:**
- `DataQualityBadge`: size reduced from `30` to `24` in entrances config — better fit for table rows and mobile cards
- Boolean cells (`BoolIcon`): icon size reduced from `medium` (24 px) to `small` (20 px) for better density
- `BoolIcon` moved from `common/Form/` to `common/` — it's a display utility, not a form component; `Form/BoolIcon.jsx` removed, consumers updated

**Massif filter fix:**
- Fix HTTP 400 in `EntrancesSearch`: was sending `massifs` (`object[]`, rejected by Typesense) — corrected to `massifs.name` in `EntrancesSearch.jsx` and `EntrancesList/index.jsx`
- Add `massifs` column to entrances EntityTable config (hidden by default, comma-separated names via `keyArray('name')`)

**i18n:**
- New keys in all 14 language files: `"Card view"`, `"Table view"`, `"Clear sort"`, `"Data"`

## 🤷‍♂️ Why

The old entrance list pages (massif, country, region) showed a static data-quality-sorted list with basic pagination and no search or filter. Replacing them with the advanced search EntityTable gives users filtering, sorting, column selection, CSV export, and both table and card views — for free, since the infrastructure already exists.

The mobile card view makes the EntityTable usable on phones — the HTML table was too wide and not touch-friendly.

The massifs filter bug caused an HTTP 400 on any massif entrance page and any advanced search using the Massif field, making those features entirely broken.

## 🔍 How

**EntrancesListPage:**
- Removed: `fetchMassifEntrances`, `fetchCountryEntrances`, `fetchRegionEntrances` action calls, `EntranceList` component, `StyledList`, `sortByDataQuality`
- Added: `EntitySearchPage` + `EntrancesSearch` with `initialFilter` and `lockedFilter` derived from URL params
- For massifs: a `fetchFieldSearch` on `massifs.name` resolves the massif's Typesense value before locking the filter

**MobileEntityList:**
- `MobileEntityCard`: `CardActionArea` → `navigate(link(doc))`, title auto-detected (`name` or `title` field)
- Body rows use `alignItems: center` so `DataQualityBadge` aligns correctly with caption text

**massifs.name fix:**
- Typesense rejects `filter_by` on `object[]` parent fields — only leaf fields (`massifs.name`, `massifs.id`) are valid
- `sortable: false` on the new massifs column until the API adds `facet: true` to `massifs.name` in the Typesense entrance schema

## 🔗 Related API PR

The massifs sort (`group_by: massifs.name`) will remain broken (HTTP 400 on sort) until the API adds `facet: true` to `massifs.name` in `api/dbSync/entities/entrance.js`. A separate API PR is needed.

## 🧪 Testing

1. Open `/ui/massifs/<id>` → Entrances tab — results appear in the EntityTable with the massif filter pre-applied and locked; no HTTP 400
2. Same for `/ui/countries/<id>/entrances` and `/ui/regions/<id>/<rid>/entrances`
3. Search and filter work within the locked context (e.g. filter by city inside a massif's entrances)
4. On mobile (< 600 px), results render as cards with entity icon and "Load more" button
5. "Load more" appends without resetting; a new search (new filter or query) resets the list
6. Toggle card ↔ table with the icon button — works on both mobile and desktop toolbars
7. Change visible columns via "Data" (mobile) or "Change columns" (desktop) — card fields update
8. Sort on mobile — results reload sorted; ASC/DESC toggle works; long column names truncate with `…`
9. On desktop at narrow width — toolbar truncates instead of overflowing; CSV button becomes "CSV"
10. `DataQualityBadge` is compact and aligned with text in both table rows and mobile cards
11. Boolean fields (e.g. "Has siphons") show small-size icons aligned with values

## 📸 Previews

image

image

image

image

image

image

image


ClemRz (changes requested):
### Issues (Must Fix)

- **[DesktopEntityTable.jsx:500-511]** The `page` prop is missing from `StyledTablePagination`. The old `EntityTable` passed `page={page}` to keep the pagination control in sync with internal state. Without it, MUI's TablePagination will default to page 0 and won't reflect the current page after navigation. This is a regression.
  ```suggestion
  <StyledTablePagination
    showFirstButton
    showLastButton
    rowsPerPageOptions={pageSizeOptions}
    component="div"
    count={nbTotalRows}
    rowsPerPage={rowsPerPage}
    page={page}
    labelRowsPerPage={formatMessage({ id: 'Results per page:' })}
    onPageChange={handleChangePage}
    onRowsPerPageChange={handleChangeRowsPerPage}
  />
  ```

- **[lang/bg.json, ca.json, de.json, fr.json, he.json, id.json]** A UTF-8 BOM (`EF BB BF`) was added to 6 language files (visible in the diff as `{` → `{`). The other language files don't have a BOM, and JSON files should not include one — it can cause parsing issues in some environments and creates an inconsistency across the lang files. This looks like an editor artifact. Please remove the BOM from these files.

- **[EntrancesList/index.jsx:57-65]** `fetchFieldSearch` is called directly (not dispatched) inside a `useEffect` without a cleanup/abort mechanism. If the component unmounts before the promise resolves (e.g., user navigates away quickly), `setCountyValue` will be called on an unmounted component. Consider using an `AbortController` or a `let cancelled = false` flag in the effect cleanup to avoid this.

- **[EntrancesList/index.jsx:67-76]** Same concern for the massif `fetchFieldSearch` call — no cleanup on unmount.

- **[EntitySearchPage.jsx:37-48]** The `initialFilter` prop is excluded from the `useEffect` dependency array (with an eslint-disable). This means if `EntitySearchPage` is rendered with a new `initialFilter` value without changing `entityType`, the initial fetch won't re-run. In the current usage this is fine because `EntrancesListPage` uses `key={searchKey}` to force remounting, but it's fragile — a future consumer might not know about this requirement. Consider adding a comment explaining why `initialFilter` is intentionally excluded and that consumers must use `key` to trigger re-fetches.

- **[DesktopEntityTable.jsx:232-244]** The `handleRowClick` always uses `window.open(url, '_blank')`. The old code had `if (isMobile) navigate(url); else window.open(url, '_blank')` for better mobile UX (in-app navigation instead of opening a new tab). Since `DesktopEntityTable` can now be shown on mobile (via the view toggle), consider restoring the mobile-aware navigation or documenting why it was intentionally removed.

- **[MobileEntityList.jsx:97-104]** The `useEffect` that manages `allRows` uses a mutable ref (`isAppending`) to distinguish "load more" from "new search". This pattern works but is subtle — if `rows` reference changes without the user clicking "Load more" (e.g., a background refetch), it will reset the list. Consider adding a brief comment explaining the contract: `rows` must only change in response to `onPageChange` calls or new searches.

- **[lang/*.json — "Sort by" key]** The `"Sort by"` key is left as English (`"Sort by": "Sort by"`) in 13 out of 14 non-English language files (only `fr.json` has a translation: `"Trier par"`). Since translations are AI-generated per the project conventions, this might just need a translation pass. Worth noting so it doesn't ship untranslated.

### Nitpicks (Optional)

- **[FixedContent.jsx:51]** The condition was changed from `isString(title)` to `title != null`. This is semantically broader — it will now render the `<Title>` wrapper for any truthy value (including `0` or empty string `""`). The `isString` check was more precise. Since `title` is typed as `PropTypes.oneOfType([PropTypes.node, PropTypes.string])`, this is probably fine in practice, but `title !== undefined && title !== null` or just keeping the Ramda check would be more explicit.

- **[EntrancesSearch.jsx:217-219]** The template literal wrapping `formatMessage` is redundant: `` `${formatMessage({ id: 'Ease of reach' })}` `` is equivalent to `formatMessage({ id: 'Ease of reach' })`. The backticks add no value when there's no concatenation.

- **[entitiesConfig.jsx]** The formatting changes (expanding single-line objects to multi-line) are a nice readability improvement but inflate the diff significantly. Consider separating pure formatting changes from functional changes in future PRs for easier review.

- **[EntrancesSearch.jsx:29]** The `DataUsageIcon` import from MUI is used for the data quality slider. `DataUsageIcon` visually represents a pie/donut chart — consider whether a more semantically fitting icon (like `BarChartIcon` or `InsightsIcon`) might better represent "data quality" to users.

ClemRz (commented):
### Follow-up Review (post-fixes)

All issues from the first review have been addressed. The fixes look solid — cleanup flags in effects, `page={page}` restored, BOM removed, `useOpenLink` integrated, `isNewQuery` prop for `MobileEntityList`, eslint-disable comment added, translations filled in, and `FixedContent` condition tightened. Nice work on the `PersonSearch` bug fix too — removing the hardcoded `type: 'AUTHOR'` filter and adding the chip-based type selector is a good UX improvement.

A few remaining observations:

- **[report.md]** This file (the code review response) was committed to the repo. It should be removed before merging — it's not project documentation and will clutter the root directory.

- **[DesktopEntityTable.jsx:382-404 vs 405-432]** Minor inconsistency in the CSV button responsive behavior: the *enabled* button always shows "Export to CSV" (relying on text-overflow ellipsis to truncate), while the *disabled* button explicitly switches to "CSV" on small screens via `isSmall`. Consider applying the same `isSmall` logic to the enabled button for consistent behavior, or removing it from the disabled one since the ellipsis approach already handles truncation.

- **[MobileEntityList.jsx:20]** `MobileEntityCard` uses `useNavigate` + `navigate(link(doc))` for in-app navigation. This is fine for mobile cards, but it's inconsistent with `DesktopEntityTable` which uses `useOpenLink`. If a user toggles to card view on desktop, clicking a card will navigate in the same tab rather than opening a new one. Consider using `useOpenLink` here too for consistency with the desktop table behavior, or document this as an intentional UX choice (cards always navigate in-place).

### Nitpicks (Optional)

- **[EntrancesSearch.jsx:227, 251]** Two redundant template literals remain: `` `${formatMessage({ id: 'Aesthetic' })}` `` and `` `${formatMessage({ id: 'Data quality' })}` ``. These can be simplified to `formatMessage({ id: 'Aesthetic' })` and `formatMessage({ id: 'Data quality' })` respectively (the Depth/Development ones are fine since they concatenate with the unit).

- **[PersonSearch.jsx:23]** Extra blank line between the import block and the `TYPE_OPTIONS` constant. Minor formatting nit.

ClemRz (commented):
### Third Review — Post-Fixes

All issues from the first review and the follow-up have been addressed. The latest commit cleans up the remaining redundant template literals and the extra blank line in `PersonSearch`. `report.md` is gone. The code is in good shape overall.

- **[MobileEntityList.jsx:103-110]** The second `useEffect` (on `rows`) resets `page` to 0 when `isAppending.current` is `false`. However, the first `useEffect` (on `isNewQuery`) also resets `page` to 0. When a new query fires, both effects run: `isNewQuery` triggers first (resetting `allRows`, `page`, and `isAppending`), then `rows` changes and triggers the second effect (which sets `allRows` to the new rows and resets `page` again). This double-reset is harmless but redundant. Consider removing `setPage(0)` from the `else` branch of the second effect since the `isNewQuery` effect already handles it. Not blocking — just a clarity improvement.

- **[EntrancesList/index.jsx:57-76]** The `fetchFieldSearch` calls use `.then(r => { ... }).catch(() => { ... })` with inline braces. While functional, the `catch` silently swallows errors (sets value to `null`). Consider logging the error with `console.error` so debugging is easier if the field search API starts failing silently in production.

- **[EntityTable.jsx:73]** `viewMode` is initialized from `isMobile` (from `react-device-detect`), which is a static value determined at page load. If a user resizes their browser window from desktop to mobile width, the initial view mode won't adapt. This is fine for now since `react-device-detect` is used elsewhere in the project the same way, but worth noting as a future improvement opportunity (could use `useMediaQuery` instead for responsive behavior).

### Nitpicks (Optional)

- **[MobileToolbar.jsx:133-173]** The `<Toolbar>` JSX inside the `return()` is not indented relative to the return statement. Prettier doesn't flag this, but it reads oddly compared to the rest of the codebase where returned JSX is indented one level inside the parentheses.

- **[EntrancesList/index.jsx:18-23]** `getFlagEmoji` is a nice utility. If it's needed elsewhere in the future, it could live in `src/util/` — but fine here for now since it's only used in this file.

- **[Country/index.jsx:51]** The destructuring `const { dataCountry } = useSelector(state => state.statisticsCountry)` is correct but differs from the pattern used in `Region/index.jsx` which destructures as `const { statistics: dataRegion }`. This inconsistency comes from the existing reducers having different state shapes, so it's not something to fix here — just noting it for awareness.

---

Overall this is a solid PR. The refactoring from the old paginated list to `EntitySearchPage` + `EntrancesSearch` is well-executed, the mobile card view is a nice UX addition, and the locked filter pattern is clean. The `useOpenLink` integration, cleanup flags in effects, and `isNewQuery` prop all address the concerns from previous reviews. Good to merge once the API-side `massifs.name` facet is deployed (or accepted as a known limitation for now).

urien (approved):
Comme utilisateur je trouve que tout fonctionne très bien

ClemRz (commented):
### Fourth Review — Final Pass

All issues and suggestions from the three previous reviews have been addressed. The latest commits add the `iso3166` region fix (replacing the fragile `county` field-search approach with the deterministic ISO 3166-2 code), remove the redundant `setPage(0)` in `MobileEntityList`, and fix the `MobileToolbar` indentation. The code is clean and lint-free.

One remaining observation:

- **[EntrancesList/index.jsx:71,113,135]** `valueLabels` is defined, populated (with `{ iso3166: region.name }` for the region case), and passed to `EntrancesSearch` — but `EntrancesSearch` never declares it in its props or uses it. The intent seems to be displaying the human-readable region name ("Ain") instead of the raw ISO code ("FR-01") in the locked filter chip. Currently, the `ActiveFilterChips` component will display `Region: FR-01` for the locked `iso3166` chip, which is technically correct but not user-friendly. Consider either:
  1. Wiring `valueLabels` through to `ActiveFilterChips` (adding a `valueOverrides` prop that replaces the displayed value for specific keys), or
  2. Removing the unused `valueLabels` variable if the raw ISO code display is acceptable for now.

  Not blocking — the feature works correctly either way, and the `FILTER_LABELS` mapping already shows "Region" as the label, so users understand what the chip represents.

### Nitpicks (Optional)

- **[EntrancesList/index.jsx:47]** The `eslint-disable-next-line` for the first `useEffect` (fetching country/region) omits `dispatch` from the deps array. While `dispatch` is stable in practice (Redux guarantees this), including it would satisfy the exhaustive-deps rule without the disable comment. Minor — consistent with how the massif effect on line 52 also disables the rule.

---

Overall this is a well-executed PR. The iso3166 approach for region filtering is more robust than the previous county field-search, the code is well-commented explaining the Typesense field semantics, and all previous feedback has been incorporated cleanly. Good to merge.

ClemRz (approved):

- **[EntrancesList/index.jsx:113 + SearchElements.jsx:642-650]** `valueLabels` is defined and passed through `EntrancesSearch` to `ActiveFilterChips`, but `ActiveFilterChips` never declares it in its props or uses it in its rendering logic. The locked `iso3166` chip currently displays the raw ISO code (e.g. `Region: FR-01`) instead of the human-readable name (`Region: Ain`). Consider either:
  1. Adding a `valueLabels` prop to `ActiveFilterChips` and using it to override `formattedValue` when a key is present in the map, or
  2. Removing the unused `valueLabels` variable from `EntrancesList/index.jsx` and `EntrancesSearch` if the raw ISO code display is acceptable for now.

  This was flagged in the 4th review and appears to still be unresolved in the latest push.

- **[EntityTable.jsx:73]** `viewMode` is initialized from `isMobile` (from `react-device-detect`), which is a static value determined at page load based on user-agent sniffing. If a user resizes their browser from desktop to mobile width, the initial view mode won't adapt. This is consistent with how `react-device-detect` is used elsewhere in the project, but worth noting as a future improvement — `useMediaQuery(theme.breakpoints.down('sm'))` would give truly responsive behavior.

### Nitpicks (Optional)

- **[EntrancesList/index.jsx:18-23]** `getFlagEmoji` is a nice utility. If it's needed elsewhere in the future, it could live in `src/util/` — fine here for now since it's only used in this file.

- **[DesktopEntityTable.jsx:5-6]** Import order: `useOpenLink` (internal hook) is imported before `styled` and `useMediaQuery` (external MUI imports). Per project conventions, external library imports should come first. Minor — not blocking.

---

Overall this is a well-executed PR. The refactoring from the old paginated list to `EntitySearchPage` + `EntrancesSearch` is clean, the mobile card view is a solid UX addition, and the locked filter pattern with ISO 3166-2 codes is robust. All issues from the four previous reviews have been addressed. The code is lint-free and well-organized. Good to merge.

PR #1324, opened 2026-05-17
feat(documents): improve document creation UX, search filters and add Report type

### LINKED ISSUE #269: [DOC SUBMISSION] No ISSN for doc other than Collection
Il pourrait être intéressant de ne pas pouvoir saisir le numéro d'ISSN pour un article ou pour un numéro. 
Les utilisateurs le mette pour la collection et le ressaisisse pour le numéro et pour les articles. Ce n'est pas nécessaire, il faudrait leur indiquer

Paul-AUB:
## 🤔 What

Improvements to the document creation form and document search filters:

- **Authors always required**: the Authors field is now always visible in the document creation/edit form, regardless of whether files are uploaded. Pre-filled with the logged-in user; co-authors can always be added.
- **Document type icons**: icons are now displayed in the document type selector (both in the creation form and in the search page).
- **Consistent BBS subject display**: the subject dropdown in document search now uses the same hierarchical depth-based formatting (indentation + font weight) as the creation form.
- **Search form layout**: document type selector moved outside the advanced filters accordion, placed inline with the search bar (side-by-side on desktop, stacked on mobile). Title, description, and subjects moved into the advanced filters "Content" section.
- **Report document type**: added missing `Report` document type with icon, translations in all supported languages, and correct form layout fallback.

closes #1321

## 🤷‍♂️ Why

- Authors were only settable when a file was uploaded, preventing authorship declaration on file-less documents.
- The document type is a primary filter and should always be visible, not buried in advanced options.
- Inconsistent BBS subject presentation between search and creation was confusing.
- The `Report` type existed in the API but was unhandled in the frontend, falling through to the default layout silently.

## 🔍 How

- Extracted `AuthorsSection` component in `FormContent.jsx`; authors validation is now independent of file uploads.
- `DocumentTypeSelect.handleSelect` now preserves `authors` when switching document type (was clearing them via `resetContext`).
- Extracted `DOCUMENT_TYPE_ICONS` and `DOCUMENT_TYPE_FALLBACK_ICON` into `hooks/documentTypeHelpers.js` as component references (not JSX instances), shared between the creation form and search page.
- Created `hooks/subjectHelpers.js` with `SUBJECT_DEPTH_STYLES`, `getSubjectCode`, and `sortSubjects` — shared between `MultipleSubjectsSelect` and `DocumentSearch`.
- `SearchSelect` now accepts an optional `sx` prop (spread last) to allow per-call overrides without breaking defaults.
- Added `Report` to `DocumentTypes` enum with `SummarizeIcon` and translations for all 14 language files.

## 🔗 Related API PR

None.

## 🧪 Testing

1. Create a document without uploading any file → Authors section is visible and pre-filled.
2. Add a co-author → multiple author chips appear; submission works.
3. Switch document type during creation → authors are preserved.
4. Open document search → document type selector is visible next to the search bar without opening advanced filters.
5. Open advanced filters → Title, Description, and Subjects fields are in the "Content" section.
6. Select a BBS subject in search → hierarchical indentation matches the creation form autocomplete.
7. Create a document of type `Report` → falls into the standard text layout (title, description required, authors visible).

## 📸 Previews

_Screenshots to be added._


ClemRz (commented):

- **[AddFileForm/index.jsx:~160-166]** The `updateOption` logic clears authors when switching to `DOCUMENT_AUTHORIZE_TO_PUBLISH` if the only author is the current user, and re-fills them when switching away. However, if the user has manually added co-authors and then switches to `DOCUMENT_AUTHORIZE_TO_PUBLISH`, the authors are preserved (only cleared when `authorsIsOnlyMe`). This asymmetry might be confusing. Consider adding a comment explaining the intent, or documenting why co-authors are kept in that case but the solo-author is removed.

- **[FormContent.jsx:~56-63]** The `AuthorsSection` effect has an empty dependency array `[]` to run only on mount, but it reads `currentUser.id` which could be undefined if the component mounts before the auth state is hydrated. If `currentUser` is populated asynchronously (e.g., token decode happens after mount), the pre-fill will not trigger. Consider adding `currentUser.id` to the dependency array with a guard to only run once (e.g., a `useRef` flag).

- **[SearchElements.jsx:~544]** The `...sx` spread works fine when `sx` is `undefined` (no-op), but for defensive clarity and to match MUI conventions, consider providing a default parameter value: `sx = {}`.

- **[Translation files]** The old key `"You are the author of this document"` is still present in all language files but is no longer referenced in code. Consider removing it in this PR or a follow-up to avoid dead translation keys accumulating.

### Nitpicks (Optional)

- **[subjectHelpers.js]** The file lives in `src/hooks/` but exports no hooks - it is a pure utility module (constants + functions). Consider placing it in `src/util/` instead. Same applies to `documentTypeHelpers.js` which was already there, so this is more of a pre-existing pattern observation.

- **[DocumentSearch.jsx:~163]** The `docTypeOptions` array is recomputed on every render. Since `documentTypes` comes from Redux and is stable between dispatches, wrapping it in `useMemo` would be a minor optimization.

- **[FormContent.jsx:~49]** `AuthorsSection` is defined inside the same file as `FormContent`. Since it has its own state (via `useEffect`) and could grow, consider extracting it to its own file for maintainability. Not blocking since it is small enough for now.

ClemRz (approved):

- **[Translation files]** The new key `"I hold the publication rights for this content"` is untranslated in 13 out of 15 languages (only `fr.json` has a proper translation). All other files show the English string as the value. This is a user-facing label in the file upload form — consider translating it in this PR or flagging it as a follow-up task.

- **[Translation files]** The old key `"You are the author of this document"` is still present in all 15 language files but is no longer referenced anywhere in code. Consider removing it in this PR to avoid dead keys accumulating.

- **[subjectHelpers.js / MultipleSubjectsSelect.jsx]** The shared `SUBJECT_DEPTH_STYLES` uses different padding values (`pl: 0, 1, 2, 3`) compared to the old `DEPTH_STYLES` in `MultipleSubjectsSelect` (`pl: 1, 1.5, 2, 2.5`). Since `MultipleSubjectsSelect` renders these inside `<Box component="li" {...props}>` where the `li` already has MUI Autocomplete padding, the visual result in the creation form autocomplete will shift left compared to before. This may be intentional for consistency with the search page, but worth verifying visually that the autocomplete dropdown still looks properly indented.

- **[DocumentSearch.jsx:~163]** The `sortSubjects(subjects)` call inside the render (line ~210) creates a new sorted array on every render. Unlike `docTypeOptions` which is wrapped in `useMemo`, this one is not memoized. Consider wrapping it similarly:
  ```javascript
  const sortedSubjects = useMemo(() => sortSubjects(subjects), [subjects]);
  ```
  Then use `sortedSubjects.map(...)` in the JSX.

### Nitpicks (Optional)

- **[en.json:~2134]** Pre-existing issue: `en.json` contains a duplicate key `"You are the author of this document"` (appears twice). JSON parsers silently use the last value, but this is a sign of past merge conflicts. Worth cleaning up alongside the dead key removal.

- **[AGENTS.md]** The BOM-related documentation added to `packages/web-app/AGENTS.md` references PowerShell-specific commands. Since the project targets macOS/Linux (Node 20, Yarn), this guidance seems niche. Consider whether it's needed or if a simpler "use UTF-8 without BOM" note suffices.

- **[DocumentTypeSelect.jsx:~249]** Minor: `{SelectedOtherIcon && <SelectedOtherIcon />}` — since `SelectedOtherIcon` is derived from `selectedOther` which is already checked via the surrounding conditional rendering, the `&&` guard is technically redundant. Not a problem, just a style observation.

urien (approved):
C'est parfait, merci beaucoup @Paul-AUB. Je pense que l'organisation que tu proposes est très pratique et va faciliter le travail des utilisateurs. On est en train de former une équipe de "prospecteurs" qui vont pouvoir tester l'incorporation rapide de photos des entrées (mais aussi les autres fonctionalités qui sont facilitées sur téléphone

PR #1326, opened 2026-05-17
feat(map): create entrance from map right-click with coords pre-filled

### LINKED ISSUE #1148: [MAP] Création d'entité
Il serait intéressant depuis la carte de pouvoir cré&er une entité que l'on connait mais qui ne figure pas sur la carte. Il pourrait être utilise de pouvoir récupérer les coordonnées d'un point sur la carte

Depuis la carte un clic droit sur un point de la carte ouvre un menu contextuel qui permet de créer une entrée, un organisation ou un massfi.
Pour l'entrée ou l'organisation la page de création serait déjà préremplie avec les coordonnées du point où l'on a fait le clic droit
Dans le menu contextel les coordonnées du point s'affichent. Quand on clique dessus ces coordonnées sont copiées dans le presse papier

Paul-AUB:
## 🤔 What

- Add a "Create an entrance here" option to the map right-click context menu, with coordinates pre-filled in the creation form
- Disable the entrance creation "Create" button until all mandatory fields (name, language, latitude, longitude) are filled
- Replace generic `+` icons with entity-specific icons (entrance, massif, organization, bibliography) on "New" buttons across list pages and document tabs

Closes #1148

## 🤷‍♂️ Why

- Creating an entrance from the map is a natural workflow: the user right-clicks a location, selects "Create an entrance here", and lands on the form with coordinates already set
- The "Create" button being always clickable was misleading — users could submit an invalid form (missing required coords, name, language)
- Using entity-specific icons improves visual consistency and helps users immediately identify what they're creating

## 🔍 How

**Map → entrance creation flow (`feat(map)`):**
- Added a `MenuItem` to the existing map right-click `Menu` (MUI) with `EntityIcon iconType="entrance"`
- Handler checks auth via `usePermissions`: if authenticated → `navigate('/ui/entity/add/entrance?lat=…&lng=…')`; if not → opens login dialog and stores the pending URL in state; a `useEffect` on `isAuth` triggers navigation after successful login
- `AddEntrance.jsx` reads `?lat` and `?lng` from `useSearchParams` and passes them as `entranceValues` to `EntranceForm`
- `EntranceForm` merges partial `entranceValues` with `defaultEntranceValues` via spread (`{ ...defaultEntranceValues, ...(entranceValues ?? {}) }`) so other fields keep their defaults

**Required fields guard (`fix(entrances)`):**
- Uses a single `watch([...])` call on all 6 required fields (latitude, longitude, name, language — for both entity type modes)
- Computes `isSubmitDisabled` based on `entityType` and passes it to `FormActionRow`

**Icon consistency (`refactor(icons)`):**
- `NewEntityButton` accepts an optional `icon` prop (defaults to `AddCircleIcon` for backward compatibility)
- All entity list pages (Entrances, Massifs, Organizations, Documents) pass their respective `EntityIcon`
- Entry and Massif document tabs use `EntityIcon iconType="bibliography"` / `AddLinkIcon` consistently

## 🔗 Related API PR

None.

## 🧪 Testing

1. **Map → entrance creation (authenticated):** right-click the map → "Create an entrance here" → form opens with lat/lng pre-filled, button enabled
2. **Map → entrance creation (unauthenticated):** right-click → "Create an entrance here" → login dialog → after login, navigates to the form with coords
3. **Required fields guard:** open `/ui/entity/add/entrance` → "Create" button is disabled → fill name, language, lat, lng → button enables
4. **Icons:** check "New" buttons on `/ui/entrances`, `/ui/massifs`, `/ui/organizations`, `/ui/documents` and document tabs on entrance/massif pages

## 📸 Previews

image

image


ClemRz (commented):
Nice feature — the right-click → create entrance flow is intuitive and well-integrated with the existing context menu. The icon consistency pass and the submit guard are solid improvements too. A few things to consider:

### Issues (Must Fix)

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:107-110]** The `!lat || !lng` check treats `0` as falsy, which would incorrectly disable the submit button when latitude or longitude is exactly `0`. Latitude 0 is the equator — a perfectly valid coordinate. Consider using a stricter check:
  ```suggestion
  const isSubmitDisabled =
    (lat === '' || lat === null || lat === undefined) ||
    (lng === '' || lng === null || lng === undefined) ||
    (entityType === ENTRANCE_AND_CAVE
      ? !caveName || !caveLanguage
      : !entranceName || !entranceLanguage);
  ```
  Alternatively, since the form stores coordinates as strings (from `InputCoordinate`), you could check `lat === ''` for the text-input case and `lat == null` for the unset case. The pre-filled numeric `0` from `parseFloat` would also need handling — something like `(lat ?? '') === ''` won't work for numeric 0 either. A helper like `const isEmpty = v => v === '' || v == null;` applied to lat/lng would be clearest.

- **[packages/web-app/src/components/common/Maps/MapClusters/index.jsx:268-288]** The auth-gated navigation pattern (check `isAuth` → navigate or show login dialog → `useEffect` to navigate after login) is already encapsulated in the `useAuthNavigate` hook (`src/hooks/useAuthNavigate.js`). Consider reusing it to reduce duplication and keep the pattern consistent across the codebase. The challenge is that the URL is dynamic (depends on `contextCoords`), but `useAuthNavigate` could potentially be called with a computed URL, or you could extract a small wrapper. If reuse isn't straightforward here due to the dynamic URL, at minimum a comment explaining why the hook wasn't used would help future readers.

- **[packages/web-app/src/components/appli/Entry/Documents/index.jsx:5-6]** The `EntityIcon` import (internal component) is placed between MUI imports (external libraries). Per project conventions, external library imports should come first, then internal ones. Consider moving it after line 9 (after all `@mui` imports):
  ```suggestion
  import { Box, Button, Divider, Tooltip } from '@mui/material';
  import AddLinkIcon from '@mui/icons-material/AddLink';
  import CancelIcon from '@mui/icons-material/Cancel';
  import { styled } from '@mui/material/styles';
  ...
  import { EntityIcon } from '../../../../pages/EntityCreation/entityConfig';
  ```

### Nitpicks (Optional)

- **[.agents/skills/github-workflow/SKILL.md]** This change (adding a "show and confirm" step to the agent workflow) is unrelated to the entrance-creation feature. Consider splitting it into a separate commit or PR for a cleaner git history.

- **[Multiple lang files: ar.json, el.json, es.json, it.json, ja.json, nl.json, pt.json, ro.json]** The BOM (byte-order mark) removals are a welcome cleanup but unrelated to the feature. Worth noting for reviewers scanning the diff — these are harmless formatting fixes.

- **[packages/web-app/src/components/appli/Massif/Documents.jsx:49]** The label change from "Add" to "Associate" is a nice semantic improvement (the button links an existing document, it doesn't create one). Just flagging it as a behavioral change that's separate from the icon work — users familiar with the old label might need a moment to adjust.

ClemRz (commented):
Nice work addressing the previous review feedback — the `isEmpty` helper correctly handles the `0` coordinate case now. The feature is well-scoped and the auth-gated flow with the comment explaining why `useAuthNavigate` wasn't reused directly is appreciated.

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:107]** The `isEmpty` helper is redefined on every render. Since it's a pure function with no closure over component state, consider extracting it outside the component (next to `defaultEntranceValues` / `defaultCaveValues`) to avoid the minor allocation on each render and make it reusable if needed elsewhere.

- **[packages/web-app/src/components/common/Maps/MapClusters/index.jsx:271-276]** If the user dismisses the login dialog without authenticating, `pendingEntranceUrl` remains set in state. If they later log in through a different flow (e.g., clicking a different auth-gated button), the `useEffect` would fire and navigate them to the stale entrance-creation URL unexpectedly. Consider clearing `pendingEntranceUrl` when the context menu closes (in the `onClose` handler of the `Menu`), or when the login dialog is dismissed without success. The `useAuthNavigate` hook avoids this by using a ref (`waitingForAuth`) that's only set to `true` immediately before showing the dialog, but it has the same theoretical issue.

- **[packages/web-app/src/pages/EntityCreation/AddEntrance.jsx:14-18]** If the URL contains non-numeric query params (e.g., `?lat=abc&lng=def`), `parseFloat` returns `NaN`, and the form would receive `{ latitude: NaN, longitude: NaN }`. The `isEmpty` check in `EntranceForm` wouldn't catch this since `NaN` is not `''`, `null`, or `undefined`. Consider adding a `Number.isFinite()` guard:
  ```suggestion
  const entranceValues =
    lat && lng && Number.isFinite(parseFloat(lat)) && Number.isFinite(parseFloat(lng))
      ? { latitude: parseFloat(lat), longitude: parseFloat(lng) }
      : null;
  ```

### Nitpicks (Optional)

- **[.agents/skills/github-workflow/SKILL.md]** Unrelated to the feature — consider splitting into a separate commit for cleaner git history.

- **[Multiple lang files]** The BOM removals are a welcome cleanup. Just noting for other reviewers scanning the diff that these are harmless formatting fixes unrelated to the feature.

- **[packages/web-app/src/components/appli/Massif/Documents.jsx:49]** The label change from "Add" to "Associate" is semantically more accurate (the button links an existing document rather than creating one). Worth noting as a minor behavioral change separate from the icon work.

ClemRz (approved):
All previous feedback has been addressed cleanly. The `isCoordEmpty` helper correctly handles the `0` coordinate edge case, the `Number.isFinite` guard in `AddEntrance` prevents NaN from reaching the form, and the `pendingEntranceUrl` is properly cleared on menu close. The auth-gated flow comment explaining why `useAuthNavigate` wasn't reused directly is helpful for future readers.

- **[packages/web-app/src/components/common/Maps/MapClusters/index.jsx:355]** The `onClose` inline function `() => { setContextCoords(null); setPendingEntranceUrl(null); }` creates a new reference on every render. While the previous single-setter version did the same, now that it contains two calls and the `Menu` component may use referential equality for optimization, consider extracting it into a `useCallback`. This is minor — MUI's `Menu` doesn't memo-compare `onClose` — but it would be consistent with the other handlers in this component (e.g., `handleContextCopy`, `handlePreferenceChange`).

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:88-95]** The `defaultValues` object is rebuilt on every render of `EntranceForm` because of the spread `{ ...defaultEntranceValues, ...(entranceValues ?? {}) }`. Since `useForm` only reads `defaultValues` on mount, this doesn't cause bugs, but it does allocate a new object each render. Consider wrapping it in `useMemo` if the component re-renders frequently, or simply note that it's harmless since `useForm` ignores subsequent changes to `defaultValues`.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:33]** Nice naming choice — `isCoordEmpty` is more descriptive than the generic `isEmpty` from the previous iteration.

- **[.agents/skills/github-workflow/SKILL.md]** Unrelated to the feature (adds a "show and confirm" step to the agent workflow). Consider splitting into a separate commit for cleaner git history.

- **[Multiple lang files]** The BOM removals are a welcome cleanup. Just noting for other reviewers that these are harmless formatting fixes unrelated to the feature.

Overall this is a well-scoped, well-implemented feature. The right-click → create entrance flow is intuitive, the form validation guard prevents invalid submissions, and the icon consistency pass improves the UX. Good work addressing all prior review feedback. 👍

PR #1327, opened 2026-05-18
fix(DocumentValidation): restore document preview on row click

ClemRz:
# fix(DocumentValidation): restore document preview on row click

## 🤔 What

- Clicking a row in the Document Validation table now opens the "Detailed document view" dialog again, allowing moderators to preview a document before deciding on its validation status.
- Checkboxes are still toggled by clicking the checkbox cell directly.

## 🤷‍♂️ Why

Users reported that they could no longer preview a document before accepting/rejecting it. The root cause was that `EntityTable.handleRowClick` short-circuited to checkbox toggling whenever `onSelected` was provided, making the `onRowClick` handler (which opens the preview dialog) unreachable.

## 🔍 How

**`EntityTable.jsx`** — Changed priority logic in `handleRowClick`: it now only delegates to `handleRowSelect` when `onSelected` is provided but `onRowClick` is *not*. When both are present, `onRowClick` takes priority. Checkboxes still work independently because they have their own `onClick` with `stopPropagation`.

## 🔗 Related API PR

None.

## 🧪 Testing

1. Navigate to Dashboard → Document Validation
2. Click on a document row → the "Detailed document view" dialog should open
3. Close the dialog
4. Click on a checkbox → should toggle selection without opening the dialog
5. Select multiple documents and use the action buttons → should still work as before

### DIFF
--- a/packages/web-app/src/components/appli/AdvancedSearch/SearchResults.jsx
+++ b/packages/web-app/src/components/appli/AdvancedSearch/SearchResults.jsx
@@ -48,7 +48,6 @@ const SearchResults = ({ onSelected, hideExport, entityType, compact }) => {
             }
       }
       onSelected={!onSelected ? null : ids => onSelected(ids, results)}
-      onRowClick={onSelected ? () => false : null}
       compact={compact}
     />
   );
--- a/packages/web-app/src/components/appli/Duplicates/DuplicatesList.jsx
+++ b/packages/web-app/src/components/appli/Duplicates/DuplicatesList.jsx
@@ -89,7 +89,6 @@ const DuplicatesList = ({
           setPage(pageNum);
           setRowsPerPage(pageSize);
         }}
-        onRowClick={() => false}
         onSelected={ids => {
           setSelectedDuplicates(ids);
         }}
--- a/packages/web-app/src/components/appli/EntitiesForm/Document/FormContent.jsx
+++ b/packages/web-app/src/components/appli/EntitiesForm/Document/FormContent.jsx
@@ -1,4 +1,5 @@
 import React, { useCallback, useContext, useEffect, Suspense } from 'react';
+import PropTypes from 'prop-types';
 import {
   Accordion,
   AccordionDetails,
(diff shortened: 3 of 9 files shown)


Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[packages/web-app/src/components/common/EntityTable/EntityTable.jsx:334]** The priority change introduces a regression in `SearchResults` and `DuplicatesList`. Both callers pass `onRowClick={() => false}` together with `onSelected` to suppress row navigation while still relying on `handleRowSelect` for toggling selections. Under the old logic, `if (onSelected)` short-circuited to `handleRowSelect` first, so `onRowClick` was never reached. Under the new `if (onSelected && !onRowClick)` logic, the first branch is skipped (because `onRowClick` is truthy), `onRowClick()` is called and returns `false`, and `handleRowSelect` is never called. Result: clicking a row in those two views no longer toggles the selection — the user must click directly on the checkbox.

   Specific callers affected:
   - `SearchResults.jsx:51` — `onRowClick={onSelected ? () => false : null}`
   - `DuplicatesList.jsx:92` — `onRowClick={() => false}` (always, regardless of selection mode)

   The fix is straightforward: since `() => false` was acting as a no-op guard in those callers (the old code never called it when `onSelected` was set), both callers should simply omit / pass `null` for `onRowClick` when they want pure row-selection behaviour. Alternatively, this PR should also update those two files to remove the `() => false` guard, since the new EntityTable logic makes it obsolete and actively harmful.

2. **[scripts/translations/sort.js:14-18]** `detectIndent` falls back to `'\t'` when no indented line is found. The previous hardcoded default was `2` spaces. A tab-indented output from an edge-case empty file could cause unexpected diffs if CI ever compares the sorted output against an existing 2-space-indented file. Consider defaulting to `'  '` (two spaces) to match the prior behaviour for files that don't yet have indented content.

### Nitpicks (Optional)

3. **[scripts/translations/sort.js:15]** The regex `/^(\s+)"/m` will also match leading whitespace before a *value* on its own line (if any future translation string spans multiple lines). Since current translation files are flat `{ "key": "value" }` objects this is harmless, but a comment noting the assumption would help future maintainers (e.g. `// assumes flat key-value JSON with uniform indentation`).

Paul-AUB (approved):

PR #1328, opened 2026-05-19
Feat/1207 private message UI

dawoldo:
## 🤔 What

Implement a complete frontend private messaging system to allow registered cavers to securely contact each other.
- Adds `/ui/messages` and `/ui/messages/:conversationId` routes.
- Adds a conversation list page with two tabs: Active and Archived conversations.
- Adds conversation detail views displaying message history chronologically (with newer messages at the bottom).
- Adds infinite scroll pagination for older messages in conversation details.
- Adds a "New Message" compose modal featuring autocomplete search (filtering out the current user).
- Adds input validation restricting messages to 5,000 characters (with dynamic character counters).
- Adds a navigation `MessagesIcon` badge displaying the count of unread messages.
- Adds "Archive" and "Unarchive" actions to manage the status of conversation threads.
- Adds message profile link redirects in detail views.
- Adds a direct "Message this caver" button on public caver profiles to start new/open existing conversations.
- Adds email notification preferences switches for private messages and subscriptions in settings.
- Adds inline message reporting that copies message context (sender, ID, body, date) to clipboard and opens the wiki contact form.

## 🤷‍♂️ Why
To allow secure communication between platform members without compromising PII (Only Nicknames and Caver IDs are exposed). Furthermore, it empowers users to organize their correspondence via archiving and control their notification preferences directly.

## 🔍 How
- **Redux Abstraction**: Built `MessagingReducer` and `NotificationPreferencesReducer` and integrated them into the main `GCReducer`. Managed state slice endpoints with statuses (`IDLE`, `LOADING`, `SUCCEEDED`, `FAILED`) and action thunks.
- **Optimistic Updates**: Handled send/reply actions and archive/unarchive transitions optimistically to minimize delay and provide a responsive UI.
- **Layout & Styling**: Implemented using MUI components (`List`, `ListItem`, `Badge`, `Avatar`, `TextField`, `Tabs`, etc.) and customized using standard styled-components matching the GrottoCenter layout rules.
- **Internationalization**: Defined English and French translations for all messaging system components, tooltips, dialogs, and errors in `en.json` and `fr.json`.

## 📸 Previews

image

image

image


ClemRz (commented):
## Code Review — PR #1328: Private Messaging UI

Solid feature implementation overall. The Redux patterns, action structure, and component organization are well done and consistent with the existing codebase. Here are my findings:

### Issues (Must Fix)

- **[ComposeDialog.jsx:27]** The quicksearch reducer stores the error as `error` (singular), but the destructuring uses `errors` (plural). This means `searchErrors` will always be `undefined`, and the `hasError` prop on `AutoCompleteSearch` will never be `true`.
  ```suggestion
  const { results: searchResults, isLoading: isSearchLoading, error: searchError } = useSelector(
    state => state.quicksearch
  );
  ```
  And update the usage on line 178:
  ```suggestion
  hasError={!!searchError}
  ```

- **[Person.jsx:68–80]** `handleMessageClick` awaits `dispatch(fetchConversations(...))` and reads `action.conversations` from the return value. However, `fetchConversations` dispatches a success action with `conversations` as a property, and `dispatch()` returns that action object. If the user has more than 50 conversations, the existing conversation with the target person might not be in the first page, leading to a false negative (opening compose instead of navigating to the existing conversation). Consider using the Redux state after the fetch completes, or calling a dedicated API endpoint to check for an existing conversation with a specific participant.

- **[ConversationDetail.jsx:67]** `theme.spacing(0.5)` uses a non-integer value, which violates the project convention. Use `'4px'` or `theme.spacing(1)` instead.
  ```suggestion
  marginTop: '4px',
  ```

- **[ConversationDetail.jsx:267]** `mb: 0.5` in sx prop uses a non-integer spacing value. Use `mb: 1` or `mb: '4px'`.

- **[ConversationDetail.jsx:288]** `mt: 0.5` in sx prop uses a non-integer spacing value. Use `mt: 1` or `mt: '4px'`.

- **[ConversationDetail.jsx:357]** `mt: 0.5` in sx prop uses a non-integer spacing value. Use `mt: 1` or `mt: '4px'`.

- **[NotificationPreferences.jsx:29]** Hardcoded English fallback string `'Failed to load preferences'`. All user-facing text must use `formatMessage`. Same issue on line 38 with `'Failed to update preferences'`.
  ```suggestion
  return <Alert severity="error">{error || formatMessage({ id: 'An error occurred while fetching conversations.' })}</Alert>;
  ```
  (Or add a dedicated translation key for preference loading errors.)

- **[ArchiveConversation.js / UnarchiveConversation.js]** These actions silently swallow errors with only a `console.error`. Unlike the other messaging actions that dispatch `_FAILURE` actions, these don't dispatch any failure action, so the UI has no way to inform the user that the operation failed. The conversation will appear to be archived/unarchived (optimistic update in the reducer) but the server state won't match. Consider adding `ARCHIVE_CONVERSATION_FAILURE` / `UNARCHIVE_CONVERSATION_FAILURE` action types and reverting the optimistic update on failure.

- **[DataHomepage/index.jsx:54–56]** Fetching unread message count inside the DataHomepage component (which renders on the landing page) is an unusual placement. The `MessagesIcon` component in the AppBar already dispatches `fetchUnreadMessageCount` on mount. This creates a duplicate fetch on every homepage visit. Consider removing it from DataHomepage since the AppBar icon already handles it for all authenticated pages.

- **[ConversationDetail.jsx:332]** `onKeyPress` is deprecated in React. Use `onKeyDown` instead for the Enter-to-send behavior.
  ```suggestion
  onKeyDown={e => {
    if (e.key === 'Enter' && !e.shiftKey) {
      e.preventDefault();
      handleSend();
    }
  }}
  ```

- **[Person.jsx:160–170]** The "Message this caver" button is shown when `!canEdit && userId` — meaning any logged-in user can see it on any other person's page. However, the condition doesn't check whether the viewed person is actually a registered user (has an account). If the person is just an "author" entry (no real account), sending a message to them would likely fail at the API level. Consider adding a check for whether the person has a real account (e.g., checking a flag or role).

- **[MessagingReducer.js]** The `SEND_MESSAGE_SUCCESS` case appends the message to `activeConversationMessages.items` unconditionally. If the user sends a message from the ComposeDialog (creating a new conversation), `conversationId` in the message won't match the currently viewed conversation (if any). This could append a message to the wrong conversation's view. Consider guarding with a check that `message.conversation` matches the currently displayed conversation.

- **[index.jsx (Messages page)]** The conversation list and detail are rendered side-by-side on desktop, but on mobile (`xs`), the detail pane is hidden (`display: { xs: 'none', md: 'block' }`). When a user navigates to `/ui/messages/:conversationId` on mobile, they'll see only the conversation list with no way to view the conversation detail. Consider implementing a mobile-friendly navigation pattern (e.g., showing only the detail view when a `conversationId` is present on small screens).

- **[Translation files]** Only `en.json` and `fr.json` were updated. The other 13 language files are missing the new keys. While `react-intl` falls back to the `defaultMessage`, the keys should be added to all language files (with English values as placeholders) and then sorted, per the project's translation workflow. Run:
  ```bash
  yarn translations:sync-with-en <file>
  ```
  for each language file.

- **[Issue comment context]** The issue owner (ClemRz) commented "Please ignore anything related to 'Alert for news'", but the `NotificationPreferences` component includes the `alert_for_news` toggle. Clarify whether this should be included in this PR or deferred.

### Nitpicks (Optional)

- **[ComposeDialog.jsx:189]** `inputProps` is deprecated in MUI v7 TextField. Use `slotProps={{ htmlInput: { maxLength: 5100 } }}` instead. Same applies to `ConversationDetail.jsx:339`.

- **[ConversationDetail.jsx:269–274]** The sender name link uses a raw `<a>` tag instead of React Router's `<Link>` component. Since this is an internal route (`/ui/persons/:id`), using `<Link>` would enable client-side navigation and avoid a full page reload. However, since `target="_blank"` is used, this is acceptable.

- **[en.json / fr.json]** The diff shows that previously unsorted keys (email verification related) were moved to their correct alphabetical positions. This is good housekeeping but is unrelated to the messaging feature — worth noting for clean git history.

- **[GCReducer.js:140]** `notificationPreferences` is placed before `moveEntranceToCave` in the combineReducers call, breaking alphabetical order. Consider placing it after `notifications` for consistency.

ClemRz (changes requested):
## Code Review — PR #1328: Private Messaging UI (Second Pass)

Good progress since the first review. The quicksearch error field, mobile responsiveness, DataHomepage duplicate fetch, SEND_MESSAGE_SUCCESS guard, and archive/unarchive failure dispatching have all been addressed. Here are the remaining findings:

### Issues (Must Fix)

- **[NotificationPreferences.jsx:29]** Hardcoded English fallback strings `'Failed to load preferences'` (line 29) and `'Failed to update preferences'` (line 38). All user-facing text must use `formatMessage`. These were flagged in the first review and are still present.
  ```suggestion
  return <Alert severity="error">{error || formatMessage({ id: 'An error occurred. Please try again.' })}</Alert>;
  ```

- **[EntitiesForm/Person/index.jsx + PersonEdit/index.jsx]** The `isOurAccount` branch in `PersonForm` (email/password change fields + `NotificationPreferences`) is unreachable dead code. `PersonEdit/index.jsx` (line 27-29) redirects to `/ui/account` when `isOurAccount` is true, so `PersonForm` is only ever rendered for admin/moderator editing other users — where `isOurAccount` is always `false`. Either pass `isOurAccount` from the Account page (which has its own form), or remove this branch until it's wired up properly.

- **[ArchiveConversation.js:26 / UnarchiveConversation.js:26]** Missing `if (error.isAuthError) return;` at the top of the catch block. Per project convention (documented in steering and used in 30+ other actions), authenticated actions must skip the failure dispatch on 401 since `checkAuthStatus` already handles logout. Without this, the optimistic rollback fires unnecessarily during the logout/reload cycle.
  ```suggestion
  } catch (error) {
    if (error.isAuthError) return;
    console.error('Archiving conversation failed:', error);
    dispatch({
      type: ARCHIVE_CONVERSATION_FAILURE,
      ...
    });
  }
  ```

- **[SendMessage.js:35]** Same issue — missing `if (error.isAuthError) return;` before `throw error`. Without it, the calling component (ComposeDialog/ConversationDetail) will display a generic error message to the user during the brief moment before the page reloads from logout.

- **[Translation files]** Only `en.json` and `fr.json` have the new messaging keys. The other 13 language files are missing them. Per the project's translation workflow, run:
  ```bash
  for file in packages/web-app/public/lang/*.json; do node scripts/translations/sync-with-en.js "$file"; done
  for file in packages/web-app/public/lang/*.json; do node scripts/translations/sort.js "$file"; done
  ```

- **[en.json / fr.json]** Several translation keys added in this PR are unused in the code: `"Alert for news"`, `"Convert"`, `"Converter"`, `"Coords converter"`, `"Not implemented yet"`, `"Section under development"`. If these are for future work, consider adding them in the PR that actually uses them to keep the diff focused.

- **[Person.jsx:61-87]** `handleMessageClick` paginates through all active conversations (50 at a time) to find an existing conversation with the target person. Each page dispatches `FETCH_CONVERSATIONS_SUCCESS` which replaces the Redux state, causing the conversation list UI to flash if the user has the messages page open. Consider either: (a) using a dedicated API endpoint to check for an existing conversation with a specific participant, or (b) reading from the Redux store after a single fetch rather than relying on the action return value for all pages.

- **[Person.jsx:130]** The "Message this caver" button is shown when `!canEdit && userId && person?.type !== 'AUTHOR'`. This means the button appears on your own profile when viewed by another logged-in user, which is correct. However, it also appears for users who might be soft-deleted or banned (the API would reject the message, but the button still shows). Consider whether the person's status should be checked.

### Nitpicks (Optional)

- **[GCReducer.js:147]** `notificationPreferences` is placed after `notifications` which is good alphabetically, but `messaging` is placed before `moveEntranceToCave` — this is correct alphabetical order. No issue here, just confirming.

- **[ConversationDetail.jsx:326]** The "Load more" button text uses `defaultMessage: 'Load more'` but the key already exists in `en.json`. The `defaultMessage` is redundant (react-intl will use the loaded translations). This is consistent with the rest of the file though, so not blocking.

- **[en.json]** The key `"Unknown"` (capitalized) is added alongside the existing `"unknown"` (lowercase). Make sure the correct one is used in each context — `"Unknown"` is used as a fallback for participant nicknames which makes sense.

ClemRz (changes requested):
## Code Review — PR #1328: Private Messaging UI (Third Pass)

Good progress since the second review. The `isAuthError` handling is now present in all messaging actions, the quicksearch error field is correctly destructured, mobile responsiveness is handled with conditional display, the `SEND_MESSAGE_SUCCESS` guard checks the current conversation, and archive/unarchive failures properly revert the optimistic update. The DataHomepage duplicate fetch has been removed. The `onKeyPress` → `onKeyDown` and `slotProps` suggestions have been adopted. Well done.

### Issues (Must Fix)

- **[NotificationPreferences.jsx — entire file]** The `NotificationPreferences` component is dead code. It is never imported or rendered anywhere in the application. The `PersonEdit` page redirects to `/ui/account` when `isOurAccount` is true, and the `PersonForm` (in `EntitiesForm/Person/index.jsx`) does not import or render `NotificationPreferences`. The Account page (`pages/Account/index.jsx`) has its own notification toggle (`sendNotificationByEmail`) using a different mechanism. Either wire this component into the Account page (replacing or augmenting the existing toggle), or remove it from this PR and add it in a follow-up when the integration point is clear. Shipping unreachable code adds maintenance burden and confusion.

- **[Person.jsx:68–87]** `handleMessageClick` fetches the first 50 active conversations to find an existing conversation with the target person. If the user has more than 50 conversations and the target conversation is beyond page 1, it won't be found — resulting in opening the compose dialog instead of navigating to the existing conversation. This creates a poor UX (duplicate conversations could be created). Consider either: (a) using a dedicated API endpoint to check for an existing conversation with a specific participant (e.g., `GET /api/v1/messages/conversations?participantId=X`), or (b) accepting this limitation and documenting it as a known trade-off with a TODO comment.

- **[Translation files]** Several translation keys added in this PR are unused in the codebase: `"Alert for news"`, `"Convert"`, `"Converter"`, `"Coords converter"`, `"Not implemented yet"`, `"Section under development"`. These keys have no corresponding `formatMessage` or `<Translate>` usage anywhere in the source. Adding unused keys clutters the translation files and creates work for translators. Consider removing them from this PR and adding them when the code that uses them is implemented.

- **[Person.jsx:130]** The "Message this caver" button is hidden when `person?.isBanned` is true, which is good. However, it's still shown for soft-deleted persons (the API would reject the message). If the person data includes a deletion flag, consider hiding the button for deleted persons too.

- **[ConversationDetail.jsx:230–240]** `handleConfirmReport` uses `navigator.clipboard.writeText()` which can fail silently in some browsers (e.g., when the page is not focused, or in non-secure contexts). The `.catch()` only logs to console. Consider showing a user-facing warning (e.g., a snackbar) when the clipboard write fails, so the user knows to manually copy the details.

- **[MessagingReducer.js:120–140]** In `FETCH_CONVERSATION_MESSAGES_SUCCESS`, the unread count is cleared for the conversation in both active and archived lists. If the same conversation ID appears in both lists (which shouldn't happen, but defensively), the `unreadCountToClear` would only use the active one. This is fine in practice but worth noting.

- **[Issue context]** The issue owner commented "Please ignore anything related to 'Alert for news'" — the `NotificationPreferences` component does not include `alert_for_news`, which is correct. However, the translation key `"Alert for news"` is still added to all 15 language files without any code using it. This should be removed per the comment.

### Nitpicks (Optional)

- **[GCReducer.js]** The `combineReducers` object is not in strict alphabetical order — `updatePersonGroups` appears after `updatePerson`, and `countryEntrances`/`countryRegions`/`regionEntrances` are at the end. This is pre-existing, but the new entries (`messaging`, `notificationPreferences`) are placed correctly.

- **[Translation files — BOM removal and `"or"` fix]** The BOM removal from `ar.json`, `el.json`, `es.json`, `it.json`, `ja.json`, `nl.json`, `pt.json`, `ro.json` and the `"or": "or"` → localized translation fix are good housekeeping but unrelated to the messaging feature. Worth noting for git history clarity — these could have been a separate commit.

- **[ConversationDetail.jsx:262]** The `Link` component from `react-router-dom` is imported but used alongside inline `style={{ color: 'inherit', textDecoration: 'underline' }}`. Consider using MUI's `Link` component with `component={RouterLink}` for consistent styling, or use `sx` prop instead of inline styles.

ClemRz (changes requested):
## Code Review — PR #1328: Private Messaging UI (Fourth Pass)

Good progress since the third review. The `isAuthError` handling is correctly implemented in all messaging actions, the quicksearch error field is properly destructured, mobile responsiveness is handled with conditional display logic, the `SEND_MESSAGE_SUCCESS` guard checks the current conversation, archive/unarchive failures properly revert the optimistic update, the DataHomepage duplicate fetch has been removed, `onKeyDown` replaces `onKeyPress`, and `slotProps` replaces deprecated `inputProps`. The translation keys have been synced to all 15 language files. Well done on addressing the prior feedback.

### Issues (Must Fix)

- **[NotificationPreferences.jsx — entire file + actions + reducer]** The `NotificationPreferences` component, its actions (`GetNotificationPreferences.js`, `UpdateNotificationPreferences.js`), and its reducer (`NotificationPreferencesReducer.js`) are dead code. The component is never imported or rendered anywhere in the application. The `PersonEdit` page redirects to `/ui/account` when `isOurAccount` is true, and the Account page has its own `PreferencesSection` that manages `sendNotificationByEmail` through the existing `updateAccount` action — it does not use this component. The reducer is registered in `GCReducer` (adding unused state to the store), and the API routes (`getNotificationPreferencesUrl`, `updateNotificationPreferencesUrl`) are defined but never called. This was flagged in the second and third reviews. Either wire this component into the Account page (replacing or augmenting the existing toggle), or remove the entire feature (component, actions, reducer, API routes, and associated translation keys) from this PR and add it in a follow-up when the integration point is clear. Shipping unreachable code adds maintenance burden and confusion.

- **[Person.jsx:75–86]** `handleMessageClick` fetches the first 50 active conversations to find an existing conversation with the target person. If the user has more than 50 conversations and the target conversation is beyond the first page, it won't be found — resulting in opening the compose dialog and potentially creating a duplicate conversation. This was flagged in the third review with a request to either use a dedicated API endpoint or document the limitation with a TODO comment. Neither has been done. At minimum, add a `// TODO:` comment acknowledging this limitation so it's tracked for future improvement.

- **[Translation files]** Several translation keys added in this PR are unused in the codebase: `"Alert for news"`, `"Convert"`, `"Converter"`, `"Coords converter"`, `"Not implemented yet"`, `"Section under development"`, `"Notification Preferences"`, `"Email notifications for messages"`, `"Email notifications for subscriptions"`, `"Failed to load preferences"`, `"Failed to update preferences"`. The last five are only referenced by the dead `NotificationPreferences` component. The first six have no corresponding `formatMessage` or `<Translate>` usage anywhere. Adding unused keys clutters the translation files and creates unnecessary work for translators. Consider removing them from this PR.

- **[ConversationDetail.jsx:230–240]** `handleConfirmReport` uses `navigator.clipboard.writeText()` which can fail silently in some browsers (e.g., when the page is not focused, or in non-secure contexts). The `.catch()` only logs to console. Consider showing a user-facing warning (e.g., a snackbar via `onError` from `useNotification`) when the clipboard write fails, so the user knows to manually copy the details.

- **[Person.jsx:130]** The "Message this caver" button is hidden for banned persons (`!person?.isBanned`), which is good. However, it's still shown for soft-deleted persons. If the person data includes a deletion flag (e.g., `isDeleted`), consider hiding the button for deleted persons too, since the API would reject the message.

### Nitpicks (Optional)

- **[ConversationDetail.jsx:262]** The `Link` component from `react-router-dom` is used with inline `style={{ color: 'inherit', textDecoration: 'underline' }}`. Consider using the `sx` prop instead of inline styles for consistency with the rest of the file.

- **[GCReducer.js]** The `combineReducers` object is not in strict alphabetical order at the end — `updatePersonGroups`, `countryEntrances`, `countryRegions`, `regionEntrances` are placed after `updatePerson`. This is pre-existing, but worth noting that the new entries (`messaging`, `notificationPreferences`) are placed correctly.

urien:
Quand on veut envoyer un message à une personne bannie ou dont le mel n'est pas valide il n'y a pas d'alerte
image
On peut imaginer que le bannissement prendra fin ou que la personne aura mis un mel valide mais il faudrait sans doute mettre une alerte lorsqu'on envoi un message a une personne qui fait partie de l'une de ces deux catégories

ClemRz (approved):
## Code Review — PR #1328: Private Messaging UI (Fifth Pass)

Excellent progress since the fourth review. The major blockers have been resolved:
- The dead `NotificationPreferences` component, actions, and reducer have been removed and the functionality is now properly integrated into the Account page's `PreferencesSection`.
- The `Person.jsx` conversation lookup now includes a `TODO` comment acknowledging the 50-conversation limitation.
- All messaging actions correctly handle `isAuthError`.
- Archive/unarchive failures properly revert optimistic updates.
- `slotProps` replaces deprecated `inputProps`, `onKeyDown` replaces `onKeyPress`.

The code is in good shape. The remaining items are suggestions and nitpicks — nothing blocking.

- **[ConversationDetail.jsx:345 + index.jsx:226]** `timeZone="UTC"` forces all message timestamps to display in UTC. The rest of the application uses the browser's local timezone (no explicit `timeZone` prop on date formatting). This inconsistency was also flagged by Paul-AUB in the PR comments. Consider removing `timeZone="UTC"` and `timeZoneName="short"` to let `FormattedDate` use the user's local timezone, consistent with the rest of the app.

- **[en.json + all language files]** Five translation keys added in this PR are unused in the codebase: `"Convert"`, `"Converter"`, `"Coords converter"`, `"Not implemented yet"`, `"Section under development"`. No `formatMessage` or `<Translate>` call references them. Consider removing them to keep the diff focused and avoid unnecessary translator work.

- **[Person.jsx:116]** The `canMessage` condition checks `!person?.isDeleted` which is good. However, `person?.type !== 'AUTHOR'` relies on the person data including a `type` field. If the API doesn't always return this field, the button might show for authors. Worth verifying the API response shape includes `type` for all person pages.

- **[ConversationDetail.jsx:230–240]** `handleConfirmReport` now shows a user-facing error via `onError` when clipboard write fails — good improvement. However, `window.open` is called unconditionally after the clipboard operation (even on failure). Consider only opening the contact page after a successful copy, or at least warning the user that the details weren't copied before redirecting.

### Nitpicks (Optional)

- **[ConversationDetail.jsx:262]** The `Link` component from `react-router-dom` is used with `sx={{ color: 'inherit', textDecoration: 'underline' }}`. This is fine and consistent with MUI's `sx` prop pattern. No action needed.

- **[GCReducer.js]** The `combineReducers` entries are in correct alphabetical order. The new `messaging` entry is properly placed between `menuNotifications` and `moveEntranceToCave`. Clean.

- **[Account/index.jsx]** The `PreferencesSection` fetches notification preferences directly via `fetch` + `checkAuthStatus` rather than through a Redux action/reducer. This is a pragmatic choice that avoids the dead-code issue from prior reviews, but it does mean the notification preferences aren't in the Redux store (they're in local component state). This is acceptable for now since no other component needs this data, but worth noting if preferences need to be shared across components in the future.

ClemRz (approved):
## Code Review — PR #1328: Private Messaging UI (Sixth Pass)

The two commits since the fifth-pass approval cleanly address the remaining suggestions:
- Unused translation keys (`Convert`, `Converter`, `Coords converter`, `Not implemented yet`, `Section under development`) have been removed from all 15 language files.
- `handleConfirmReport` now only opens the contact page after a successful clipboard copy (addressing the clipboard failure UX concern).
- A documentation comment in `Person.jsx` explains the reliance on the undocumented `type` field from the API.
- French translations added for previously untranslated messaging keys (`Failed to copy...`, `Failed to load notification preferences`, `Notification Preferences`).

The merge from `develop` brings in unrelated features (MFA, polygon validation, shapefile import improvements) which are not part of this PR's scope — they were merged upstream and pulled in via the merge commit. No issues introduced by that merge in the messaging code.

- **[ConversationDetail.jsx:343–344 + index.jsx:226–227]** `timeZone="UTC"` and `timeZoneName="short"` are still present on `FormattedDate` components. This forces all message timestamps to display in UTC, which is inconsistent with the rest of the application (which uses the browser's local timezone). This was flagged in the fifth review and by Paul-AUB in the PR comments. Consider removing these props to let dates render in the user's local timezone — or at minimum, document this as a deliberate choice with a code comment explaining why UTC was chosen.

### Nitpicks (Optional)

- **[ConversationDetail.jsx:384]** The inline `style={{ color: replyText.length > 5000 ? 'red' : 'inherit' }}` could use MUI's `sx` prop or theme color (`error.main`) for consistency with the rest of the file. Same pattern in `ComposeDialog.jsx:224`.

- **[Person.jsx:114–116]** The comment block is helpful. Minor wording suggestion: "even though the field is not documented in the Swagger specification" could be shortened to "undocumented in the API spec" since the comment is already 3 lines.

urien (approved):
Cela me semble très bien. Si cela te convient @dawoldo tu peux résoudre les conflits et publier cette importante nouvelle fonctionnalité

PR #1330, opened 2026-05-20
fix(AuthorLink): prevent crash when verb is empty string

Paul-AUB:
# fix(AuthorLink): prevent crash when verb is empty string

## 🤔 What

- Fix a crash on history/snapshot pages (e.g. `/ui/entrances/:id/snapshots`)

## 🤷‍♂️ Why

`Contribution.jsx` passes `verb=""` (empty string) to `AuthorLink` when a reviewer exists but no author. Inside `AuthorLink`, `verbMessages[""]` is `undefined`, so the fallback `{ id: "" }` was passed to `formatMessage`. `@formatjs/intl` treats an empty string as a missing id and throws, crashing the whole page.

## 🔍 How

Guard `verb` before calling `formatMessage` in `AuthorLink`:

```js
const verbLabel = verb
  ? formatMessage(verbMessages[verb] ?? { id: verb })
  : '';
```

The existing `{verb && (...)}` check already suppresses the label display when `verb` is falsy, so behaviour is unchanged for all other cases.

## 🔗 Related API PR

N/A

## 🧪 Testing

Navigate to `/ui/entrances/156670/snapshots` — page should load without error.

## 📸 Previews

N/A

Closes #1329

### DIFF
--- a/packages/web-app/src/components/common/AuthorLink/index.js
+++ b/packages/web-app/src/components/common/AuthorLink/index.js
@@ -14,7 +14,9 @@ const verbMessages = defineMessages({
 
 const AuthorLink = ({ author, verb = 'Posted' }) => {
   const { formatMessage } = useIntl();
-  const verbLabel = formatMessage(verbMessages[verb] ?? { id: verb });
+  const verbLabel = verb
+    ? formatMessage(verbMessages[verb] ?? { id: verb })
+    : '';
   if (!author?.id || !author?.nickname)
     return <span>{formatMessage({ id: 'author.unknown' }, { verb: verbLabel })}</span>;
 


ClemRz (approved):

- **[packages/web-app/src/components/common/AuthorLink/index.js:21]** When `verb` is falsy and `author` is also unknown, the component renders `formatMessage({ id: 'author.unknown' }, { verb: '' })` which produces `" by an unknown author"` (empty verb interpolated). This is a minor UX edge case — consider whether a separate message without the verb placeholder (e.g., `"by an unknown author"`) would read better, or whether this scenario even occurs in practice. Not blocking since the crash is fixed and the output is still intelligible.

- **[packages/web-app/src/components/common/Contribution/Contribution.jsx:63]** The root cause is `verb={author ? 'Updated' : ''}` — passing an empty string rather than `undefined`. An alternative (or complementary) fix would be to pass `undefined` here instead of `''`, which would let the default parameter (`verb = 'Posted'`) kick in. That said, the defensive guard in `AuthorLink` is the right place to prevent this class of crash regardless of callers, so the current approach is fine.

### Nitpicks (Optional)

- **[Branch name]** The branch is named `feature/fix-history` — per project conventions, a bugfix branch would typically be `fix/history-crash` or `fix/author-link-empty-verb`. Minor naming inconsistency, not blocking.

PR #1331, opened 2026-05-20
feat(home): redesign homepage and FAQ page

Paul-AUB:
## 🤔 What

Complete redesign of the homepage and FAQ page. Closes #1312.

- **Header**: hero with CTA button (Explore the map), supporter logos always visible on mobile
- **HeroStats**: new strip of 4 key stats (entrances, documents, countries, cavers) below the header
- **Welcome**: simplified 3-column section (Free access / Better together / BBS-SA) with signup CTA
- **Footer**: full redesign — 4-column layout (Brand | Navigation | Follow us | Technical), CC-BY-SA and ODbL licence bar; removed Publisher.jsx, FooterLinks.jsx, FooterDisclaimer.jsx, SocialLinks.jsx, DonateForm.jsx
- **RandomEntry**: compact inline card layout
- **RecentChanges / Association / PartnersSection / LatestBlogNewsSection**: semantic HTML (`<section>`, `aria-labelledby`, `component="h2"`), lazy-loaded below the fold
- **FAQ**: complete rewrite — MUI Accordion with keyboard navigation, `<main>` semantic root, h1/h2 heading hierarchy, `alpha()` for colours, first item open by default, all deprecated palette tokens removed
- **i18n**: new keys propagated to all 15 languages (FR fully translated, 13 others translated)

## 🤷‍♂️ Why

The homepage lacked a clear value proposition above the fold, key stats were buried after several scrolls, the footer was outdated and missing licence information, and the FAQ used deprecated MUI patterns with broken links and an invisible anchor.

## 🔍 How

- Styled components in object notation throughout (no template literals)
- `formatMessage` only — no legacy `<Translate>` component
- `React.lazy` + `Suspense` for below-fold sections
- `alpha()` from `@mui/material/styles` for transparent colours
- `& a, & a:visited` CSS selector in NavList to force link colour regardless of component internals

## 🔗 Related API PR

None.

## 🧪 Testing

- Visit `/ui/` — verify hero, stats strip, welcome section, lazy-loaded sections, footer columns and licence bar
- Visit `/ui/faq` — verify accordion open/close, external link to protection doc, contributors page link
- Mobile: check buttons stack vertically, stats display 2×2, FSE/UIS logos visible
- Switch language to FR and verify translations display correctly

## 📸 Previews

image

image
image
image
image
image
image


ClemRz (changes requested):
Great work on this redesign — the homepage feels much more purposeful now with the hero CTA, stats strip, and cleaner footer. The FAQ rewrite with MUI Accordion is a solid improvement. A few things to address:

### Issues (Must Fix)

- **[RecentChangesCard.jsx:49]** `entities.pop()` mutates the array passed in from `changeInfo.subEntityTypes`, which comes from the Redux store. This violates Redux immutability and can cause subtle bugs (stale renders, missing items on re-render). Use a non-mutating approach:
  ```suggestion
  const lastEntity = entityFmt(entities[entities.length - 1], formatMessage);
  const otherEntities = entities
    .slice(0, -1)
    .map(e => entityFmt(e, formatMessage))
    .join(' ');
  ```

- **[FeedbackButton/index.jsx:8]** `CONTACT_URL` is hardcoded to `https://fr.wikicaves.org/contact` — non-French users will land on a French page. The project already has `contactLinks` in `conf/externalLinks.js` with locale-aware URLs (`fr` and `*`). Consider using `InternationalizedLink` or selecting the URL based on the current locale, similar to how the Footer does it.

- **[Header.jsx:79-81]** `dispatch(loadDynamicNumber('entrances'))` is called in `Header` but the result is never consumed (no `useSelector`). `HeroStats` already dispatches the same action and displays the value. This is dead code — consider removing the `useEffect`, the `useDispatch` hook, and the `loadDynamicNumber` import from `Header.jsx`.

- **[Translation files]** The keys `"Refresh"`, `"Feedback"`, and `"Find a cave among {count} entrances"` are added to all 15 language files but are not used anywhere in the codebase. Unused keys add maintenance burden. Either remove them or, if they're planned for a follow-up, note that in the PR description.

- **[PartnersSection.jsx]** This component still uses `LandingSection`, `GridRow`, and `GridFullColumn` from the deprecated `helpers/GridSystem`. The other sections in this PR use MUI `Grid` and `styled('section')` consistently. Consider aligning `PartnersSection` with the same pattern for consistency within this PR.

- **[LatestBlogNewsSection.jsx]** The blog news section now correctly filters by locale (FR vs EN) — nice fix for issue #10 from the original ticket. However, the `FRENCH_BLOG_URL` and `ENGLISH_BLOG_URL` constants are defined locally. Consider moving them to `conf/externalLinks.js` alongside the existing `frenchRssUrl`/`englishRssUrl` for discoverability.

### Nitpicks (Optional)

- **[Header.jsx:30]** The `Sitename` styled `h1` uses a fixed `fontSize: 55` (px). On very narrow viewports (<360px), this might overflow. The `HeaderRoot` has responsive `backgroundSize` breakpoints but the title doesn't scale down. A `clamp()` or a breakpoint-based font size could help.

- **[RandomEntryCard.jsx:155]** The condition `cave && (cave.depth || cave.length)` will skip rendering when `depth` or `length` is `0`. If `0` is a valid value (e.g., a cave at surface level), consider using `!isNil(cave.depth)` instead.

- **[Faq.jsx:95]** `<FaqRoot aria-label={...}>` — since `FaqRoot` is a `<main>` element and the page already has an `<h1>` inside it, the `aria-label` is redundant (screen readers will announce the heading). Not harmful, just unnecessary.

- **[Association.jsx]** The `GOALS` array has 5 items rendered in a 3-column grid (`md: 4`), which means the last row has 2 items and 1 empty slot. This is fine visually but worth noting if a 6th goal is ever added — it would fill perfectly.

urien (approved):
Pour moi c'est parfait.

ClemRz (changes requested):
Great work addressing the feedback from the previous review — the `entities.pop()` mutation is fixed, the hardcoded French URL in FeedbackButton now uses locale-aware `contactLinks`, the dead `loadDynamicNumber` dispatch in Header is removed, the deprecated GridSystem helpers are gone from PartnersSection, and the unused translation keys (`Refresh`, `Feedback`, `Find a cave among {count} entrances`) have been cleaned up. The overall redesign looks solid.

### Issues (Must Fix)

- **[FeedbackButton/index.jsx:32]** `theme.spacing(1.5)` is called in the mobile breakpoint padding. This project uses array-based spacing (`[2, 4, 8, 16, 24, 32, 64]`), so `theme.spacing(1.5)` resolves to `undefined` (array index 1.5 doesn't exist). Use a raw pixel value instead:
  ```suggestion
  padding: '6px 8px'
  ```
  Or use integer indices: `${theme.spacing(1)} ${theme.spacing(2)}` (which gives `4px 8px`).

- **[Translation files]** The key `"View the protection procedure document"` is added to all 15 language files but is not used anywhere in the codebase (the FAQ protection accordion no longer links to a document). Consider removing it to avoid dead translation keys.

- **[RandomEntryCard.jsx:186]** The condition `cave && (cave.depth || cave.length)` will evaluate to `false` when `depth` or `length` is `0`. If `0` is a valid value (e.g., a horizontal cave with 0 depth), this section won't render. Consider using `!isNil(cave.depth) || !isNil(cave.length)` — `isNil` from Ramda is already imported in this file.

### Nitpicks (Optional)

- **[homepage/index.jsx:23]** The `HomepageWrapper` hardcodes `fontFamily: 'Roboto, Helvetica, Arial, sans-serif'`. The theme already defines a comprehensive font stack in `grottoTheme.js`. Consider using `theme.typography.fontFamily` instead to stay in sync if the theme ever changes.

- **[Association.jsx]** The `GOALS` array has 5 items rendered in a 3-column grid (`md: 4`), which means the last row has 2 items and 1 empty slot. This is fine visually but worth noting — if a 6th goal is ever added it would fill perfectly, or the grid could use `md: 2.4` (not possible with MUI Grid) so consider whether 4 or 6 items would be a better fit.

ClemRz (approved):

PR #1334, opened 2026-05-21
fix(document): do not reset form fields when changing document type

Paul-AUB:
## 🤔 What

- Preserve all form field values when the user changes the document type in the create/edit document form
- Add a payload filter at submit time so that only fields relevant to the selected type are sent to the API

## 🤷‍♂️ Why

Previously, switching the document type called `resetContext(...)` in `DocumentTypeSelect`, which wiped the entire form state back to defaults — keeping only `title`, `mainLanguage`, `mainLanguageName`, and `authors`. This meant:

- A user who had filled in a description, publication date, authors, identifier, etc. would lose all that work by accidentally clicking a different type.
- Switching back to the original type did not restore the values; they were gone.
- There was also no symmetrical benefit: fields from a previous type were not filtered out of the API payload either, so irrelevant data could be submitted silently.

## 🔍 How

**1. Stop resetting on type change (`DocumentTypeSelect.jsx`)**

Replaced the `resetContext({ type, title, mainLanguage, ... })` call in `handleSelect` with a single `updateAttribute('type', newDocType)`. `updateAttribute` was already exposed by `DocumentFormContext`, so no context shape change was needed. Field visibility in `FormContent.jsx` is already conditional on the type, so hidden fields are never rendered — the data just stays alive in state until the user switches back.

**2. Filter the API payload at submit time (`documentTypeHelpers.js`)**

Added `filterDocumentPayload(docAttributes)` — a pure function that, given the current document state, returns only the fields that are actually visible/relevant for the current type:

- **Event** — minimal set: `title`, `description`, `datePublication`, `iso3166`, `authors` (no files, no language, no advanced metadata)
- **Authorization To Publish** — intermediate set: adds language and files (without license), but no advanced metadata or ISO regions
- **Everything else** (simple media + all standard types) — full set; sub-type-specific fields like `parent`, `pages`, `library`, `issue` are included but will be `null` for types that don't use them, so `buildFormData` already skips them

The field groups mirror exactly the conditional rendering in `FormContent.jsx`, including the global `AuthorsSection` (rendered for all non-unknown types) and the Advanced Metadata accordion (rendered for all types except Event and Authorization).

**3. Apply the filter before building FormData**

`filterDocumentPayload` is called at the top of `postDocument`, `updateDocument`, and `updateDocumentWithNewEntities` before the `FormData` / JSON body is assembled.

## 🔗 Related API PR

None.

## 🧪 Testing

1. Open the document creation form.
2. Select **Article**, fill in title, description, authors, publication date, identifier, and a parent document.
3. Switch to **Image** — all field values must be preserved (visible or not).
4. Switch back to **Article** — all previously entered values must reappear.
5. Submit as **Image** and inspect the network request — the POST body must not contain `parent`, `pages`, `library`, `issue`.
6. Switch to **Event**, submit — the POST body must not contain `files`, `mainLanguage`, `license`, `identifier`, `editor`, `subjects`, `creatorComment`.
7. Repeat steps 1–6 in edit mode (PUT) for an existing document.

## 📸 Previews

_N/A — no visual change; the fix affects state management and API payload construction._


urien:
Quand je rajoute un fichier et que je sélectionne "il existe une autorisation" cela fige le formulaire, je ne peux pas sélectionner d'autorisation ni changer de règle concernant les droits
image

PR #1335, opened 2026-05-21
feat(entrances): improve UX and access control for sensitive entrances

### LINKED ISSUE #1340: Non-admin users cannot edit any property of sensitive entrances
**Describe the bug**

Non-admin users cannot update any property (name, hazards, altitude, etc.) of a sensitive entrance (`isSensitive: true`) via the edit form. The form's submit button remains permanently disabled, so no PUT request is ever sent to the API.

The root cause is in the edit form logic:

1. For sensitive entrances, the API does not return coordinates to non-admin users (by design — to protect the physical location).
2. The `EntranceDetail` component hides the `CoordinateFormSection` when `!permissions.isAdmin && initialIsSensitive` (correct behavior — non-admins shouldn't see/edit coordinates of sensitive entrances).
3. However, the parent `EntranceForm` component disables the submit button whenever `latitude` or `longitude` is empty:
   ```js
   const isSubmitDisabled =
     isCoordEmpty(lat) ||
     isCoordEmpty(lng) || ...
   ```
4. Since coordinates are null/empty and the coordinate fields are hidden, the user can never satisfy this validation — the form is permanently locked.

**To Reproduce**

1. Log in as a non-admin user (any user without the Administrator role)
2. Navigate to a sensitive entrance (e.g., `/ui/entrances/138163` or `/ui/entrances/156694`)
3. Click "Edit properties"
4. Try to change the entrance name or any other field
5. Observe that the submit button is disabled and cannot be clicked

**Expected behavior**

Non-admin users should be able to edit non-geographic properties (name, hazards, altitude, discovery year, touristic flag, etc.) of sensitive entrances. Only the coordinate fields should be restricted.

**Affected files**

- `packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx` — `isSubmitDisabled` logic
- `packages/web-app/src/components/appli/EntitiesForm/Entrance/EntranceDetail.jsx` — hides coordinate section

**Suggested fix**

The `isSubmitDisabled` check should skip the coordinate validation when the entrance is sensitive and the user is non-admin:

```js
const isSensitiveAndNonAdmin = entranceValues?.isSensitive && !permissions.isAdmin;

const isSubmitDisabled =
  (!isSensitiveAndNonAdmin && (isCoordEmpty(lat) || isCoordEmpty(lng))) ||
  (entityType === ENTRANCE_AND_CAVE
    ? !caveName || !caveLanguage
    : !entranceName || !entranceLanguage);
```

Additionally, the `makeEntranceData` transformer should avoid sending null coordinates when the user doesn't have access to them, or the API should ignore null coordinate values on sensitive entrances from non-admin users.

**Additional context**

- An admin (with Administrator role) can successfully edit the same entrances because the API returns coordinates to admins and the coordinate section is visible.
- The `isSensitive` flag is designed to protect cave locations, not to prevent editing of non-geographic metadata.
- This affects any entrance where `isSensitive` is `true` and the user lacks the Administrator role.

Paul-AUB:
## 🤔 What

Improve the UX for sensitive entrances by properly restricting location-related actions for non-admin users.

- Replace the blank/missing map area with a dedicated `SensitiveLocationPlaceholder` component (lock icon + explanatory message) for non-admins viewing a sensitive entrance
- Hide the `SensitiveCaveWarning` banner from non-admin users (was leaking the sensitivity status without useful context)
- Prevent non-admins from adding new locations on a sensitive entrance
- Fix the entrance edit form: skip coordinate validation for non-admin users editing a sensitive entrance, so the form can be saved without location data

## 🤷‍♂️ Why

Sensitive entrances must not expose their precise location. Previously the map area was simply hidden, leaving a confusing gap in the UI. Non-admins could also attempt to add locations or save an edit form that incorrectly required coordinates.

closes #1340

## 🔍 How

- New `SensitiveLocationPlaceholder` component rendered in place of the map when `entrance.isSensitive && !isAdmin`
- `Entry/index.jsx`: conditionalize `SensitiveCaveWarning` on `isAdmin`, swap conditional rendering to always show the `Box` (flex container) and pick map vs placeholder inside
- `Entry/Locations/index.jsx`: gate the "Add location" button on `!isSensitive || permissions.isAdmin`
- `EntitiesForm/Entrance/index.jsx`: derive `isSensitiveDisabled` from `isAdmin` + `entranceValues.isSensitive`, then only require coordinates when `!isSensitiveDisabled`
- i18n: new keys added to all 15 language files (`Sensitive location` + description)

## 🔗 Related API PR

_None — behaviour relies on existing `isSensitive` flag already returned by the API._

## 🧪 Testing

1. Log in as a **regular user** and navigate to a sensitive entrance:
   - Map area should show the lock placeholder instead of the map
   - The "Add location" button should not appear
   - Editing the entrance should not require coordinates to save
2. Log in as an **admin** and navigate to the same entrance:
   - Map and `SensitiveCaveWarning` should still appear normally
   - "Add location" button should be available
   - Coordinate validation should still apply in the edit form

## 📸 Previews

image

image

### DIFF
--- a/packages/web-app/public/lang/ar.json
+++ b/packages/web-app/public/lang/ar.json
@@ -1742,6 +1742,7 @@
   "Semi-minor radius: ": "نصف القطر شبه الثانوي: ",
   "Send reset email": "إرسال بريد إعادة التعيين",
   "Senegal": "السنغال",
+  "Sensitive location": "Sensitive location",
   "Sensitive massif": "كتلة جبلية حساسة",
   "Sensitivity Management": "إدارة الحساسية",
   "Serbia": "صربيا",
--- a/packages/web-app/public/lang/ar.json
+++ b/packages/web-app/public/lang/ar.json
@@ -1742,7 +1742,7 @@
   "Semi-minor radius: ": "نصف القطر شبه الثانوي: ",
   "Send reset email": "إرسال بريد إعادة التعيين",
   "Senegal": "السنغال",
-  "Sensitive location": "Sensitive location",
+  "Sensitive location": "موقع حساس",
   "Sensitive massif": "كتلة جبلية حساسة",
   "Sensitivity Management": "إدارة الحساسية",
   "Serbia": "صربيا",
--- a/packages/web-app/public/lang/bg.json
+++ b/packages/web-app/public/lang/bg.json
@@ -1742,6 +1742,7 @@
   "Semi-minor radius: ": "Малка полуос",
   "Send reset email": "Изпратете имейл за нулиране на паролата.",
   "Senegal": "Сенегал",
+  "Sensitive location": "Sensitive location",
   "Sensitive massif": "Чувствителен масив",
   "Sensitivity Management": "Управление на чувствителността",
   "Serbia": "Сърбия",
--- a/packages/web-app/public/lang/bg.json
+++ b/packages/web-app/public/lang/bg.json
@@ -1742,7 +1742,7 @@
   "Semi-minor radius: ": "Малка полуос",
   "Send reset email": "Изпратете имейл за нулиране на паролата.",
   "Senegal": "Сенегал",
-  "Sensitive location": "Sensitive location",
+  "Sensitive location": "Чувствително местоположение",
   "Sensitive massif": "Чувствителен масив",
   "Sensitivity Management": "Управление на чувствителността",
   "Serbia": "Сърбия",
--- a/packages/web-app/public/lang/ca.json
+++ b/packages/web-app/public/lang/ca.json
@@ -1742,6 +1742,7 @@
   "Semi-minor radius: ": "Semieix menor :",
   "Send reset email": "Envia correu de restabliment",
   "Senegal": "Senegal",
+  "Sensitive location": "Sensitive location",
   "Sensitive massif": "Massís sensible",
   "Sensitivity Management": "Gestió de la sensibilitat",
   "Serbia": "Sèrbia",
--- a/packages/web-app/public/lang/ca.json
+++ b/packages/web-app/public/lang/ca.json
@@ -1742,7 +1742,7 @@
   "Semi-minor radius: ": "Semieix menor :",
   "Send reset email": "Envia correu de restabliment",
   "Senegal": "Senegal",
-  "Sensitive location": "Sensitive location",
+  "Sensitive location": "Localització sensible",
   "Sensitive massif": "Massís sensible",
   "Sensitivity Management": "Gestió de la sensibilitat",
   "Serbia": "Sèrbia",
--- a/packages/web-app/public/lang/de.json
+++ b/packages/web-app/public/lang/de.json
@@ -1742,7 +1742,7 @@
   "Semi-minor radius: ": "Semi-Moll-Radius:",
   "Send reset email": "Zurücksetz-E-Mail senden",
   "Senegal": "Senegal",
-  "Sensitive location": "Sensitive location",
+  "Sensitive location": "Sensibler Standort",
   "Sensitive massif": "Sensibles Massiv",
   "Sensitivity Management": "Sensibilitätsmanagement",
   "Serbia": "Bosnisch-Serbische Republik",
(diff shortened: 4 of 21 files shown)


urien (approved):
Tout fonctionne

ClemRz (approved):

- **[packages/web-app/src/components/appli/Entry/SensitiveLocationPlaceholder.jsx:11]** The `bgcolor: 'secondary.veryLight'` palette token — is this defined in the project's MUI theme? If not, it will silently resolve to nothing and the background will be transparent. Consider verifying it exists in the theme config, or fall back to a standard token like `'action.hover'` or a manual alpha value (e.g., `'rgba(0,0,0,0.04)'`).

- **[packages/web-app/src/components/appli/Entry/Locations/index.jsx:49]** The `(!isSensitive || permissions.isAdmin)` condition is correct, but it's worth noting that when `isSensitive` is true and the user is not admin, the `icon` prop evaluates to `false` (since `&&` short-circuits). This works fine with MUI's `ScrollableContent`, but a brief inline comment explaining the intent (e.g., `// Hide "Add location" for non-admins on sensitive entrances`) would help future readers quickly understand the three-part condition.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/Entry/Locations/index.jsx:22-24]** The reformatting of `useMoveRelevanceWithUndo(moveLocationRelevance)` into a multi-line call is a Prettier auto-format correction (line length), but it's unrelated to the PR's purpose. Not blocking, just noting for clean git history.

- **[packages/web-app/public/lang/ar.json, bg.json, ca.json, de.json, el.json, es.json, he.json, id.json, it.json, ja.json, nl.json, pt.json, ro.json]** The `"Sensitive location"` key is left untranslated (English fallback) in 13 out of 15 language files. Only `fr.json` has a proper translation. This is acceptable as a placeholder (translations are AI-generated per the project conventions), but consider running the translation generation to fill these in before merge.

- **[scripts/translations/sort.js:17]** The regex change from `/^(\s+)"/m` to `/^([^\S\n]+)"/m` is a good fix — it prevents matching a newline as part of the indent. Consider adding a brief comment explaining why `[^\S\n]` is used instead of `\s` (i.e., "match whitespace except newline to avoid capturing line breaks as indentation").

ClemRz (changes requested):
Please make sure you integrate the changes on the transformers from PR #1341.

ClemRz (approved):

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/transformers.js:34-40]** The coordinate guard uses loose equality (`!= null`) combined with an empty-string check. This is fine, but consider also handling the case where coordinates are `0` (valid value — equator/prime meridian). Currently `0` would pass the check correctly since `0 != null` is `true` and `0 !== ''` is `true`, so this is safe. Just flagging for awareness since the condition is security-sensitive (protecting stored coordinates from being wiped).

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:86]** The `isSensitiveDisabled` variable is computed from `entranceValues?.isSensitive`, which reflects the *initial* prop value. In `EntranceDetail.jsx`, a similar variable uses `useRef` to capture the initial value explicitly. Here, since `entranceValues` is a prop that doesn't change during the form's lifecycle, both approaches are equivalent — but it's worth noting the subtle difference in strategy between the two files. If `entranceValues` were ever to be updated reactively (e.g., via a refetch), the behavior here would differ from `EntranceDetail`. A brief comment like `// Based on initial prop — entranceValues doesn't change during form lifecycle` would clarify the intent.

- **[packages/web-app/src/components/appli/Entry/Locations/index.jsx:49-51]** The `(!isSensitive || permissions.isAdmin)` condition is correct and well-commented. One edge case to consider: if a moderator (not admin) should eventually be allowed to add locations on sensitive entrances, this would need updating. The current implementation is admin-only, which matches the issue requirements — just noting for future extensibility.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/Entry/Locations/index.jsx:25-27]** The reformatting of `useMoveRelevanceWithUndo(moveLocationRelevance)` into a multi-line call is a Prettier auto-format correction (line length). Not blocking, just noting it's unrelated to the PR's purpose.

- **[scripts/translations/sort.js:17]** The regex change from `/^(\s+)"/m` to `/^([^\S\n]+)"/m` is a good fix — it prevents matching a newline as part of the indent. Consider adding a brief inline comment explaining why `[^\S\n]` is used instead of `\s` (i.e., "match horizontal whitespace only, excluding newlines").

PR #1337, opened 2026-05-22
fix(entrances): error management to move sections by relevance

Paul-AUB:
## 🤔 What

- Fix error propagation in `createMoveRelevanceAction` so the `.catch` handler returns `{ error: true }` on auth errors and `{ error: error.message }` on other failures
- In `useMoveRelevanceWithUndo`, display the error message via a snackbar when moving a section's relevance fails (instead of silently swallowing the error)

## 🤷‍♂️ Why

When moving an entrance section by relevance failed (e.g. due to a network or API error), the user received no feedback — the error was caught but never surfaced in the UI. The fix ensures the error message is shown to the user via a snackbar notification.

## 🔍 How

- `createMoveRelevanceAction.js`: the `.catch` block now returns `{ error: true }` for auth errors (to keep the early-return guard working) and `{ error: error.message }` for other errors so the hook can read the message.
- `useMoveRelevanceWithUndo.js`: when `result?.error` is a string, it is displayed as an error snackbar with a 6-second duration before the function returns early.

## 🔗 Related API PR

See also https://github.com/GrottoCenter/grottocenter-api/pull/1605

## 🧪 Testing

1. Trigger a relevance move on an entrance section while the API returns an error (e.g. temporarily break the network or mock a 500 response).
2. Verify that an error snackbar appears with the error message.
3. Verify that the undo snackbar does **not** appear on failure.
4. Verify that a successful move still shows the "Order updated" success snackbar with the Undo action.

## 📸 Previews

image

### DIFF
--- a/packages/web-app/src/actions/createMoveRelevanceAction.js
+++ b/packages/web-app/src/actions/createMoveRelevanceAction.js
@@ -29,18 +29,15 @@ const createMoveRelevanceAction = (entityName, urlBuilder, label) => {
       .then(checkAuthStatus(dispatch))
       .then(response => response.json())
       .then(data =>
-        dispatch({
-          type: SUCCESS,
-          moved: data.moved,
-          swapped: data.swapped
-        })
+        dispatch({ type: SUCCESS, moved: data.moved, swapped: data.swapped })
       )
       .catch(error => {
-        if (error.isAuthError) return;
+        if (error.isAuthError) return { error: true };
         dispatch({
           type: FAILURE,
           error: makeErrorMessage(error.message, `Moving ${label} relevance`)
         });
+        return { error: error.message };
       });
   };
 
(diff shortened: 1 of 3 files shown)


ClemRz (approved):

- **[hooks/useMoveRelevanceWithUndo.js:18-25]** The error message displayed in the snackbar is the raw `error.message` from the caught exception. Depending on what the API returns, this could be a technical message (e.g., "Failed to fetch" or a stack trace) rather than a user-friendly one. Consider wrapping it with a generic fallback or a translated message, e.g.:
  ```javascript
  enqueueSnackbar(
    result.error || formatMessage({ id: 'An error occurred' }),
    { variant: 'error', autoHideDuration: 6000 }
  );
  ```
  That said, if the API already returns clean messages, this is fine as-is.

- **[hooks/useMoveRelevanceWithUndo.js:40-42]** The undo path silently swallows errors (`if (undoResult?.error) return;`). Now that the initial move surfaces errors, it might be worth showing a snackbar on undo failure too — the user would otherwise think the undo succeeded. This is pre-existing behavior and outside the scope of this PR, so just flagging for a potential follow-up.

### Nitpicks (Optional)

- **[hooks/useMoveRelevanceWithUndo.js:50]** The `}}>` formatting change (closing JSX bracket on the same line) is correct per the project's `jsxBracketSameLine: true` Prettier config. It's unrelated to the bug fix but harmless — just noting it for commit history clarity.

PR #1339, opened 2026-05-23
feat(Massif): add client-side polygon validation

ClemRz:
# feat(Massif): add client-side polygon validation

Closes #1338

## 🤔 What

Adds client-side polygon validation to the massif editor using Turf.js. Validation runs on polygon state changes, displays errors in the PolygonLayersList panel, blocks form submission on blocking errors, and handles API errors inline.

- Automatic RFC 7946 winding order normalization (CCW exterior, CW holes)
- Self-intersection detection via `@turf/kinks`
- Minimum point count enforcement (≥ 3 unique vertices)
- Total area limit enforcement (35,000 km²)
- Per-polygon error icons with tooltips in the layers list
- Global error alert banners by category
- Submit button disabled when blocking errors exist
- API 4xx errors displayed inline without losing form state
- All validation messages translated into 15 languages

## 🤷‍♂️ Why

Previously, invalid polygon geometry (self-intersections, too few points, excessive area) was only caught by the API after submission, resulting in a generic error screen that lost the user's map context. Users had to start over or guess what went wrong.

This change provides immediate, actionable feedback as users draw/edit polygons, preventing invalid submissions and preserving the editing context when API errors do occur.

## 🔍 How

**New module:** `src/helpers/polygonValidation.js` — pure utility functions using individual Turf.js packages (`@turf/rewind`, `@turf/kinks`, `@turf/area`, `@turf/helpers`). Tree-shakeable, no full Turf bundle.

**PolygonMap integration:** Every polygon state change (create, edit, import, hole toggle, initial load) runs through `normalizeWinding()` then `validatePolygon()`. Results stored as `hasSelfIntersection` and `tooFewPoints` flags on each layer object. Derived state computes `totalAreaKm2`, `areaExceeded`, and `hasBlockingErrors`.

**PolygonLayersList:** Renders global `<Alert severity="error">` banners for each active error category, plus per-item `ErrorOutlineIcon` with tooltip. Existing needle warning behavior unchanged.

**MassifForm:** `onValidationChange` callback propagates blocking error state up from PolygonMap → MassifFields → MassifForm. Submit button disabled and `handleFormSubmit` guards against submission. API 4xx errors extracted and shown as dismissible inline Alert above the form fields.

## 🔗 Related API PR

N/A — this is purely client-side validation. The API already returns POLYGON_* error codes; this PR adds client-side pre-validation and better error display.

## 🧪 Testing

- Unit tests for `polygonValidation.js` (normalizeWinding, checkSelfIntersections, hasMinimumPoints, computeTotalArea, validatePolygon)
- Manual testing: draw a self-intersecting (bowtie) polygon → error icon + alert appear, submit disabled
- Manual testing: draw a polygon with only 2 points → too-few-points error
- Manual testing: draw a very large polygon exceeding 35,000 km² → area exceeded alert
- Manual testing: fix the polygon → errors clear, submit re-enabled
- Manual testing: trigger a 4xx API error → inline alert shown, form stays open

## 📸 Previews

Screenshot 2026-05-23 at 12 12 12 p m

[Martinique.geojson.zip](https://github.com/user-attachments/files/28181133/Martinique.geojson.zip)


urien (approved):
Cela fonctionne parfaitement, Merci @ClemRz 
Il y a une difficulté qui apparait lorsqu'on veut modifier les périmètres pour éviter les chevauchemements : les points sont représentés par de gros carrés et il a  très rapidement de grosses densités de carrés qui n'apparaissent pas comme lié à une ligne ou a une autre. Les carrés seraient plus fonctionnels s'ils étaient un peu plus petits et la ligne mieux visible sous ces carrés.

urien (changes requested):
J'ai mis en commentaire une proposition permettant de faciliter le travail des utilisateurs. Dis moi si c'est possible

Paul-AUB (changes requested):
### Issues (Must Fix)

1. **[`src/helpers/polygonValidation.js`]** `src/helpers/` is explicitly marked deprecated in AGENTS.md ("⚠️ DEPRECATED utilities — prefer hooks"). Pure utility modules (constants, helper functions) belong in `src/utils/`. This file should be moved to `src/utils/polygonValidation.js` and all imports updated accordingly. The test file should move to `src/utils/polygonValidation.test.js` as well.

2. **[`PolygonMap.jsx:200-202`]** The "all holes" error rendered in `PolygonLayersList` is not propagated as a blocking error. When every layer has `isHole: true`, `PolygonLayersList` shows an `Alert severity="error"` banner, but `hasBlockingErrors` in `PolygonMap` only checks `hasSelfIntersection`, `tooFewPoints`, and `areaExceeded`. The submit button will remain enabled and `onValidationChange(false)` is passed up, leaving the user able to submit an invalid geometry (no outer polygon). Consider adding `mapLayers.length > 0 && mapLayers.every(l => l.isHole)` to the `hasBlockingErrors` expression.

3. **[`src/helpers/polygonValidation.js:67`]** `normalizeWinding` calls `rewind(geojson, { reverse: true })`. According to Turf.js docs, `reverse: true` enables *reverse* winding order (CW exterior, CCW holes) — the opposite of RFC 7946. The surrounding comment reads "because rewind() always makes the exterior CCW", which describes the *default* behavior (`reverse: false`). The existing tests only verify idempotency, not actual winding direction, so a backwards winding would pass them. Consider removing `{ reverse: true }` and relying on the default RFC 7946 behavior, then handling hole winding by reversing the CCW result for `isHole` cases. This aligns with the comment as written.

4. **[`PolygonMap.jsx:382`, `PolygonMap.jsx:416`]** Both `onEdited` and `onDeleted` use `Object.values(_layers).map(layer => { ... })` to produce side effects (`setMapLayers`). `.map()` is for transformations that return a new array; `.forEach()` is the correct method when the goal is iteration with side effects. The unused return values from `.map()` are a code smell.

5. **[`PolygonMap.jsx:767-775`]** `PolygonMap.propTypes` defines `data` with only a `coordinates` shape, but `data.type` is accessed at line 592 (`data.type === 'Polygon'`) and implicitly at line 219 (centroid calculation). Adding `type: PropTypes.string` to the `data` PropTypes shape would make the expected contract explicit and improve runtime validation.

### Nitpicks (Optional)

6. **[`index.jsx:111`]** The "Please finish editing the polygon before submitting." guard uses native `window.alert()`, which is blocking and visually inconsistent with the rest of the form (which uses inline Alerts and `useNotification`). Consider using `onError` from `useNotification` or an inline warning `Alert` instead.

ClemRz:
Thanks for the thorough review @Paul-AUB! All 6 items addressed:

**Must Fix**
1. ✅ Moved `polygonValidation.js` and its test to `src/utils/`, updated all imports.
2. ✅ Added `mapLayers.length > 0 && mapLayers.every(l => l.isHole)` to `hasBlockingErrors`.

**Should Consider**
3. ✅ Removed `{ reverse: true }` — the default behavior is RFC 7946 compliant (CCW exterior, CW holes). Updated the comment and strengthened the tests to verify actual winding direction, not just idempotency.
4. ✅ Changed `.map()` to `.forEach()` in both `onEdited` and `onDeleted`.
5. ✅ Added `type: PropTypes.string` to the `data` shape.

**Nitpicks**
6. ✅ Replaced `window.alert()` with `onWarning()` from `useNotification`.

Additionally:
- Added inter-polygon crossing detection using `@turf/line-intersect` (kink markers appear on the map where polygons cross each other, submission blocked).
- API 4xx errors now show as translated toast notifications instead of an inline alert (fixes a UX issue where the error was out of viewport).
- Form stays mounted during API requests so the map isn't reset on error.
- Added translated `POLYGON_*` error code keys to all 15 language files.

urien (dismissed):
J'ai fait le test avec le fichier joint. Toute la partie gestion du polygone se déroule très bien on arrive parfaitement a traiter les pb relevés. Par contre a l'enregistrement j'ai eu un message "erreur interne"

[LSG-Schmallenberg Sued-Ost _Typ A_.geojson.txt](https://github.com/user-attachments/files/28303220/LSG-Schmallenberg.Sued-Ost._Typ.A_.geojson.txt)

urien (approved):
cela fonctionne parfaitement du point de vue utilisateur

Paul-AUB (approved):
All six items from the previous review have been resolved. Reviewing the new additions:

**`polygonValidation.js` → `src/utils/`** ✓ Correctly relocated.

**`normalizeWinding`** ✓ `{ reverse: true }` removed; the function now correctly produces CCW exterior rings and CW holes per RFC 7946.

**`allHoles` blocking error** ✓ `mapLayers.every(l => l.isHole)` is now included in `hasBlockingErrors`.

**`.forEach()` in `onEdited`/`onDeleted`** ✓ Fixed.

**`data.type` in PropTypes** ✓ Added.

**`alert()` → `onWarning()`** ✓ Fixed.

**New additions look solid:**

- `checkInterPolygonIntersections` using `@turf/line-intersect` is well-implemented, correctly tested (including the hole case), and properly wired into `hasBlockingErrors` and the kink-marker overlay.
- `POLYGON_*` API error codes are translated in all 16 language files and surfaced via `formatMessage({ id: code, defaultMessage: ... })` — good approach.
- The `FormProgressInfo` condition now guards `!massifLoading && !nameLoading`, avoiding a false transition to the success/error screen while the API call is in flight.
- `isSubmitting || massifLoading || nameLoading` on the submit button is the right guard.

1. **[`index.jsx:167`]** `isLoading={false}` is passed to `FormProgressInfo` unconditionally. While consistent with the `!massifLoading` guard on the surrounding `if`, the `labelLoading` prop is still provided and will never render. Consider removing `labelLoading` to avoid dead code, or document why the loading indicator is intentionally suppressed here.

2. **[`utils/polygonValidation.js:169-196`]** `checkInterPolygonIntersections` compares every polygon pair including holes. A hole that is properly contained inside its outer polygon will not produce false positives (no edge crossings), so the current logic is correct. Worth adding a code comment clarifying this — future readers may wonder why holes are not filtered out.

PR #1343, opened 2026-05-26
feat(auth): implement frontend MFA support for administrator accounts

Paul-AUB:
# feat(auth): implement frontend MFA support for administrator accounts

## 🤔 What

Adds full frontend support for the MFA (TOTP-based two-factor authentication) system introduced in the API ([GrottoCenter/grottocenter-api#1608](https://github.com/GrottoCenter/grottocenter-api/issues/1608)).

- **Multi-step login flow** — after email/password, admins with MFA active are prompted for their 6-digit TOTP code (auto-submit on 6 digits, numeric keyboard, paste support)
- **MFA enrollment wizard** — admins without MFA are redirected to a 3-step wizard (install authenticator app → scan QR code → verify) rendered inside the login modal to preserve the 10-min enrollment token TTL
- **MFA reset** — new "Two-factor authentication" section in account settings (admins only) showing MFA status (active/inactive) with a password-confirmed disable flow
- **Admin session expiry banner** — non-blocking dismissible warning when an admin's session expires in less than 24 hours (due to the new 10-day TTL for admins vs 90 days for regular users)
- **i18n** — 25 new keys added to all 16 language files (EN translated, FR translated, others using EN fallback via `sync-with-en`)

## 🤷‍♂️ Why

The API now enforces mandatory TOTP MFA for administrator accounts, a reduced token TTL (10 days), and stricter login rate limiting. Without these frontend changes, admin users would receive unhandled API responses and be unable to log in after the API is deployed.

## 🔍 How

Three new API response statuses are handled in the login flow:

| API status | Meaning | Frontend behaviour |
|---|---|---|
| `MfaRequired` | Admin has MFA active | Show 6-digit TOTP input (step 2a) |
| `MfaEnrollmentRequired` | Admin has no MFA yet — receives a 10-min restricted token | Show 3-step enrollment wizard inside login modal (step 2b) |
| `InvalidTotpCode` / `TotpAlreadyUsed` | Bad or replayed code | Inline error, stay on TOTP step |

New Redux slices:
- `actions/Mfa.js` — `postMfaEnroll`, `postMfaVerify`, `postMfaLogin`, `postMfaReset`
- `reducers/MfaReducer.js` — tracks enroll/verify/reset loading + error states
- `reducers/LoginReducer.js` — extended with `isMfaRequiredDisplayed`, `isMfaEnrollmentRequiredDisplayed`, `enrollmentToken`

New components:
- `components/appli/MfaEnrollment/index.jsx` — 3-step MUI Stepper wizard with QR code (`qrcode.react`) and copyable secret
- `TotpStep` sub-component in `LoginForm` — auto-submits at 6 digits, `inputMode="numeric"`, `autoComplete="one-time-code"`

UX decisions:
- Enrollment stays inside the login modal (not a page redirect) to keep the 10-min enrollment token alive
- If the enrollment token expires mid-wizard, a warning with a "Back to login" CTA is shown
- Session expiry banner is rendered above the AppBar, dismissible for the current session only

## 🔗 Related API PR

Implements the frontend counterpart of [GrottoCenter/grottocenter-api#1608](https://github.com/GrottoCenter/grottocenter-api/issues/1608).

> ⚠️ The staging site will not be fully functional until the API PR is merged and deployed.

## 🧪 Testing

Requires the API branch `feat/admin-security-hardening` to be running locally.

1. **Login with MFA active** — log in as an admin with MFA enrolled → TOTP step appears → enter valid code → logged in
2. **Enrollment flow** — log in as an admin without MFA → 3-step wizard appears → scan QR → verify code → logged in with MFA active
3. **Invalid code** — enter wrong TOTP code → inline error, stays on step 2
4. **MFA reset** — go to account settings as admin → "Two-factor authentication" section shows "Active" badge → click "Disable" → enter password → success
5. **Non-admin unaffected** — log in as regular user → no MFA steps, no session banner
6. **Session expiry banner** — force `authTokenDecoded.exp` close to `Date.now() / 1000` in Redux DevTools → warning banner appears for admins only

## 📸 Previews

_Screenshots to be added._


ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/hooks/usePermissions.js:25]** `isTokenExpired` is called without the `authState` argument on line 25, while the function signature requires it (`const isTokenExpired = authState => { ... }`). This will always return `false` (the `catch` branch catches the TypeError from accessing `.exp` on `undefined`) — meaning `isTokenExpired` exposed from the hook is always `false`, which breaks any consumer relying on it.
  ```suggestion
    isTokenExpired: isTokenExpired(authState),
  ```

- **[packages/web-app/src/pages/ApplicationShell.jsx:127-138]** `AdminSessionExpiryBanner` computes `secondsUntilExpiry` during render but never re-renders when time passes. If a user loads the page when `secondsUntilExpiry >= SECONDS_IN_DAY`, the banner will never appear even after the threshold is crossed (unless something else triggers a re-render). Consider adding a timer (`setInterval` or `setTimeout`) that forces a re-check once the threshold is about to be crossed, or at a reasonable interval (e.g., every 5 minutes).

- **[packages/web-app/src/components/appli/Login.jsx:87]** `postMfaLogin` is called with `email` and `password` from component state. If the user clears or modifies the email/password fields after the initial login (the fields are still rendered in the `LoginForm` component via `onEmailChange`/`onPasswordChange` props even though the TOTP step is shown), the TOTP submission will send stale or empty credentials. Since the TOTP step replaces the form entirely (via `totpMode`), the credentials are safe in this implementation — but the `email` and `password` props are still passed to `LoginForm` and could be cleared by a future refactor. Consider storing the credentials used for the initial login attempt in a ref or in Redux state when `MfaRequired` is received, rather than relying on the component's mutable state remaining untouched.

- **[packages/web-app/src/components/common/LoginForm/index.jsx + MfaEnrollment/index.jsx]** The `normalizeOtp` function and the TOTP input logic (handleChange, handlePaste, errorMessage, the entire JSX block) are duplicated almost identically between `TotpStep` in `LoginForm/index.jsx` and `StepVerify` in `MfaEnrollment/index.jsx`. Consider extracting a shared `TotpInput` component (or at minimum the `normalizeOtp` utility) to reduce duplication and ensure consistent behavior if one is updated later.

- **[packages/web-app/src/actions/Mfa.js:120-121]** In `postMfaReset`, the action uses `checkAuthStatus`-style auth (spreading `authorizationHeader`) but does not call `checkAuthStatus(dispatch)` from `src/actions/utils.js` as documented in the project conventions. If the user's session token is expired/revoked, the 401 won't trigger the standard logout+reload flow. Consider using `checkAuthStatus` for consistency with other authenticated actions.

- **[packages/web-app/src/pages/Account/index.jsx:688-690]** The `useEffect` that triggers `postLogout()` on `mfaReset.isSuccess` has an eslint-disable for `react-hooks/exhaustive-deps`. This is acceptable here since you intentionally only want to react to `isSuccess` changes, but the logout happens immediately without giving the user time to read the success toast. The toast and the redirect to `/` (from `postLogout`) may race. Consider adding a short delay (e.g., 1-2 seconds) before logging out so the user sees the confirmation.

- **[packages/web-app/src/pages/ApplicationShell.jsx:131]** The `sessionStorage` key `'mfaExpiryBannerDismissed'` persists for the entire browser session. If an admin dismisses the banner, logs out, and another admin logs in during the same browser session, the second admin won't see the banner. Consider including the user ID in the storage key, or clearing it on logout.

- **[packages/web-app/src/components/appli/Login.jsx:284]** There's a commented-out line `// fullScreen={isMobile}` on the main login dialog. This looks like a leftover from development — either apply it consistently (it's used on the MFA dialogs above) or remove the comment.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/MfaEnrollment/index.jsx:38]** `StepInstall` uses `Alert` from `@mui/material` directly (children-based), while the Account page uses the project's custom `Alert` component (with `content` prop). Both work, but mixing the two patterns in the same feature could confuse future contributors. Not blocking, just worth being aware of.

- **[packages/web-app/src/actions/Mfa.js]** The `loginUrl` import is only used by `postMfaLogin`. Since `postMfaLogin` essentially re-does the login request with an added `totpCode`, consider adding a comment clarifying why it hits the same `/login` endpoint rather than a dedicated MFA endpoint — this helps future readers understand the API contract.

- **[packages/web-app/src/components/appli/MfaEnrollment/index.jsx:69]** `navigator.clipboard.writeText` can fail (e.g., in non-secure contexts or if the user denies permission). Consider wrapping it in a try/catch with a fallback notification.

ClemRz (approved):

- **[packages/web-app/src/components/appli/MfaEnrollment/index.jsx:38]** `StepInstall` imports `Alert` directly from `@mui/material` and uses it with children, while the Account page imports the project's custom `Alert` component (from `components/common/Alert`) which uses a `content` prop. Both work, but mixing the two patterns within the same feature makes it easy for a future contributor to use the wrong one. Consider using the project's custom `Alert` here for consistency, or at minimum add a comment explaining the choice.

- **[packages/web-app/src/components/appli/MfaEnrollment/index.jsx:72]** The clipboard failure callback `() => {}` silently swallows errors. Consider showing a fallback notification (e.g., "Copy failed — please select and copy manually") so the user knows the action didn't succeed in non-secure contexts or when permission is denied.

- **[packages/web-app/src/pages/Account/index.jsx:688]** The `eslint-disable` for `react-hooks/exhaustive-deps` is justified here (intentionally reacting only to `isSuccess` changes), but a brief inline comment explaining *why* the other deps are excluded would help future readers understand the intent without having to reason through it.

### Nitpicks (Optional)

- **[packages/web-app/src/components/common/LoginForm/index.jsx + MfaEnrollment/index.jsx]** The `TotpStep` and `StepVerify` components still share nearly identical logic (handleChange, handlePaste, errorMessage). The `normalizeOtp` extraction is a good step. If this pattern grows (e.g., adding rate-limit feedback or retry logic), consider extracting a shared `TotpInput` component to keep them in sync.

- **[packages/web-app/src/actions/Mfa.js:95]** Nice comment explaining why `postMfaLogin` hits the login endpoint. Small suggestion: also mention that this is an unauthenticated call (no bearer token) since the user hasn't completed login yet — it helps distinguish it from `postMfaVerify` which uses the enrollment token.

---

All issues from the previous review have been addressed:
- ✅ `isTokenExpired` now receives `authState` argument
- ✅ `AdminSessionExpiryBanner` uses a timer to re-render when the threshold is crossed
- ✅ Credentials are locked in a ref (`lockedCredentials`) before the initial login
- ✅ `normalizeOtp` extracted to shared utility
- ✅ `postMfaReset` handles 401 by dispatching `postLogout()` directly
- ✅ Logout delay (1500ms) added after MFA reset success
- ✅ `sessionStorage` key includes user ID
- ✅ `fullScreen={isMobile}` applied consistently to all dialogs

Well-structured feature with clean separation of concerns. The multi-step flows are well thought out and the error handling covers the edge cases nicely.

PR #1346, opened 2026-05-29
chore: some improvements for MFA

Paul-AUB:
- StepInstall et StepVerify utilisent maintenant le custom Alert (content prop, disableMargins) — <div> wrapper sur Fade car Alert est un composant fonctionnel sans forwardRef
- Clipboard : onError avec mfaCopyFailed traduit dans les 15 langues
- eslint-disable commenté avec l'explication
- Commentaire postMfaLogin complété avec "unauthenticated"


ClemRz (approved):

- **[packages/web-app/src/components/appli/MfaEnrollment/index.jsx:20]** The custom `Alert` import comes after the MUI block but before `qrcode.react`. Consider grouping it with the other internal component imports (after all external library imports) to follow the project's import order convention (external libs → internal components → utilities → types → styles).

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/MfaEnrollment/index.jsx:194-210]** The `<div>` wrapper around `<Alert>` inside `<Fade>` is necessary since the custom Alert doesn't use `forwardRef`. A brief inline comment explaining *why* the wrapper exists (e.g., `{/* div needed: Alert lacks forwardRef for Fade */}`) would help future maintainers avoid removing it.
- **[packages/web-app/src/components/appli/MfaEnrollment/index.jsx:40]** The `StepInstall` Alert doesn't have a `<Fade>` + `<div>` wrapper while `StepVerify` does. This is fine functionally (no transition needed there), just noting the inconsistency in case it was unintentional.

urien:
Je me suis connecté en changeant mon mot de passe et en déclarant une application tierce. Parfait

PR #1347, opened 2026-05-30
feat(map): add data quality display and filter to entrance markers

Paul-AUB:
# feat(map): data quality display & filter + measure control

## 🤔 What

- Add **MeasureControl** to entrance and network maps (was already present on the general map). Idea is to help user when doing the access walk.
- Improve **cave entrance marker colors** on the general map: distinct chocolate/brown shades with white stroke for contrast on all tile layers (OSM, satellite, dark).
- Display **data quality** in the entrance popup on the general map (badge + qualitative label: "Good quality", "Satisfactory quality", "Insufficient quality").
- Add a **"Filter by quality"** section to the map data control panel, alongside the existing "Filter by size".
- Standardize the help icon across the app: replace all `InfoOutlinedIcon` with `HelpOutlineIcon`.

## 🤷‍♂️ Why

- Users requested a way to filter entrances by data quality directly on the map.
- The popup lacked data quality information that the API now returns via the geoloc endpoint.
- The existing `DataQualityBadge` used MUI `CircularProgress` which breaks when rendered via `renderToString` (Leaflet popups): CSS injected by emotion is lost in the HTML string output. The badge was rewritten as a pure inline SVG to be SSR-safe.
- Data quality logic (thresholds, colors, label keys) was duplicated in several places; it is now centralized in `src/utils/dataQuality.js`.

## 🔍 How

- **`src/utils/dataQuality.js`** (new): single source of truth for thresholds, levels, colors, label keys, and the two helper functions `getDataQualityValue` (normalizes the two API shapes: flat number vs `{ total, categories }`) and `getDataQualityLevel`.
- **`DataQualityBadge`**: rewritten as a pure inline SVG with proportional `strokeWidth` and `fontSize` — no MUI/emotion dependency, renders correctly inside Leaflet `renderToString` wrappers.
- **`MapClusters/constants.js`**: adds `CAVE_QUALITY`, `getCaveQuality`, `CAVE_QUALITY_BADGE_VALUE`, `ENTRANCE_QUALITY_FILTERS` derived from the shared utils.
- **`MapClusters/index.jsx`**: `activeQualityFilters` state; combined size+quality filtering in `useMemo`; entrances with no quality data are always shown.
- **`DataControl.jsx`**: new "Filter by quality" section with `DataQualityBadge` color dots and `DataQualityHelpButton` tooltip.
- **`EntrancePopup.jsx`**: shows `DataQualityBadge` + qualitative label when `dataQuality` is present.
- **CSS fix** (`App.css`): `.map-popup-property > svg` (direct child) instead of `.map-popup-property svg` to prevent MUI icon SVGs from being resized by the popup rule.
- All 15 language files updated with the 4 new i18n keys.

## 🔗 Related API PR

_N/A — consumes the existing `dataQuality` field already returned by the geoloc endpoint._
Thanks @ClemRz !

## 🧪 Testing

1. Open the general map; zoom in to markers mode.
2. Verify entrance markers show in three distinct sizes and brown shades (small = light chocolate, medium = saddle brown, large = very dark brown), all with a white stroke visible on satellite view.
3. Click an entrance with quality data → popup shows a colored circular badge + label ("Good quality" / "Satisfactory quality" / "Insufficient quality").
4. Open the data control panel (eye icon) → "Filter by quality" section appears when in markers mode for entrances.
5. Uncheck a quality level → matching entrances disappear; entrances without quality data remain visible.
6. Open the entrance or network map (full-screen or embedded) → measure tool is present.

## 📸 Previews

image

image

image


ClemRz (changes requested):
### Issues (Must Fix)

- **[src/App.css:75-80]** The CSS rule `.map-popup-property > svg` forces `width: 25px; height: 25px` on direct SVG children. In `EntrancePopup`, the `DataQualityBadge` (which renders a raw `<svg>`) is passed as `icon` to `<Information>`, making it a direct child of `.map-popup-property`. This will override the `size={36}` prop and squash the badge to 25x25px. Consider wrapping the badge in a `<span>` to escape the selector, or adding a more specific override:
  ```css
  .map-popup-property > svg.data-quality-badge {
    width: auto;
    height: auto;
  }
  ```
  Alternatively, wrap the SVG in a `<span>` inside `DataQualityBadge` or at the call site in `EntrancePopup`.

- **[src/components/common/DataQualityBadge/index.jsx]** The component uses `useIntl()` which works fine in the `renderToString` context (since `useRenderPopup` wraps with `IntlProvider`). However, if this badge is ever used outside a React tree (e.g., directly in a Leaflet DivIcon HTML string without the provider wrapper), the hook will throw. Consider adding a brief comment noting this dependency, since the file header says "no MUI/emotion" but doesn't mention the `react-intl` requirement.

- **[src/components/common/Maps/MapClusters/DataControl.jsx:282]** The `HelpOutlineIcon` reference in the JSX uses the identifier `HelpOutline` (without `Icon` suffix) but the import is `HelpOutlineIcon`. Looking at the diff more carefully, the JSX shows `<HelpOutline` — this might be a rendering artifact in the diff. If the actual code uses `HelpOutlineIcon` correctly (which it does based on reading the full file), disregard this. Just flagging in case.

- **[src/utils/dataQuality.js:33]** The comment "Data quality == null is handled as insufficient" on `getDataQualityLevel` could be misleading. The function technically handles `null` (since `null >= 70` is false), but callers are expected to guard with `getDataQualityValue` first. Consider rewording to clarify that callers should pass only numeric values, and null-handling is a defensive fallback rather than an intended code path.

### Nitpicks (Optional)

- **[src/components/common/Maps/MapClusters/constants.js:1]** The `brown` import from `@mui/material/colors` is still used (for `ENTRANCE_HEAT_COLORS`), but it might be worth adding a brief comment near the hardcoded hex colors (`#D2691E`, `#8B4513`, `#2C0F00`) explaining why they don't use the `brown` palette — i.e., the palette shades didn't provide enough contrast differentiation on all tile layers.

- **[src/components/common/Maps/MapClusters/DataControl.jsx]** The quality filter section is only shown when `selectedHeats.has(heatmapTypes.ENTRANCES) && isMarkersMode`. This is consistent with the size filter behavior, but it means users can't pre-configure quality filters before zooming in. Not a bug — just noting the UX choice is intentional.

Overall this is a well-structured PR. The centralization of data quality logic into `src/utils/dataQuality.js` is a solid improvement, and the SVG rewrite of `DataQualityBadge` correctly solves the emotion/renderToString incompatibility. The filtering logic is clean and the "always show entrances without quality data" decision is sensible. Nice work!

ClemRz (approved):

- **[src/components/common/Maps/common/Markers/Components/EntrancePopup.jsx:33-35]** The location line uses `entrance.city && entrance.city` which evaluates to `false` (not an empty string) when `city` is falsy, producing output like `false, false` or `false, SomeRegion`. The same pattern existed before this PR so it's not a regression, but since the component was refactored here, it might be worth fixing with a simple fallback:
  ```jsx
  value={[entrance.city, entrance.region].filter(Boolean).join(', ')}
  ```

- **[src/components/common/DataQualityBadge/index.jsx]** The `<title>` element inside the SVG provides an accessible name, which is good. However, when the badge is used at small sizes (e.g., `size={20}` in the filter panel), the tooltip won't be visible on hover in most browsers since SVG `<title>` tooltip behavior is inconsistent. Not blocking — just noting that the filter panel badges won't have hover tooltips like the old MUI `<Tooltip>` provided. The popup badge at `size={36}` is fine since the label text is shown next to it.

- **[src/utils/dataQuality.js]** Consider adding a brief JSDoc or comment on `getDataQualityColor` and `getDataQualityLabelKey` noting that they expect a numeric value (not null). Currently the only guard is the comment on `getDataQualityLevel`, but these two exported helpers are the ones most likely to be called directly by new consumers.

### Nitpicks (Optional)

- **[src/components/common/Maps/MapClusters/constants.js:1]** The `brown` import is still used (for `ENTRANCE_HEAT_COLORS`), but at first glance it looks unused since the `CAVE_SIZE_STYLE` block now uses hardcoded hex. The existing comment above `CAVE_SIZE_STYLE` explains the rationale well — just noting that a reader scanning imports might momentarily wonder.

- **[src/components/common/Maps/MapClusters/DataControl.jsx:325]** Minor: the quality filter section title uses `DataQualityHelpButton` (an `IconButton` with popover) while the size filter section uses a plain `HelpOutlineIcon` with `onClick` to open a `Popover`. The UX is slightly inconsistent — the quality help icon has button semantics (focus ring, aria) while the size one is just a styled icon. Not blocking, but unifying them in a follow-up would improve accessibility.

Overall the previous review's issues have been addressed cleanly. The `<span>` wrapper in `DataQualityBadge` correctly prevents the `.map-popup-property > svg` selector from overriding the badge size. The centralization of data quality logic into `src/utils/dataQuality.js` is well-structured, the filtering logic is sound (null quality → always shown), and the SVG rewrite is a solid fix for the renderToString/emotion incompatibility. The i18n keys are properly added to all 15 language files in sorted order. Nice iteration.

urien (approved):
De nombreux retours positifs
Une remarque a été formulée : lorsqu'on arrive sur la carte avec les clusters, si on va dans les menus les choix n’apparaissent pas. Les utilisateurs risquent de ne pas penser à retourner dans le menu une fois que les cavités apparaissent sous forme de points.
Est ce qu'il serait possible que les menus liés au choix sur les entrées soient visibles mais non sélectionnable quand on est avec un affichage sous forme de cluster. Cela permettrait de montrer que le choix existe

Issue #1208, opened 2026-03-10
Legislation guidelines user interface


ClemRz:
# Requirements Document

## Introduction

This feature adds frontend support for displaying and managing legislation links within GrottoCenter. Guidelines are directly attached to countries, regions (ISO 3166-2), and massifs via the backend API — there is no automatic inheritance between geographic levels, since massifs can span multiple countries or regions. The frontend provides CRUD forms for authenticated users to create and edit guideline records, displays guidelines on geographic entity pages (country, region, massif), and shows collected guidelines on cave and entrance pages so cavers are informed about relevant regulations before visiting. Delete and restore of guidelines are restricted to Moderators. Any authenticated user can roll back a non-deleted guideline to a previous snapshot version. The backend API handles collecting guidelines from directly associated entities for cave/entrance pages and returns them grouped by source — the frontend displays what the API returns. Full history/versioning is supported.

## Glossary

- **Guideline**: A legislation link record attached to a geographic entity (country, region, or massif). Each guideline contains a URL, a title, an optional description, a language, and metadata (author, dates, deletion status).
- **Geographic_Entity**: A country (TCountry), region (TISO31662), or massif (TMassif) to which guidelines can be directly attached.
- **Guideline_API**: The set of REST API endpoints on the backend responsible for creating, reading, updating, deleting, restoring, and rolling back guideline records, as well as retrieving guideline history. The frontend consumes these endpoints.
- **Guideline_Snapshot**: A historical version of a guideline record, capturing the state of the guideline at a specific point in time, enabling rollback.
- **Guideline_Store**: The Redux state slices responsible for holding guideline data, guideline history, loading status, and error information.
- **Guideline_Form**: The MUI-based form component used by authenticated users to create or edit guideline records.
- **Guidelines_Section**: The UI section on entity pages that displays a list of guideline records with their title, URL link, and description.
- **Guideline_History**: The UI component that displays the chronological list of snapshots for a guideline, allowing authenticated users to roll back to a previous version.
- **Authenticated_User**: A caver who has logged in and possesses a valid JWT token, represented by a non-expired `authTokenDecoded` in the Redux login state.
- **Moderator**: A caver with the Moderator or Administrator role as determined by `usePermissions().isModerator`.

### Requirement 1: Guideline API Route Configuration

**User Story:** As a developer, I want all guideline API endpoints defined in the central API routes configuration, so that endpoint URLs are consistent and never hardcoded across the codebase.

[...]

PR #1364, opened 2026-06-02
Feat/1208 legislation guidelines UI

dawoldo:
# feat(Guideline): implement legislation guidelines UI for countries, regions, and massifs (#1208)

## 🤔 What

Implemented the front-end user interface and state management for managing and viewing **Legislation Guidelines** across GrottoCenter entities:

- Created guidelines management (CRUD + soft delete/restore) at **Country**, **Region**, and **Massif** levels.
- Implemented a grouped guidelines visualization on the **Network / Cave** page, displaying guidelines inherited from the cave's associated country, region, and massif.
- Integrated guideline revision tracking, allowing users to view history snapshots of guidelines and roll back to previous versions.

## 🤷‍♂️ Why

Speleological activities, cave locations, and access rights are often subject to regional regulations, environmental laws, safety policies, or property agreements. The Legislation Guidelines feature provides an intuitive way for contributors and moderators to document these local rules and requirements. Placing them on countries, regions, and massifs and aggregating them on the cave view ensures users can easily find safety/access regulations. Implementing a complete audit log/rollback system ensures these critical guidelines remain secure and auditable.

### 1. API and Route Definition
- Configured REST endpoints in `apiRoutes.js` for guidelines creation, updates, deletion, restoration, and rollbacks.
- Corrected update endpoint to use the `PATCH` method instead of `PUT` to align with backend specifications.

### 2. Redux Actions & State Management
- Implemented async actions under `packages/web-app/src/actions/Guideline/`:
  - `CreateGuideline`: handles new guideline submissions.
  - `UpdateGuideline`: updates existing guideline values (title, description, language).
  - `DeleteGuideline` / `RestoreGuideline`: handles soft-deletion/restoration of guidelines.
  - `RollbackGuideline`: rolls back a guideline to a specific snapshot state.
  - `GetGuidelineSnapshots`: fetches the revision history.
- Created corresponding reducers (registered in `GCReducer.js`) to handle actions and track loading/error states.
- Enhanced entity reducers (`CountryReducer.js`, `RegionDetailsReducer.js`, and `MassifReducer.js`) to seamlessly catch guideline action outcomes (e.g. `POST_GUIDELINE_SUCCESS`, `PUT_GUIDELINE_SUCCESS`) and update local state immutably.

### 3. UI Components
- **`GuidelineForm`**: Standardized form using `react-hook-form` to capture titles (max 150 chars), language, and descriptions (max 500 chars). Normalizes the language property from full object to ID before dispatching updates.
- **`Guideline`**: Renders individual guidelines using the markdown-based `Contribution` component. Integrates inline `ActionButtons` for authorized modifications.
- **`Guidelines`**: Section wrapper to show and create guidelines on entity detail pages.
- **`GuidelinesGrouped`**: Combines country, region, and massif guidelines into categorized listings, integrated into the `Network` page.

### 4. Revision History and Snapshot Restore
- Updated `GenericSnapshots` to properly resolve `description` in addition to `body` fields to prevent fallback translations ("inconnu") on guideline snapshots.
- Programmed `RestoreSnapshot` to map the `guidelines` snapshot type to the `rollbackGuideline` action.

## 🔗 Related API PR

[Related API PR](https://github.com/GrottoCenter/grottocenter-api/pull/1611)

---

## 🧪 Testing

1. Log in as a moderator or administrator.
2. Navigate to a country (e.g., `/ui/countries/FR`), a region, or a massif page.
3. In the **Guidelines** section, click **"Add a new guideline"** and create a guideline.
4. Verify you can edit, delete, and restore the guideline.
5. Click the history/snapshot icon to view past versions. Edit the guideline again, then try restoring a previous snapshot to test the rollback functionality.
6. Navigate to a cave/network page (`/ui/caves/:id`) and verify that all guidelines matching the cave's country, region, and massif are correctly aggregated and grouped under the **Guidelines** accordion.

---

## 📸 Previews

image
image
image


ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:38]** The `actionElement` prop is passed to `ScrollableContent`, but that component does not define or use an `actionElement` prop. This means the "Add a new guideline" button will never be rendered. The prop is silently ignored. Consider using the `icon` prop slot (which renders in the card header), or extending `ScrollableContent` to accept a new prop for action buttons.

- **[packages/web-app/src/actions/Guideline/DeleteGuideline.js:25-26]** The action only destructures `{ id }` from the argument, dropping `isPermanent`. The `Guideline.jsx` component calls `dispatch(deleteGuideline({ id: guideline.id, isPermanent }))`, but `isPermanent` is never forwarded to `deleteGuidelineUrl(id)`. This means permanent deletion will never work. Other delete actions in the codebase (e.g., `DeleteHistory`, `DeleteComment`) all pass `isPermanent` through:
  ```suggestion
  export const deleteGuideline =
    ({ id, isPermanent }) =>
    (dispatch, getState) => {
      dispatch(deleteGuidelineAction());

      const requestOptions = {
        method: 'DELETE',
        headers: getState().login.authorizationHeader
      };

      return fetch(deleteGuidelineUrl(id, isPermanent), requestOptions)
  ```

- **[packages/web-app/src/actions/Guideline/GetGuidelineSnapshots.js]** The `fetchGuidelineSnapshots` action is defined but never imported or used anywhere. The snapshot page uses the generic `fetchSnapshot` action. If this action is intended for future use, consider adding a comment stating so. Otherwise, it's dead code that adds to the bundle and maintenance surface.

- **[packages/web-app/src/actions/Guideline/UpdateGuideline.js]** The action type constants are named `PUT_GUIDELINE*` and the URL builder is `putGuidelineUrl`, but the actual HTTP method is `PATCH`. This is confusing — the naming convention suggests PUT semantics (full replacement) while the behavior is PATCH (partial update). Consider renaming to `PATCH_GUIDELINE` / `patchGuidelineUrl` / `patchGuideline` for clarity, or at least add a comment explaining the discrepancy.

- **[packages/web-app/src/types/guideline.type.js:6]** The `language` field is typed as `string`, but the `GuidelineForm` normalizes it with `values.language?.id ?? values.language`, implying the API returns a language object (with an `id` property). Consider typing it as `oneOfType([string, shape({ id: string })])` to avoid PropTypes console warnings when the API returns the object form.

- **[packages/web-app/src/components/appli/Guidelines/Guideline.jsx:34-37]** The `useEffect` initializes `wantedDeletedState` from `guideline.isDeleted` on mount with an empty dependency array and an eslint-disable comment. If the guideline's deletion status changes from an external source (e.g., websocket or another tab), the state goes stale. Consider including `guideline.isDeleted` in the dependency array, or initializing via `useState(guideline.isDeleted)` directly instead of using an effect.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:38, GuidelinesGrouped.jsx:27]** Both components pass `id="guidelines"` to `ScrollableContent`, but that prop is not destructured or used by the component (it uses `anchorId` for the DOM `id`). The extra prop is harmless but unnecessary — consider removing it.

- **[packages/web-app/src/components/appli/EntitiesForm/Guideline/index.jsx:46-52]** The validators `titleValidator` and `descriptionValidator` can be simplified to single expressions:
  ```js
  const titleValidator = value => !value || value.length <= 150;
  const descriptionValidator = value => !value || value.length <= 500;
  ```

- **[packages/web-app/src/reducers/GuidelineSnapshotsReducer.js]** This reducer is registered in `GCReducer.js` but only `fetchGuidelineSnapshots` would populate it, and that action is never dispatched. This means the slice will always hold its initial state. Consider removing it if `GetGuidelineSnapshots` is truly unused, or wire it up.

ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:97-106]** The `fetch(getGuidelinesUrl)` call in MODE_ATTACH bypasses Redux and does not include authorization headers. Since `getGuidelinesUrl` hits the API, this may fail if the endpoint requires auth. More critically, this raw `fetch` with no error-handling for non-200 responses (e.g., 500 or 403) will silently call `response.json()` on an error response, potentially causing a runtime crash or rendering broken data. Consider either adding a status check (`if (!response.ok) throw …`) or routing through an action creator that uses `checkAuthStatus`.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:110-125]** `onSubmitCreateForm` dispatches `postGuideline` then immediately calls `closeForm()` without awaiting the result. If the API call fails, the user won't see any error feedback — the form closes and the guideline silently doesn't appear. The `handleAttachGuideline` function correctly awaits dispatch, but this handler does not. Consider awaiting the dispatch and only closing on success:
  ```suggestion
  const onSubmitCreateForm = async data => {
    const entities = { countries: [], regions: [], massifs: [] };
    if (entityType === 'countries') entities.countries = [entityId];
    else if (entityType === 'regions') entities.regions = [entityId];
    else if (entityType === 'massifs') entities.massifs = [entityId];

    const result = await dispatch(
      postGuideline({
        ...entities,
        title: data.title,
        description: data.description,
        language: data.language
      })
    );
    if (result) closeForm();
  };
  ```

- **[packages/web-app/public/lang/fr.json]** The validation messages `"Description must be less than 500 characters."` and `"Title must be less than 150 characters."` are left untranslated in all non-English language files (including French, which has translations for the other new keys). These are user-facing validation errors that should be translated — at least in French where the rest of the keys are properly localized.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:97]** The condition `allGuidelines.length === 0` means the fetch fires only once and there's no way to refresh the list if it was empty or if a new guideline was just created. If a user creates a guideline in MODE_CREATE and then switches to MODE_ATTACH, the new guideline won't appear in the autocomplete. Consider invalidating `allGuidelines` when switching back to MODE_ATTACH or using a dependency on `guidelines.length`.

- **[packages/web-app/src/reducers/CountryReducer.js:48-53, RegionDetailsReducer.js:44-49]** The `POST_GUIDELINE_SUCCESS` handling in CountryReducer and RegionDetailsReducer compares `action.guideline.countries` / `action.guideline.regions` elements using `String(c) === String(state.country.id)`. However, looking at the `patchGuideline` action (used for attach), the payload sends `countries: [entityId]` where `entityId` could be a string like `"FR"` for countries or a number for massifs. If the API returns the full guideline with populated arrays of objects (e.g., `{id: "FR", name: "France"}`), this `String(c)` comparison would fail because `String({id: "FR"})` is `"[object Object]"`. Consider verifying the API response shape or using `String(c.id || c)` — consistent with what the component does for `availableGuidelines`.

- **[packages/web-app/src/types/guideline.type.js:9-10]** The `countries` field is typed as `arrayOf(string)` and `regions` as `arrayOf(string)`, but `massifs` is typed as `arrayOf(shape({ id: number }))`. This inconsistency will trigger PropTypes warnings if the API response includes objects for countries/regions (as the `handleAttachGuideline` code in `index.jsx` seems to expect with `c.id || c`). Consider making these consistent — likely `arrayOf(oneOfType([string, shape({ id: oneOfType([string, number]) })]))` for all three.

- **[packages/web-app/src/components/appli/Guidelines/Guideline.jsx:31-33]** The previous review flagged the `useEffect` with empty deps for initializing `wantedDeletedState`. This has been updated to use `[guideline.isDeleted]` which is good, but initializing via `useState(guideline.isDeleted)` directly (without the effect) would be simpler. The effect is still needed if the external deletion state can change (e.g., via another user's action propagated through Redux), so the current approach is fine.

### Nitpicks (Optional)

- **[packages/web-app/src/conf/apiRoutes.js:189-192]** `deleteGuidelineUrl` produces a trailing `?` when `isPermanent` is false (e.g., `/guidelines/123?`). While harmless and consistent with `deleteHistoryUrl`, consider conditionally omitting the `?` for cleanliness: `${isPermanent ? '?isPermanent=1' : ''}`.

- **[packages/web-app/src/components/appli/EntitiesForm/Guideline/index.jsx:46-47]** The validators always return `true` when validation passes. `titleValidator` returns `true` for an empty value, but the field is marked `isRequired`, so react-hook-form handles empty-check separately. This works, but the validator name `titleValidator` is slightly misleading since it only validates length, not presence — consider naming it `titleLengthValidator` for clarity.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:75]** `ChoiceButton` is defined inside the same file but above `Guidelines`. This is fine stylistically, but since it has its own PropTypes and is self-contained, it could live in its own file for reusability. No action needed now, just noting for future refactors.

ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:49-59]** The `fetch(getGuidelinesUrl)` call does not include authorization headers, but pipes through `checkAuthStatus(dispatch)`. Two problems here:
  1. If `GET /guidelines` requires auth, the request will always get a 401, triggering `checkAuthStatus` to log the user out — a confusing UX.
  2. If `GET /guidelines` is a public endpoint that never returns 401, using `checkAuthStatus` is the wrong util — public endpoints in this codebase use `checkAndGetStatus` instead (see `Language.js`, `Licenses.js`, `DocumentType.js`).

  Either include the authorization header (requires access to Redux state, so this fetch should be moved to a proper action creator) or, if the endpoint is public, replace `checkAuthStatus(dispatch)` with `checkAndGetStatus`. Additionally, the `.catch` only logs to `console.error` — there's no user-facing error feedback if the fetch fails.

- **[packages/web-app/src/components/appli/Guidelines/Guideline.jsx:39-48]** `onSubmitForm` dispatches `patchGuideline` without awaiting the result, then immediately hides the form via `setIsUpdateFormVisible(false)`. If the API call fails, the user sees no error — the form closes and the guideline reverts silently. This was flagged in the first review for `onSubmitCreateForm` (which was fixed), but the same pattern persists here. Consider:
  ```js
  const onSubmitForm = async data => {
    const result = await dispatch(
      patchGuideline({
        id: guideline.id,
        title: data.title,
        description: data.description,
        language: data.language
      })
    );
    if (result) setIsUpdateFormVisible(false);
  };
  ```

- **[packages/web-app/public/lang/*.json]** All 20 new translation keys (`guidelines.*`, `"Country guidelines"`, `"Region guidelines"`, `"Massif guidelines"`, `"Description must be less than 500 characters."`, `"Title must be less than 150 characters."`) are left as English text in 13 non-French language files. While the project uses AI-generated translations and this may be acceptable as a follow-up task, at minimum the `guidelines.*` namespaced keys should have proper translations in the major languages since these are user-facing UI labels. Consider running the translation generation pipeline before merge.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:49]** The guideline list is fetched once when the user switches to `MODE_ATTACH` and is never invalidated. If a user just created a guideline in `MODE_CREATE`, switches to `MODE_ATTACH`, and searches for it — it won't appear because `allGuidelines` is stale. Consider refetching when transitioning to `MODE_ATTACH` regardless of whether `allGuidelines` is already populated:
  ```js
  // Remove the guard or use a different condition:
  useEffect(() => {
    if (mode === MODE_ATTACH) {
      setIsLoadingGuidelines(true);
      // ... fetch
    }
  }, [mode, dispatch]);
  ```
  This actually already re-runs on `mode` change, which is correct. But if the user goes CREATE → ATTACH → CREATE → ATTACH, the second ATTACH won't refetch because `mode` is the same value (`MODE_ATTACH`). Consider adding a counter or timestamp as a dependency to force re-fetch.

- **[packages/web-app/src/components/common/Layouts/Fixed/ScrollableContent.jsx:151-152]** The `content` prop is now optional (no longer `.isRequired`). This is a breaking change for any consumer that relied on the PropTypes warning to catch missing content. Existing usages of `ScrollableContent` in the codebase all pass `content` as a prop — consider keeping it required and adding `children` as an alternative via a custom validator, or at least requiring one of the two:
  ```js
  content: (props, propName, componentName) => {
    if (!props.content && !props.children) {
      return new Error(`${componentName} requires either 'content' or 'children' prop.`);
    }
    return null;
  },
  ```

### Nitpicks (Optional)

- **[packages/web-app/src/reducers/GCReducer.js]** The diff moves `verifyEmail` and `resendVerificationEmail` to different positions in the import list and `combineReducers` call. This is unrelated to the guidelines feature and muddies the git blame. Consider keeping unrelated import re-orderings out of this PR.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:256]** The "Add a new guideline" button's `onClick` handler sets `setMode(MODE_ATTACH)` — which opens the attach flow, not the create flow. The button label says "Add a new guideline" which could imply creating, but the UX intent (search first, then create) is valid. Just worth noting that the `formatMessage id` might be slightly confusing vs the actual initial behavior. A label like "Manage guidelines" or the existing `guidelines.attach_existing` might be clearer as the initial action.

ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:77-87]** `onSubmitCreateForm` checks `if (result)` to decide whether to close the form, but the `postGuideline` action creator returns a truthy value on both success (`dispatch(postGuidelineSuccess(data))`) and failure (`dispatch(postGuidelineFailure(...))`). The only falsy return is on auth error (`return undefined`). This means the form closes even when the API returns a non-401 error, and the user never sees feedback. Consider returning `null` explicitly in the `.catch` block of `CreateGuideline.js`:
  ```js
  .catch(error => {
    dispatch(
      postGuidelineFailure(
        makeErrorMessage(error.message, `Creating a new guideline`),
        error.message
      )
    );
    return null; // signal failure to the caller
  });
  ```

- **[packages/web-app/src/components/appli/Guidelines/Guideline.jsx:39-48]** Same issue: `onSubmitForm` does `const result = await dispatch(patchGuideline(...)); if (result) setIsUpdateFormVisible(false);` — but `patchGuideline` also returns truthy on failure. The form will always close regardless of the outcome. Apply the same `return null` fix to `UpdateGuideline.js` catch block.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:103-115]** `handleAttachGuideline` calls `closeForm()` unconditionally after `await dispatch(patchGuideline(...))`. If the attach API call fails, the user sees no error and the form disappears. Consider:
  ```js
  const result = await dispatch(
    patchGuideline({ id: selectedGuideline.id, countries, regions, massifs })
  );
  setIsSubmitting(false);
  if (result) closeForm();
  ```
  (Combined with the `return null` fix in `UpdateGuideline.js`.)

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:275]** The `guidelines.none` translation key receives `values={{ entityType }}` where `entityType` is the raw collection name ("countries", "regions", "massifs"). This produces grammatically incorrect messages like "No guidelines attached to this countries." Consider either mapping to a user-friendly singular label before passing, or using a different i18n key per entity type:
  ```js
  const entityLabel = formatMessage({ id: entityType === 'countries' ? 'country' : entityType === 'regions' ? 'region' : 'massif' });
  ```

- **[packages/web-app/public/lang/fr.json:1086]** `guidelines.fetch_error` is left as English text (`"Failed to load guidelines"`) in the French translation file. Other guidelines keys in French are properly translated. Consider: `"Échec du chargement des directives"`.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:49-65]** The `useEffect` dependency array includes `onError` and `formatMessage`. While both are memoized/stable, the linter may still warn. More importantly, the fetch has no abort controller — if the user rapidly toggles between modes, stale responses could overwrite newer state. Consider adding an `AbortController` or a boolean flag to ignore stale responses:
  ```js
  useEffect(() => {
    if (mode !== MODE_ATTACH) return;
    let cancelled = false;
    setIsLoadingGuidelines(true);
    fetch(getGuidelinesUrl)
      .then(checkAndGetStatus)
      .then(response => response.json())
      .then(data => { if (!cancelled) setAllGuidelines(data || []); })
      // ...
    return () => { cancelled = true; };
  }, [mode, attachFetchTrigger]);
  ```

- **[packages/web-app/public/lang/ar.json, bg.json, ca.json, el.json, he.json, id.json, ja.json, nl.json, ro.json]** All new guideline-related keys plus `"Description must be less than 500 characters."` and `"Title must be less than 150 characters."` remain as English text in these 9 language files. The project does use AI-generated translations, so this may be intentional for a follow-up pass, but it's worth noting for completeness before merge.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/Guidelines/Guideline.jsx:76]** `SnapshotButton` receives a `content` prop, but looking at `SnapshotButton`'s PropTypes definition in `UtilityFunction.js`, it doesn't accept a `content` prop. This prop is silently ignored and appears to be dead code.

- **[packages/web-app/src/reducers/GCReducer.js]** The imports for `verifyEmail` and `resendVerificationEmail` were repositioned alphabetically together with the new guideline imports. While the alphabetization is welcome, mixing unrelated import reordering into this PR muddies the diff. Not blocking, just noting for clean git history.

ClemRz (approved):

- **[packages/web-app/src/components/appli/Guidelines/Guideline.jsx:49-53]** `onDeletePress` and `onRestorePress` dispatch their respective actions without awaiting the result. If the delete/restore API call fails, the optimistic `wantedDeletedState` update goes stale — the spinner keeps spinning because `isActionLoading` is `true` (wanted ≠ actual), but there's no error feedback to the user and no way to recover without refreshing. Consider awaiting the dispatch and resetting `wantedDeletedState` on failure:
  ```js
  const onDeletePress = async isPermanent => {
    setWantedDeletedState(true);
    const result = await dispatch(deleteGuideline({ id: guideline.id, isPermanent }));
    if (!result) setWantedDeletedState(guideline.isDeleted);
  };
  const onRestorePress = async () => {
    setWantedDeletedState(false);
    const result = await dispatch(restoreGuideline({ id: guideline.id }));
    if (!result) setWantedDeletedState(guideline.isDeleted);
  };
  ```
  Note: this also requires adding `return null` to the catch blocks in `DeleteGuideline.js` and `RestoreGuideline.js` (which currently don't return anything on failure, meaning the promise resolves to `undefined` — a falsy value that would actually work but inconsistently with the explicit `return null` pattern in `CreateGuideline.js` and `UpdateGuideline.js`).

- **[packages/web-app/src/actions/Guideline/DeleteGuideline.js:44-48, RestoreGuideline.js:44-48, RollbackGuideline.js:44-48]** Unlike `CreateGuideline.js` and `UpdateGuideline.js` which explicitly `return null` in the catch block to signal failure to the caller, these three action creators don't return anything on error. This inconsistency means callers can't reliably use `if (result)` to check success. Consider adding `return null;` after the `dispatch(...Failure(...))` call in each of these catch blocks for consistency.

- **[packages/web-app/src/components/appli/Region/index.jsx:208-213]** The `region` PropTypes shape does not include `guidelines`, but the component accesses `region.guidelines?.length` at line 194. Consider adding `guidelines: PropTypes.array` to the shape definition for PropTypes completeness.

- **[packages/web-app/public/lang/ar.json, bg.json, ca.json, el.json, he.json, id.json, ja.json, nl.json, ro.json]** The `guidelines.*` namespaced keys are properly translated in these language files, but the standalone heading keys (`"Guidelines"`, `"Country guidelines"`, `"Region guidelines"`, `"Massif guidelines"`) remain as English text. This creates an inconsistent user experience where section headings appear in English while the controls and messages are localized. These are high-visibility section titles on entity pages. Consider running the translation pipeline for these 4 keys in the 9 affected files.

### Nitpicks (Optional)

- **[packages/web-app/src/reducers/GCReducer.js:80-82]** The `verifyEmail` and `resendVerificationEmail` imports and combineReducers entries are positioned between `signUp` and `restoreGuideline` instead of in strict alphabetical order (they should come after `updatePersonGroups`). While this was noted in the prior review as a diff-muddying change, the final result is still not alphabetical. Not blocking, just worth tidying up in a follow-up.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:57-84]** The `useEffect` dependency array includes `formatMessage` and `onError`. While both should be stable references from hooks, including them is technically unnecessary (they don't change between renders) and adds noise to the deps list. If ESLint's exhaustive-deps rule isn't flagging them, they could be omitted for clarity. Very minor, no action needed.

urien (approved):
Du point de vue utilisateur cela fonctionne parfaitement, j'ai pu créer une directive et la réutliser
Il y a un pb en page d'accueil lorsqu'on créé une directive elle apparait, mais n'est pas un objet cliquable. Il serait souhaitable qu'elle n'apparaisse pas. 

Image

A priori ce n'est pas a traiter dans ce ticket

urien (changes requested):
Oups..
J'ai réussi a associer une directive a différents massifs et a modifier les massifs pour appliquer la sensibilité
Désormais j'ai une erreur lorsque je souhaite éditer les massifs après avoir appliqué la directive
https://orange-rock-0d4f87503-1364.westeurope.azurestaticapps.net/ui/massifs/810/edit
Par contre il est possible déditer un massif qui n'a pas de directive
https://orange-rock-0d4f87503-1364.westeurope.azurestaticapps.net/ui/massifs/1032/edit

Comme administrateur je peux faire un soft delete sur une directive mais le bouton suppression définitive ne produit rien

une directive avec soft delete peut être remise en place

ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/components/appli/Guidelines/Guideline.jsx:89-91]** The `SnapshotButton` passes `type="guidelines"` but the `SnapshotPage` component (`Entry/Snapshots/index.jsx`) does not handle `guidelines` in its `entityByType` map or its `fetchByType`/`fetchParentByType` maps. This means the snapshot page cannot fetch the current guideline item — `currentTItem` is always `null`. The `RestoreSnapshot` component then receives `actualItem={null}`, which means the `!actualItem?.isDeleted` guard always passes, so the rollback button shows even for deleted guidelines (violating Requirement 9 AC7). Additionally, after the rollback completes, `window.opener.location.reload()` fires but the "back" navigation (`backTo`) works fine since `SnapshotButton` encodes the origin URL. Consider adding a `guidelines` case to `entityByType` and a fetch action, or at minimum passing the guideline data via query params or local state so `RestoreSnapshot` can correctly gate the rollback button on `isDeleted`.

- **[packages/web-app/public/lang/*.json]** The translation keys `guidelines.delete_error` and `guidelines.restore_error` are present in `en.json` but missing from **all** other language files (including `fr.json` which has the rest of the guideline keys translated). These are user-facing error messages that appear when delete/restore operations fail. They need to be added to the other language files to avoid `MISSING_TRANSLATION` console errors and English-only fallback in production.

- **[packages/web-app/src/actions/Guideline/RollbackGuideline.js:40-47]** The `.catch` block doesn't explicitly `return null` (or `return false`) on failure, unlike the other guideline action creators (`CreateGuideline`, `UpdateGuideline`, `DeleteGuideline`, `RestoreGuideline`) which do. While the implicit `undefined` return is currently falsy and happens to work for `if (result)` checks, this is fragile and inconsistent with the established pattern in this same PR. Consider adding `return null;` after the failure dispatch for clarity and consistency.

- **[packages/web-app/public/lang/*.json]** Several translation keys added to all 15 language files are not referenced anywhere in the source code: `guidelines.add`, `guidelines.attach_existing_description`, `guidelines.create_new_description`, `guidelines.attach_success`, and `guidelines.create_success`. These appear to be remnants from an earlier design iteration. Unused keys add maintenance burden — consider removing them, or if they're intended for future use, add a comment in the source indicating so.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:57-89]** The `useEffect` fetches `getGuidelinesUrl` (a public endpoint) using bare `fetch` → `checkAndGetStatus`. While this correctly uses `checkAndGetStatus` for a public endpoint (good fix from prior review), the fetch has no abort mechanism for component unmount beyond the `cancelled` flag. If the component unmounts while the fetch is in flight, the `.then` callbacks will still execute React state setters (`setAllGuidelines`, `setIsLoadingGuidelines`). The `cancelled` flag prevents this in the happy path, but if `checkAndGetStatus` throws (4xx/5xx), the `.catch` correctly checks `cancelled` too — so this is actually fine. Just noting for awareness.

- **[packages/web-app/src/components/appli/Guidelines/Guideline.jsx:97-101]** The `Guideline` component always renders `ActionButtons` when `isEditAllowed` is true, but in `GuidelinesGrouped` (cave/network page), `isEditAllowed={false}` is passed, correctly hiding edit controls. However, the `Guideline` component still renders the enclosing `<Box sx={{ float: 'right' }}>` wrapper even when all buttons inside would be hidden (if `isEditAllowed` is somehow true but the user lacks all permissions). This is minor since the wrapper is empty in that case, but consider guarding the wrapper render on `permissions.isAuth` to avoid an empty floating box.

- **[packages/web-app/src/reducers/CountryReducer.js:49-53 / RegionDetailsReducer.js:44-49]** The `POST_GUIDELINE_SUCCESS` case uses `String(c.id || c) === String(state.country.id)` for comparison. For countries, the ID is a string like `"FR"`, while for regions it could be numeric. The `|| c` fallback handles the case where the API returns a bare string/number instead of an object. This works, but `String(c.id || c)` would produce `"undefined"` if the API ever returns `{ id: undefined }`. A safer pattern would be `String(c?.id ?? c)` (nullish coalescing) to only fall back when `id` is null/undefined rather than any falsy value like `0`. Not blocking, but worth hardening.

### Nitpicks (Optional)

- **[packages/web-app/src/reducers/GCReducer.js:78-82]** The `verifyEmail` and `resendVerificationEmail` entries are positioned between `signUp` and `restoreGuideline` in the `combineReducers` call, breaking alphabetical order. They should come after `updatePersonGroups`. This was noted in prior reviews — not blocking, but keeping the list sorted helps navigation in this large file.

- **[packages/web-app/public/lang/ar.json, bg.json, ca.json, etc.]** The heading keys `"Country guidelines"`, `"Region guidelines"`, and `"Massif guidelines"` remain untranslated (English text) in all non-English files. These are visible section headings on entity pages. While the project uses AI-generated translations and this can be a follow-up, it's worth noting as it creates an inconsistent UX where `"Guidelines"` is translated (in `fr.json` → "Directives d'accès") but the sub-headings stay in English.

- **[packages/web-app/src/components/appli/EntitiesForm/Guideline/index.jsx:46-47]** The validators `titleLengthValidator` and `descriptionLengthValidator` return `true` for falsy values (empty string). This is fine because the `title` field has `isRequired`, and `description` is optional. But the function names could be slightly more precise (e.g., `isTitleLengthValid`) to make it clear they validate length constraints only.

urien (approved):
Tout est parfait pour moi. Merci @dawoldo

ClemRz (approved):

- **[packages/web-app/src/reducers/GCReducer.js:93-94, 199-200]** `updateGuideline` is imported _before_ `updateEntrance`, breaking alphabetical order in both the import list and the `combineReducers` call. Alphabetical ordering is the established convention in this file:
  ```suggestion
  import updateEntrance from './UpdateEntranceReducer';
  import updateGuideline from './UpdateGuidelineReducer';
  ```
  Same swap needed in the `combineReducers` object.

- **[packages/web-app/public/lang/ar.json, bg.json, ca.json, el.json, he.json, id.json, ja.json, nl.json, ro.json]** The `"Guidelines"` key (the main section heading) remains untranslated in 7+ languages — it renders as English "Guidelines" while the other `guidelines.*` namespaced keys in the same files are properly localized. This creates an inconsistent UX where the section title is in English but the buttons/messages inside are in the user's language. Consider running the translation pipeline for this one key before merge.

- **[packages/web-app/src/components/appli/Region/index.jsx:216-221]** The `region` PropTypes shape does not include `guidelines`, but the component accesses `region.guidelines?.length` at line 193. Adding it prevents PropTypes warnings in development:
  ```js
  region: PropTypes.shape({
    id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
    name: PropTypes.string,
    latitude: PropTypes.number,
    longitude: PropTypes.number,
    guidelines: PropTypes.array
  }),
  ```

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:62-80]** The `useEffect` dependency array includes `formatMessage` and `onError`. While both are stable hook references that won't trigger re-runs, including them adds noise to the deps. If ESLint's exhaustive-deps rule doesn't require them (since they're stable), consider removing them for clarity.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/EntitiesForm/Massif/PolygonMap.jsx]** The `KinkPane` component fix for the massif edit crash is a good solution to the race condition. It was introduced because this branch changed something that surfaced the issue. Worth noting in the PR description for reviewers unfamiliar with the commit history.

- **[packages/web-app/src/components/appli/Guidelines/index.jsx:284]** The initial "add guideline" button uses `guidelines.attach_existing` as its label and starts in `MODE_ATTACH` mode. This is intentional per the UX discussion (search first, create second), but a more neutral entry label like "Add a guideline" might be slightly less confusing. Already discussed in PR comments and accepted — not blocking.

- **[packages/web-app/src/reducers/GCReducer.js:100-104]** The imports at the end of the import block (`countryOrganization`, `massifOrganization`, `regionOrganization`, `updatePersonGroups`, `verifyEmail`) break alphabetical ordering relative to the rest of the file. These pre-date this PR, but since the PR touches this area anyway, it could be a good opportunity to sort them. Very optional.

PR #1367, opened 2026-06-03
chore(ci): group dependabot PRs and ignore blocked upgrades

ClemRz:
# chore(ci): group dependabot PRs and ignore blocked upgrades

## 🤔 What

Reconfigure Dependabot to:
- Group all dependency updates into a single PR per directory (root `/` and `/packages/web-app/`)
- Ignore major version bumps for packages we can't upgrade yet

## 🤷‍♂️ Why

Dependabot was opening individual PRs for each dependency, most of which got closed without merging because we handle upgrades in manual batches. This reduces noise and stops it from repeatedly proposing upgrades we've intentionally skipped.

## 🔍 How

Added `groups` with a `*` wildcard pattern so all updates are batched into one PR per ecosystem entry.

Added `ignore` rules for blocked packages (from PR #1263 analysis):

| Package | Reason |
|---|---|
| `eslint` (major) | Pinned at v8 — `eslint-config-airbnb` doesn't support v9+ flat config |
| `eslint-plugin-cypress` (major) | v6 targets ESLint v10 flat config |
| `eslint-plugin-storybook` (major) | Tied to Storybook v10 upgrade |
| `typescript` (major) | v6 has breaking defaults (`strict: true`, `types: []`) |
| `@storybook/*` (major) | Requires coordinated ecosystem migration |
| `storybook` (major) | Same as above |
| `react-intl` (major) | v9/v10 require migration guide |
| `fast-check` (all) | v4.6.0 uses `pure-rand` subpath exports incompatible with Jest 27 / CRA 5.x |
| `leaflet.locatecontrol` (all) | Pre-1.0, no changelog; conservative skip |

Removed explicit `open-pull-requests-limit` since the default (5) is more than enough when everything is grouped.

## 🔗 Related API PR

N/A

## 🧪 Testing

N/A — config-only change. Dependabot will pick up the new config on next scheduled run.

## 📸 Previews

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,20 +1,49 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
-# Please see the documentation for all configuration options:
 # https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
 updates:
-  - package-ecosystem: 'npm' # See documentation for possible values
+  - package-ecosystem: 'npm'
     directory: '/'
-    # Updates currently disabled: re-enable them (with the value 5) when we had tests checking everything is ok after updating the packages
-    open-pull-requests-limit: 5
     schedule:
       interval: 'monthly'
+    groups:
+      all-root-dependencies:
+        patterns:
+          - '*'
+    ignore:
+      # Pinned at v8 — eslint-config-airbnb does not support ESLint v9+ flat config
+      - dependency-name: 'eslint'
+        update-types: ['version-update:semver-major']
+      # v6 targets ESLint v10 (flat config); incompatible with our v8 setup
+      - dependency-name: 'eslint-plugin-cypress'
+        update-types: ['version-update:semver-major']
+      # Tied to Storybook v10 major upgrade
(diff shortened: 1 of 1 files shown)


Paul-AUB (approved):

1. **[.github/dependabot.yml — web-app section]** `typescript` is ignored in the root `/` ecosystem entry but not in the `/packages/web-app/` entry. If `typescript` (or `ts-node`, etc.) appears in `packages/web-app/package.json` as a dev dependency, a major bump would still trigger a grouped PR for that directory. Worth confirming whether it needs to be mirrored in the web-app `ignore` list.

2. **[.github/dependabot.yml:58-59]** `fast-check` and `leaflet.locatecontrol` have no `update-types` restriction, meaning all versions (including patches) are silently suppressed. This is documented in the PR description, but consider adding a comment directly in the YAML (similar to the others) so the rationale is visible when someone revisits the file months from now without the PR context.

### Nitpicks (Optional)

3. **[.github/dependabot.yml — root section]** `eslint-plugin-cypress` and `eslint-plugin-storybook` are blocked under the root `/` directory. If those plugins are also installed under `/packages/web-app/`, the ignores would need to be mirrored there too. Low-risk since the monorepo likely hoists them, but worth a quick check.

PR #1371, opened 2026-06-05
fix(massif): description not saved/displayed after creation

Paul-AUB:
# fix(massif): description not saved/displayed after creation

## 🤔 What

- Fix massif description not appearing after creation
- Add missing Redux action handlers in `MassifReducer` for all description lifecycle actions

## 🤷‍♂️ Why

When a user created a description on a massif, the description was correctly sent to the API and saved in the database, but never appeared in the UI afterward. The Redux state for the massif was never updated because `MassifReducer` did not listen to the description action types.

Closes #1368

## 🔍 How

`CaveReducer` and `EntranceReducer` already handle `POST_DESCRIPTION_SUCCESS`, `UPDATE_DESCRIPTION_SUCCESS`, `DELETE_DESCRIPTION_SUCCESS`, `RESTORE_DESCRIPTION_SUCCESS`, and `DELETE_DESCRIPTION_PERMANENT_SUCCESS` to keep their entity's `descriptions` array in sync with the server response via `arrFindReplaceOrAdd`.

`MassifReducer` was missing all five of these cases. Added them following the exact same pattern as the other two reducers.

## 🔗 Related API PR

None — this is a pure frontend fix; the API was already creating the description correctly.

## 🧪 Testing

1. Open any massif page (e.g. one with no existing description)
2. Click **New** in the Description section
3. Fill in title, body, and language, then submit
4. The description should appear immediately without requiring a page reload

## 📸 Previews

_No visual change beyond the description now appearing after creation._

### DIFF
--- a/packages/web-app/src/reducers/MassifReducer.js
+++ b/packages/web-app/src/reducers/MassifReducer.js
@@ -13,6 +13,13 @@ import { RESTORE_MASSIF_SUCCESS } from '../actions/Massif/RestoreMassif';
 import { LINK_DOCUMENT_TO_MASSIF_SUCCESS } from '../actions/LinkDocumentToMassif';
 import { UNLINK_DOCUMENT_TO_MASSIF_SUCCESS } from '../actions/UnlinkDocumentToMassif';
 import { MOVE_DESCRIPTION_RELEVANCE_SUCCESS } from '../actions/Description/MoveRelevance';
+import { POST_DESCRIPTION_SUCCESS } from '../actions/Description/CreateDescription';
+import { UPDATE_DESCRIPTION_SUCCESS } from '../actions/Description/UpdateDescription';
+import {
+  DELETE_DESCRIPTION_SUCCESS,
+  DELETE_DESCRIPTION_PERMANENT_SUCCESS
+} from '../actions/Description/DeleteDescription';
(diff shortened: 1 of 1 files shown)


urien (approved):
Parfait, cela fonctionne

Image

ClemRz (approved):

- **[MassifReducer.js:66-91]** Pre-existing architectural concern (not blocking): The description actions (`POST_DESCRIPTION_SUCCESS`, etc.) are dispatched globally without any entity-type context. This means all three reducers (Cave, Entrance, Massif) will process the same action. If a user creates a description for a massif while `state.cave` still holds data from a previous page visit, `arrFindReplaceOrAdd` will add that description to the cave's descriptions array too. This PR correctly follows the established pattern from `CaveReducer` and `EntranceReducer`, so it's not something to block on here — but it might be worth a follow-up issue to scope description actions per entity (e.g., via an `entityType` field on the action payload).

### Nitpicks (Optional)

- **[Branch name]** The project convention uses `fix/<scope>` for bug fixes (per steering docs). This branch is named `feature/fix-massif-description` — ideally it would be `fix/massif-description`. Very minor, no action needed.

---

Good fix — clean, minimal, and follows the existing pattern exactly. The bug was clearly identified and the solution is consistent with how `CaveReducer` and `EntranceReducer` already handle these cases. LGTM.

PR #1376, opened 2026-06-08
feat(entrance): misc UX improvements — geolocation, full-screen map, document display

Paul-AUB:
# feat(entrance): misc UX improvements — geolocation, full-screen map, document display

## 🤔 What

- Add a "Use my location" button in the entrance creation/edit map to auto-center the marker on the user's GPS position, with translated error messages for denied access, timeout, and indeterminate position.
- Enable full-screen mode on the map marker selector used during entrance creation/edit.
- Align the "Associated documents" section display in Entry and Massif pages to be consistent with the redesigned Document page style.
- Group Dependabot PRs and ignore blocked upgrades in CI configuration.

## 🤷‍♂️ Why

Locating an entrance on the map by manually panning is cumbersome when the user is in the field with GPS available. The full-screen option gives more space for precise marker placement. The document section alignment reduces visual inconsistency introduced by the recent Document page overhaul.

## 🔍 How

- New `LocateMeControl` Leaflet control component renders a locate button inside the map; it calls the browser Geolocation API and dispatches `setView` on success, or shows a translated snackbar on error.
- `MapMarkerSelector` wires in `LocateMeControl` and sets `fullscreenControl: true` on the Leaflet map instance.
- `Entry/Documents` and `Massif/Documents` adopt the same card/chip layout used by the new Document page.
- 4 new i18n keys added and translated across all 15 language files.

## 🔗 Related API PR

_N/A_

## 🧪 Testing

1. Open the entrance creation form and expand the map.
2. Click the locate button — the marker should jump to your GPS position.
3. Deny location in the browser — an error snackbar should appear.
4. Toggle full-screen on the map — the map should expand to fill the viewport.
5. Visit an entrance or massif page — the associated documents section should match the style of the Document page.

## 📸 Previews

_N/A_


ClemRz (approved):

- **[MapMarkerSelector.jsx:82]** `hasGeolocation` is evaluated at module load time (top-level `const`). This works fine in a browser environment, but if this module ever gets imported during SSR or testing where `navigator` is undefined, it would throw. Since the project is a CRA SPA this is probably fine, but wrapping it in a safer check (`typeof navigator !== 'undefined' && Boolean(navigator.geolocation)`) would make it more robust.

- **[MapMarkerSelector.jsx:108-113]** The one-time initialization effect depends on `validLatitude` and `validLongitude` which are derived from `boundMinMax(toFloat(...))`. If the form fields start empty (creation mode), `toFloat(undefined)` returns `NaN`, `boundMinMax(-90, 90, NaN)` returns `NaN`, and the `isNaN` guard prevents initialization — good. However, consider adding `initialized` to the dependency lint comment or verifying the ESLint exhaustive-deps rule is satisfied, since `setCurrentPosition` and `setZoomLevel` are stable but `initialized` gating the entire effect is a pattern that can confuse future readers. A brief comment like "runs once when valid coords appear" would help.

- **[MapMarkerSelector.jsx:119-137]** `handleLocateMe` doesn't pass `enableHighAccuracy: true` in the geolocation options. Since the stated use case is placing a cave entrance marker in the field, GPS-level accuracy matters. Consider adding `enableHighAccuracy: true` to the options object `{ timeout: 10000, maximumAge: 60000, enableHighAccuracy: true }` for a better field experience on mobile.

- **[LocateMeControl.jsx:27]** The error tooltip uses `open={error ? true : undefined}` which means it stays permanently visible until the user clicks locate again (which resets the error). This could be confusing UX-wise — consider auto-dismissing the error state after a few seconds (e.g., a `setTimeout` that clears `locateError` in the parent), or at least making the tooltip closeable on click elsewhere.

- **[LocateMeControl.jsx:14]** The `LocateMeControl` component accepts `loading` and `error` as optional props but doesn't provide default parameter values in the destructuring. Per the project conventions (React 19 — no `defaultProps`, use JS defaults), consider: `({ onClick, loading = false, error = null })`.

- **[entityConfig.jsx:50]** The `BadgeIcon` prop default (`AddCircleIcon`) is a good pattern, but since `EntityIcon` is now exported and used across multiple features (Entry Documents, Massif Documents, EntityCreation page), consider adding a brief JSDoc comment to document what `BadgeIcon` expects (an MUI SvgIcon component). This helps future contributors.

### Nitpicks (Optional)

- **[MapMarkerSelector.jsx:1]** `useRef` is still imported from React. It's used by the inner `MapBind` component so it's correct — just noting it in case anyone wonders during a quick scan.

- **[MapMarkerSelector.jsx:83]** `ACCURACY_CIRCLE_STYLE` uses a hardcoded color `#1976d2` which happens to match MUI's default primary color. If the theme changes, this won't follow. Consider using theme tokens if the circle should match the app's primary color, though for a map overlay this is a minor concern.

- **[.agents/skills/translate/SKILL.md]** This file is unrelated to the UX improvements described in the PR title. Consider splitting it into a separate PR for a cleaner git history, though it's not blocking.

- **[Entry/Documents/index.jsx, Massif/Documents.jsx]** The icon swap from `AddLinkIcon` to `EntityIcon` with `BadgeIcon={LinkIcon}` is a nice consistency improvement. Just noting that the "Associate" button now renders a more complex icon (CustomIcon + badge) which is slightly heavier than a simple MUI icon — negligible perf impact but worth being aware of if this pattern proliferates.

urien (approved):
Cela fonctionne parfaitement.
Merci @Paul-AUB , il me reste a le tester sur le terrain

PR #1385, opened 2026-06-10
Feat/1307 organization responsible geo entity UI

dawoldo:
# Organization and Geographic Entity Associations (UI)

## 🤔 What

- Implemented state management (Redux reducers and actions) and unified API endpoints for managing geo-associations.
- Created `AssociationSection` and `ManagedEntitiesSection` reusable components to list and manage associations on both sides of the relationship.
- Introduced `AssociationForm` to allow authenticated users to search, attach, or remove organizations from geographic entities.
- Added and updated localization keys for all new UI components.

## 🤷‍♂️ Why

Previously, organizations were limited to a single 1-to-1 relationship with geographic entities. The reality of cave management and speleological organizations requires a more flexible approach where an organization can manage multiple regions/massifs, and a geographic entity can be managed by multiple organizations. This change brings the frontend UI in line with the new many-to-many backend architecture, giving users the tools they need to correctly map real-world responsibilities.

## 🔍 How

- **State Management**: Refactored Redux reducers and actions to correctly handle lists of associated entities, supporting the asynchronous operations (set/remove) needed for many-to-many data structures.
- **API Integration**: Centralized API route configuration for all association-related endpoints, fetching lists of organizations instead of single records.
- **UI Architecture**:
  - `AssociationSection`: A shared, condition-aware component for geographic pages (Country, Region, Massif) to display assigned organizations.
  - `ManagedEntitiesSection`: A symmetrical component for Organization pages to display the geographic areas managed by that organization.
  - `AssociationForm`: Provides an interactive interface for authenticated users to manage associations, utilizing autocomplete for efficient organization searches with loading states and error handling.
- **Localization**: Synced all new translation keys across supported languages to ensure full internationalization coverage.

## 🔗 Related API PR

[- [Insert Related API PR Link Here]](https://github.com/GrottoCenter/grottocenter-api/pull/1644)

## 🧪 Testing

_Additional instructions needed to test the changes._

1. Log in as an authenticated user with appropriate permissions.
2. Navigate to a Geographic Entity page (e.g., a Country, Region, or Massif).
3. Verify that the new `AssociationSection` correctly lists any currently associated organizations.
4. Use the `AssociationForm` to search for and attach a new organization to the geographic entity.
5. Attempt to detach/remove an existing associated organization.
6. Navigate to the page of the newly associated Organization.
7. Verify that the `ManagedEntitiesSection` correctly reflects the geographic entity you just linked.
8. Switch the application language to verify localization of the new forms and sections.


ClemRz (changes requested):
### Issues (Must Fix)

- **[OrganizationAssociation/index.jsx] Missing `entityId` and `parentEntityId` props in all three calling components** — `AssociationSection` declares `entityId` as `.isRequired` in PropTypes and uses it in `handleSet`, `handleRemove`, and the `useEffect` that refreshes data. However, none of the three parent components (`Country/index.jsx`, `Region/index.jsx`, `Massif/Massif.jsx`) pass `entityId` or `parentEntityId`. This means all set/remove operations will call the API with `undefined` IDs, producing broken URLs (e.g., `/countries/undefined/organization`) and the data refresh will also fail silently.

  In `Country/index.jsx`:
  ```suggestion
  <AssociationSection
    organizations={country.organizations}
    entityType="country"
    entityId={country.id}
    isLoading={isLoading}
  />
  ```

  In `Region/index.jsx`:
  ```suggestion
  <AssociationSection
    organizations={region.organizations}
    entityType="region"
    entityId={regionId}
    parentEntityId={countryId}
    isLoading={isLoading}
  />
  ```

  In `Massif/Massif.jsx`:
  ```suggestion
  <AssociationSection
    organizations={massif.organizations}
    entityType="massif"
    entityId={massif.id}
    isLoading={isLoading}
  />
  ```

- **[Multiple files] Many translation keys used in components are missing from `en.json` and all language files** — Only 2 keys were added (`"An error occurred while saving."` and `"Managed entities"`), but the new components use at least 9 additional keys that don't exist in any language file:
  - `"Set organization"`
  - `"Search or create organization"`
  - `"View successor organization"`
  - `"Remove association"`
  - `"Add/Edit organization"`
  - `"No organization is currently associated with this country."`
  - `"No organization is currently associated with this region."`
  - `"Are you sure you want to remove the association with \"{name}\"?"`
  - `"No existing organization found. A new one named \"{name}\" will be created."`

  These will all render as raw key strings in the UI. Per project conventions, all user-facing text must have a corresponding entry in `en.json` (and ideally in all 15 language files), and the files must be sorted after modification.

- **[OrganizationAssociation/index.jsx:51] `eslint-disable` comment lacks written justification** — The project convention requires a justification comment above any `eslint-disable`. Consider adding an explanation of why `entityId`, `dispatch`, etc. are intentionally excluded from the dependency array (e.g., "These values are stable across renders and including them would cause unwanted refetches").

- **[OrganizationAssociation/index.jsx:33] Shared reducer state across concurrent operations** — All three entity types (country, region, massif) share a single `status` per type. If the user quickly performs a set and then a remove, or acts on different entities of the same type from different tabs, the `SUCCEEDED` status from one operation could trigger the refresh `useEffect` for a stale context. Consider adding an operation-specific identifier (e.g., the entity ID) in the reducer state and comparing it in the effect, or resetting the status to IDLE after handling success.

- **[AssociationForm.jsx:56–78] Search fires on every keystroke change after debounce, even when a selection was already made** — When the user selects an organization from the dropdown, `inputValue` gets updated to the selected org's name, which triggers another search via the `useEffect`. Consider skipping the fetch when `selectedOrg` is already set and its name matches `debouncedInput`.

- **[Country/index.jsx:185–189] Potential crash when `country` is null during loading** — `country.organizations` is accessed without optional chaining. If `isLoading` is true but `country` hasn't been fetched yet (initial load), this will throw. The Massif component correctly guards with `massif?.organizations`, but the Country and Region components access it unguarded inside a block that only checks `!isLoading && !error` conditionally.

- **[ManagedEntitiesSection.jsx:61–64] Region URL construction is fragile** — The logic `parts.length > 1 ? parts[0] : region.countryId` falls back to `region.countryId`, but there's no guarantee the API provides that field. If `region.id` doesn't contain a hyphen and `countryId` isn't present, the link will be `/ui/countries/undefined/regions/...`. Consider documenting the expected API shape or adding a guard.

- **[Action creators] `Content-Type` header is not explicitly set for JSON bodies** — The `authorizationHeader` from Redux only includes `Authorization`. While this follows the existing codebase pattern (and may work if the backend infers JSON), explicitly adding `'Content-Type': 'application/json'` would be more robust, especially for PUT/DELETE with bodies. This is a codebase-wide pattern issue, so not blocking, but worth noting.

### Nitpicks (Optional)

- **[lang/*.json] Non-English translation files use English fallback text** — Both new keys (`"An error occurred while saving."` and `"Managed entities"`) are added with English text in all 14 non-English files. This is acceptable as a placeholder, but consider running the AI translation pipeline before merge if available.

- **[OrganizationAssociation/index.jsx:82–93] Empty state JSX returns `false` for massif** — When `isEmpty` is true and `entityType === 'massif'`, the expression `entityType !== 'massif' && (...)` evaluates to `false`, which React doesn't render. This works but could be clearer with an explicit `null` return or by restructuring the conditional.

ClemRz (changes requested):
### Issues (Must Fix)

- **[All lang/*.json files] Translation keys corrupted by sort script** — The sort operation applied to language files has broken several translation keys that contain escaped quotes (`\"`). The sort script appears to truncate keys at the first `\"` boundary, producing nonsensical key fragments. For example:
  - `"The BBS (\"Bulletin Bibliographique Spéléologique\" in french) is..."` → now keyed as `" in french) is an annual review..."`
  - `"Are you sure you want to remove the association with \"{name}\"?"` → now keyed as `"?"`
  - `"No existing organization found. A new one named \"{name}\" will be created."` → now keyed as `" will be created."`
  - `"To be used for a cave requiring special protection..."` → now keyed as `", location of the entrance..."`
  - `"Try to go straight to the point..."` → now keyed as `"."`
  - `"Additional information about the pages..."` → now keyed as `")."`
  
  This breaks **both** the two new PR translation keys (they'll never match the `formatMessage({ id: ... })` calls in `AssociationForm.jsx` and `OrganizationAssociation/index.jsx`) **and** 3 previously-working translation keys that are now corrupted across all 15 language files. The original full-length keys have been deleted (visible as `-` lines in the diff).
  
  This likely happened during the merge conflict resolution + re-sort. The fix is to restore the correct full keys and re-run the sort with a script that properly handles escaped quotes.

- **[OrganizationAssociation/index.jsx] No role-based permission check for add/remove operations** — The component only checks `isAuth` (any logged-in user), but the Massif component uses `usePermissions().isModerator` for destructive operations. Given that managing geographic entity associations is an organizational responsibility, consider gating add/remove behind at least `isLeader` or checking against the backend's expected permission model. Even if the API rejects unauthorized requests, showing the UI controls to users who can't use them leads to a confusing UX.

- **[OrganizationAssociation/index.jsx:55] `eslint-disable` now has a justification comment — but the dependency list still has a subtle issue** — The effect depends on `status` alone, but `isPending` is also read inside the effect. If `isPending` becomes stale (e.g., component re-renders between dispatch and status change), the effect could miss a status transition. Consider adding `isPending` to the dependency array (it's a boolean that changes infrequently) or using a ref for the pending flag to avoid this category of bugs.

- **[ManagedEntitiesSection.jsx:57–66] Region URL construction is fragile** — The logic splits `region.id` on `-` and assumes the first part is a country code. If a region ID happens to have multiple hyphens (e.g., `GB-ENG-SW`), `parts[0]` would still work but `parts[1]` would be incomplete (`ENG` instead of `ENG-SW`). Consider using `split('-', 1)` / `indexOf` logic, or better yet, have the API return `countryId` as a separate field to avoid guessing.

- **[Massif/Massif.jsx:323] Missing optional chaining for `massif.organizations`** — While the section is inside a `{massif && ...}` guard, the `organizations` property itself could be `undefined` on the massif object. The condition uses `massif?.organizations?.length > 0` (safe), but the prop passes `organizations={massif.organizations}` without `?.`. Since `AssociationSection` defaults `organizations = []`, this works, but using `massif?.organizations` in the prop would be more defensive and consistent with the condition above it.

- **[Action creators: Country/Region/MassifOrganization.js] Error object not structured per project conventions** — The action creators dispatch `error` directly from the catch block (which is a native Error or whatever `checkAuthStatus` throws). Per the architecture steering file, the preferred pattern is to dispatch a structured error with `{ code, message, details, status }` extracted from `error.body`. This would allow the component to display localized error messages using the API error code as an i18n key.

### Nitpicks (Optional)

- **[All non-English lang/*.json] New PR-specific keys use English fallback text** — Keys like `"An error occurred while saving."`, `"Managed entities"`, `"Remove association"`, etc. have English text as placeholder in all 14 non-English files. Consider running the translation pipeline before merge if available, or filing a follow-up task.

- **[OrganizationAssociation/index.jsx:88–93] Empty state for massif returns `null` from a ternary** — When `isEmpty` and `entityType === 'massif'`, the ternary evaluates the inner `entityType === 'massif' ? null : (...)` path. This works but is slightly hard to read. A more explicit pattern would be a simple early `if` block before the main JSX.

ClemRz (changes requested):
### Issues (Must Fix)

- **[Country/index.jsx:183, Region/index.jsx:190, Massif/Massif.jsx:323] AssociationSection is never rendered when no organizations exist, preventing users from adding the first one** — The guard condition `(country?.organizations?.length > 0 || isLoading)` means the entire `ScrollableContent` (and thus the `AssociationSection` with its "Add/Edit organization" button) is hidden when the organizations array is empty or absent. This directly contradicts Requirement 7 AC1 and Requirement 9 AC2 which state that authenticated users should always see the set/change controls. An authenticated user visiting a country/region/massif with no associations has no way to add the first one.

  For Country, consider:
  ```suggestion
  {(country?.organizations?.length > 0 || isLoading || canManageAssociations) && (
  ```
  (Where `canManageAssociations` would need to be computed in the parent or the condition simplified to always render the section, letting `AssociationSection` handle its own visibility.)

  The simplest fix might be to always render the `ScrollableContent` when `country` is present:
  ```suggestion
  <ScrollableContent
    anchorId="organizations"
    title={formatMessage({ id: 'Organizations' })}
    content={
      <AssociationSection
        organizations={country?.organizations}
        entityType="country"
        entityId={country?.id}
        isLoading={isLoading}
      />
    }
  />
  ```
  The `AssociationSection` already handles the empty state internally (shows a message for country/region, returns null for massif). The `massif` case already avoids rendering empty text, so the section would just show the "Add" button to authorized users — which is the expected behavior.

- **[OrganizationAssociation/index.jsx:55] Shared reducer state creates a race condition across multiple entity instances** — Since `countryOrganization`, `regionOrganization`, and `massifOrganization` are single global reducer slices (no entity-specific scoping), if two Country pages are open in different tabs (or the user navigates quickly), a `SUCCEEDED` status from one operation will trigger the refresh effect for any mounted `AssociationSection` of the same entity type, potentially refreshing stale data or closing the wrong form. The `pendingOperationRef` mitigates this for the *current* component's lifecycle but doesn't handle: (1) the effect firing for a *different* entity ID when status transitions happen from another context, or (2) the reducer never being reset to IDLE after success, meaning navigating to a new entity of the same type will immediately trigger the success branch on mount if `pendingOperationRef` somehow gets set. Consider resetting the reducer status to IDLE after handling success, or storing the target `entityId` in the reducer to validate against.

- **[OrganizationAssociation/index.jsx:32] Permission check uses `isLeader || isModerator || isAdmin` but requirements say "any authenticated user"** — The issue requirements (Requirement 7 and 9) state: "Per the backend rules, any authenticated user can create, modify, or delete associations" and AC2 states "WHEN the current user is an Authenticated_User, THE Association_Section SHALL display set, change, and remove controls." The current implementation restricts add/remove to Leader/Moderator/Admin roles. If this is an intentional decision (tighter than the spec), consider adding a comment explaining why. Otherwise, `canManageAssociations` should just be `isAuth`.

- **[ManagedEntitiesSection.jsx:56-67] Region URL construction could produce incorrect `regionId` for composite ISO codes** — For region IDs like `FR-ARA` the logic works fine (`countryId = 'FR'`, `regionId = 'ARA'`). But the original region entity was likely fetched with `region.id = 'FR-ARA'` as a full ISO 3166-2 code. The constructed URL becomes `/ui/countries/FR/regions/ARA`, which requires the routing to handle that correctly. If some routes expect the full ISO code instead of just the suffix, this would produce broken links. Consider documenting this assumption or validating against actual route patterns.

- **[AssociationForm.jsx:47-50] Debounced search comparison relies on exact name match** — The skip condition `selectedOrg && selectedOrg.name === debouncedInput.trim()` avoids re-fetching when the user selects an option. However, if the user manually types text that happens to match the selected org's name (e.g., edits and retypes), the search won't fire even though the user likely wants fresh results. This is a minor UX edge case but worth noting.

- **[All 3 action creators] No reset/IDLE action defined** — There's no way for the component to reset the reducer back to IDLE state (e.g., when unmounting or before starting a new operation). This means stale `SUCCEEDED`/`FAILED` states persist in the store. Consider adding a `RESET_*_ORGANIZATION` action that returns to `initialState`, dispatched on component unmount or at the start of a new operation cycle.

### Nitpicks (Optional)

- **[All non-English lang/*.json] Two keys remain with English fallback text in all 14 non-English files** — `"Are you sure you want to remove the association with \"{name}\"?"` and `"No existing organization found. A new one named \"{name}\" will be created."` have English text as placeholder in all non-English files. The other 9 new keys are properly translated. Consider running the translation pipeline for these two before merge.

- **[scripts/translations/update-en.js] `isValidTranslationKey` now allows parentheses but blocks `{` characters** — This means keys containing ICU `{name}` placeholders (like the two new keys in this PR) would fail validation and never be auto-added by the script. The script likely relies on these keys being manually added (as was done here), but it's worth confirming this is intentional behavior — the `{` exclusion might be too aggressive if future keys use simple string interpolation syntax.

- **[OrganizationAssociation/index.jsx:92] Empty-state ternary for massif is functional but hard to read** — The nested ternary `entityType === 'massif' ? null : (...)` inside the `isEmpty` branch works but could be clearer with an early return or a separate `if` statement.

ClemRz (changes requested):
### Issues (Must Fix)

- **[actions/Massif/MassifOrganization.js] `resetMassifOrganization` is imported but never defined or exported** — `OrganizationAssociation/index.jsx` imports `resetMassifOrganization` from this file (line 16), and calls it on success (line 55) and on unmount (line 72). However, `MassifOrganization.js` does not export a `RESET_MASSIF_ORGANIZATION` constant or a `resetMassifOrganization` action creator. Additionally, `MassifOrganizationReducer.js` has no `case` for a reset action type (unlike the Country and Region reducers which both handle their respective `RESET_*` actions). This means `resetMassifOrganization` will be `undefined` at runtime, and `dispatch(undefined)` will throw or silently fail, leaving stale `SUCCEEDED` state in the massif organization reducer after every operation. The fix is to add the same pattern used in `CountryOrganization.js` and `RegionOrganization.js`:

  In `MassifOrganization.js`, add:
  ```js
  export const RESET_MASSIF_ORGANIZATION = 'RESET_MASSIF_ORGANIZATION';
  export const resetMassifOrganization = () => ({ type: RESET_MASSIF_ORGANIZATION });
  ```

  In `MassifOrganizationReducer.js`, import `RESET_MASSIF_ORGANIZATION` and add:
  ```js
  case RESET_MASSIF_ORGANIZATION:
    return initialState;
  ```

- **[OrganizationAssociation/index.jsx:34] `reducerKey` dynamic string relies on naming convention** — The component builds the Redux selector key as `` `${entityType}Organization` `` (e.g., `"countryOrganization"`). This works because the keys in `GCReducer` happen to match, but it creates an implicit coupling that's easy to break during refactoring. Consider a lookup object (`{ country: 'countryOrganization', region: 'regionOrganization', massif: 'massifOrganization' }`) or at minimum a comment documenting the expected reducer keys.

- **[AssociationForm.jsx:95–99] Submitting with `inputValue.trim()` alone (no `id`) creates an org inline** — When no `selectedOrg` exists but `inputValue` has text, `handleSubmit` calls `onSubmit({ name: inputValue.trim() })`. This triggers `setCountryOrganization(entityId, undefined, orgName)`, which sends `{ name: "..." }` to the PUT endpoint. This implements inline org creation (Requirement 7 AC3), but there's no explicit 200-char limit on the input field as specified in AC3. Consider adding `inputProps={{ maxLength: 200 }}` to the `TextField`.

- **[ManagedEntitiesSection.jsx:57–66] Region URL parsing has a graceful fallback but the UX could be confusing** — When `countryId` can't be determined from the region ID and `region.countryId` is absent, the link falls back to `/ui/countries` (the country list). This prevents a broken link, which is good. Consider logging a warning in development so this unexpected data shape is noticed during testing rather than silently degrading.

### Nitpicks (Optional)

- **[OrganizationAssociation/index.jsx:44] `isPending` state and `pendingOperationRef` ref seem redundant** — Both track whether an operation is in progress. `isPending` is used for UI rendering (disable buttons) and `pendingOperationRef` is used inside the effect to distinguish "our" status transitions. This pattern is functional and works, but adding a short comment explaining why both are needed (React state for re-render, ref for effect guard) would help future readers.

- **[Multiple action creators] `checkAuthStatus` is used without `checkAndGetStatus`** — The Country/Region/Massif action creators use `checkAuthStatus(dispatch)` which is correct for authenticated endpoints per the architecture docs. Just noting this follows conventions.

ClemRz (approved):

- **[OrganizationAssociation/index.jsx:5–6] Import order violation** — The internal hook import (`usePermissions` from `../../../hooks`) appears before the external library import (`react-router-dom`). Per project conventions, external libraries should come first, then internal imports. Consider moving line 6 (`react-router-dom`) above line 5:
  ```suggestion
  import { Link as RouterLink } from 'react-router-dom';
  import { usePermissions } from '../../../hooks';
  ```

- **[AssociationForm.jsx:47–50] Debounced search may fire redundantly after selection** — When a user selects an organization, the `Autocomplete`'s `onInputChange` fires with the selected name, which updates `inputValue`, which eventually triggers the search effect via `debouncedInput`. The guard `selectedOrg && selectedOrg.name === debouncedInput.trim()` catches this *most* of the time, but there's a race window: `setSelectedOrg(newValue)` and `setInputValue(newInputValue)` are set in separate handlers (onChange vs onInputChange), so the `selectedOrg` check in the effect might evaluate against the stale `null` before the next render cycle updates it. In practice the cleanup function (`active = false`) and the debounce delay prevent visible issues, but consider setting `selectedOrg` inside the `onInputChange` handler when `reason === 'reset'` to tighten the guard.

- **[Massif/Massif.jsx:326] `massif.id` used without optional chaining inside `{massif && ...}` guard** — While `massif` is guarded by the outer conditional (`{massif && (...)}`), the prop `entityId={massif.id}` will work fine. However, one line above, `organizations={massif?.organizations}` uses optional chaining inconsistently with the non-optional `massif.id`. Consider `entityId={massif?.id}` for stylistic consistency, or drop the `?.` from `massif?.organizations` since `massif` is guaranteed truthy here.

### Nitpicks (Optional)

- **[CRSMenu/index.jsx:21–28] Unrelated change bundled in the PR** — The `defineMessages` addition for `'Decimal degrees (WGS84)'` and `'Degrees Minutes Seconds'` is a translation-infrastructure improvement unrelated to the organization association feature. It also introduces a double blank line after the `defineMessages` block. Not blocking, but flagging for clean git history awareness.

- **[CRSMenu/index.jsx:29] Double blank line** — There are two consecutive blank lines between the `defineMessages` block and the `CRSMenu` component definition. The project Prettier config typically enforces single blank lines between declarations.

- **[ManagedEntitiesSection.jsx:29] `py: 0.5` uses a non-integer spacing value** — Per project conventions, `theme.spacing()` values should be integers. The `sx={{ px: 0, py: 0.5 }}` on `ListItem` passes `0.5` which translates to `4px` via MUI's spacing system. While this works since MUI accepts fractional values in `sx`, the convention states integer-only spacing. Consider `py: 0` (too tight) or `py: 1` (standard), or use a raw pixel value `py: '4px'` if 4px is specifically needed.

- **[scripts/translations/update-en.js] `isValidTranslationKey` now allows parentheses** — The original code blocked keys containing `(` or `)`. The new version removes this check, which means keys like `"Degrees Minutes Seconds"` can now be extracted even if a key ever contained parentheses. This is intentional per the `defineMessages` fix, but noting for awareness that any future key with parentheses (e.g., from dynamic expressions) would no longer be filtered.

urien (changes requested):
Le problème persiste. La validation ne produit aucun effet

Image

urien (approved):
Tout fonctionne bien
Relation de 2 organisations a un pays
Relation d'une organisation a un département
Relation avec une organisation inexistante dans la base
Parfait @dawoldo

PR #1386, opened 2026-06-10
feat(snapshot): refactor revision history pages

Paul-AUB:
# feat(snapshot): refactor revision history pages

## 🤔 What

Complete refactor of the revision history pages (`/ui/:type/:id/snapshots`) and the History inline section in the Entry page.

## 🤷‍♂️ Why

The previous implementation had several critical issues:
- Current entity was read from `localStorage`, breaking direct navigation via shared links
- No `PageContainer`/`PageHeader` — the page had no title, no back button, and no chrome
- The all-snapshots page always tagged every item as "Initial version" (root cause: mishandled flattened data structure from `sortSnapshots`)
- Current versions were always pinned to the top regardless of their actual date, breaking chronological order
- The History inline section had a stale `useEffect` dep causing a stuck spinner after delete/restore
- Several PropTypes were wrong or missing, and a null-safety bug in `RestoreSnapshot` could throw a `TypeError`

## 🔍 How

**Snapshot pages:**
- Replace `localStorage` hack with Redux selectors for all entity types (`entrances`, `caves`, `documents`, `massifs`, `persons`, `organizations`) and sub-entities (fetched via `parentType` param)
- Add `PageContainer` / `PageHeader` with translated back button and proper title
- Rewrite `AccordionSnapshotListPage` with a unified chronological timeline: current versions and past snapshots are merged and sorted by date instead of being rendered in two separate groups
- Fix `previousMap` computation to correctly handle the flattened structure from `sortSnapshots` (each element is `{ [type]: [singleSnapshot] }`)
- Add sub-entity support on the all-snapshots page for descriptions, locations, histories, riggings, comments
- Rewrite `DocumentSnapshots` with a `Property` grid and `HighLightsLine` diffs for all fields
- Add green/red highlight backgrounds for added/removed boolean characteristics in `EntranceCaveSnapshots`
- Fix rigging table to highlight purely-added rows in green (`oldData={isAdded ? '' : oldRow?.field}`)

**History section:**
- Fix `useEffect` missing dep (`history.isDeleted`) — removes eslint-disable
- Replace `float: right` layout with flexbox
- Replace IIFE in JSX with `useMemo`

**Code quality (from code review):**
- Fix `RiggingTable` PropTypes: `previous` was `arrayOf` instead of `shape({ obstacles: arrayOf(...) })`
- Fix `RestoreSnapshot`: add missing PropTypes, fix `content.cave?.id` null-safety
- Fix `EntranceCaveSnapshots` PropTypes: `information: PropTypes.node` → `entrance` + `previous`
- Fix `DocumentSnapshots` PropTypes: add missing `dateInscription`, `dateValidation`, `validatorComment` in `previous` shape; apply `show()` helper consistently
- Fix `AccordionSnapshot` PropTypes: add `description.title` to snapshot shape
- Fix `sortSnapshots` date fallback: `0` → `new Date(0)`
- Fix typo: `HighligtedTableCell` → `HighlightedTableCell`

**i18n:** Add and translate 9 new keys in 14 languages.

## 🔗 Related API PR

N/A

## 🧪 Testing

- Navigate directly to `/ui/entrances/:id/snapshots` — current version loads via Redux (no click needed from Entry page)
- Navigate to `/ui/entrances/:id/snapshots?all=true` — timeline is in chronological order with current versions interleaved at their actual date
- Navigate to `/ui/histories/:id/snapshots?parentId=:entranceId&parentType=entrances` — sub-entity history works
- Restore a revision on an entrance that has a cave — no TypeError
- Delete a History item in Entry page — spinner does not get stuck

## 📸 Previews

image

image

image


ClemRz (changes requested):
### Issues (Must Fix)

- **[UtilityFunction.js + RestoreSnapshot.jsx]** The `SnapshotButton` now uses `rel="noopener noreferrer"` (previously `rel="opener"`), which means pages opened via this button will have `window.opener === null`. However, `RestoreSnapshot.jsx:135` still calls `window.opener.location.reload()`, which will throw a `TypeError: Cannot read properties of null`. This is a runtime crash that breaks the restore flow entirely. Consider reverting to `rel="opener"` (since these are same-origin pages on the same domain, `noopener` isn't strictly needed) or replacing the `window.opener.location.reload()` mechanism with an alternative approach (e.g., dispatching a re-fetch action after restore).

- **[AccordionSnapshotListPage.jsx:25-51]** The `previousMap` and `latestByGroup` computation uses an IIFE that runs on every render. Since `data` can be a large array of snapshots, consider wrapping this in `useMemo` keyed on `data` to avoid unnecessary recalculations during re-renders.

- **[AccordionSnapshotListPage.jsx:54-115]** Similarly, `timelineItems` is computed via another IIFE on every render and depends on `hasData`, `currentTItem`, `isCurrentItemLoading`, `latestByGroup`, etc. A `useMemo` here would prevent re-creating all these React elements on unrelated state changes.

- **[Translation files]** The key `"No previous revisions."` is added to all 15 language files but appears unused in the codebase (no JSX or JS file references it). Consider removing it to keep the translation files clean, or if it's planned for future use, note that it's left untranslated (English fallback) in all non-English locales.

- **[Translation files]** Two keys (`"This is the only version of this {type}."` and `"No previous revisions."`) are left untranslated in all non-English locales (values are the English string). While the PR body mentions translations for 9 new keys in 14 languages, these two appear to have been missed.

- **[AccordionSnapshot.jsx:53]** `snapshot.id` is typed as `PropTypes.string` for past snapshots (it's a timestamp), but the `previous` prop types it as `PropTypes.number`. This inconsistency might cause confusion. Consider aligning the types or adding a comment explaining the difference.

- **[Snapshots/index.jsx:63]** The `SUB_ENTITY_TYPES` array is re-created on every render since it's declared inside the component body. Consider moving it outside the component (like `AccordionSnapshotListPage` does with `SUB_TYPES`).

### Nitpicks (Optional)

- **[EntranceCaveSnapshots.jsx:10]** The component still uses the `information` props-as-arguments pattern (`const EntranceCaveSnapshots = information => { const { entrance, previous } = information; ...}`). Consider destructuring directly in the parameter list for consistency with the rest of the codebase: `const EntranceCaveSnapshots = ({ entrance, previous }) => { ... }`.

- **[Highlights/index.jsx]** The change from `diffChars` to `diffWords` in `HighLightsChar` is a semantic behavior change (word-level diffing vs character-level). The component name suggests character-level diffing. Consider renaming to `HighLightsWord` or adding a comment explaining the intent, to avoid future confusion.

- **[Entry/index.jsx]** The only change here is reformatting the `formatMessage` call to one line. This is unrelated to the PR's purpose — minor but worth noting for clean git history.

ClemRz (approved):

urien (approved):
C'est bien plus lisible et utilisable
J'ai visualiser l'ensemble des historiques sur différentes cavités et cela fonctionne
J'ai fait une restauration sur une version antérieure et cela fonctionne
Merci @Paul-AUB

PR #1389, opened 2026-06-11
feat(map): allow to search for coordinates in WGS84, degminsec, Lambert 93 and Lambert III carto

Paul-AUB:
## 🤔 What

Allow users to paste coordinate strings directly into the map's search bar. The map centers on the detected point and shows a synthetic result in the dropdown — no API call is made.

Supported formats:
- WGS84 decimal: `45.1179, 5.4786`
- Decimal with cardinal/degree symbol: `45.1179° N, 5.4786° E`
- DMS: `45° 7' 4.6" N, 5° 28' 43" E`
- DDM: `N 45° 07.066 E 005° 28.717`
- Lambert 93 (EPSG:2154): `843000, 6450000` (X/Y order or inverted)
- Lambert III carto (EPSG:27573): `722000, 3165000` (X/Y order or inverted)

## 🤷‍♂️ Why

Many speleological workflows involve copy-pasting coordinates from topo reports, GPS exports, or club exchanges. Previously the search bar only handled place names — users had to manually enter coordinates into a separate form.

## 🔍 How

- Added `coordinate-parser` (npm) for WGS84/DMS/DDM parsing.
- Added `parseCoordinateString(input, projections)` to `coordinateConvert.js`:
  - Heuristic range detection for Lambert 93 and Lambert III carto (numeric ranges are unambiguous and never overlap WGS84 decimal values) — checked **before** `coordinate-parser` to prevent false-positive DMS interpretation of large integers.
  - A `PROJECTED_CRS_RANGES` config array drives detection; adding a new projected CRS requires a single entry in that array.
  - Falls back to `coordinate-parser` for WGS84/DMS/DDM.
  - Uses `projections[]` from the store (same shape as `formatCoordinatesForCopy`) — no hardcoded projection definitions.
- In `GeocodingControl`: coordinate detection short-circuits before any API call. A synthetic result with `resultType: 'coordinates'` is prepended to the dropdown; selecting it calls `map.setView` with no popup.
- 15 unit tests using real proj4 (no mocks).

## 🔗 Related API PR

None.

## 🧪 Testing

1. Open the map.
2. Type or paste any of the following into the search bar:
   - `45.1179, 5.4786` → WGS84, centers near Grenoble
   - `45° 7' 4.6" N, 5° 28' 43" E` → DMS, same point
   - `843000, 6450000` → Lambert 93 (requires Lambert 93 projection in the store)
   - `722000, 3165000` → Lambert III carto (requires EPSG:27573 in the store)
3. Verify the dropdown shows a single entry with the `MyLocation` icon and the detected format as subtitle.
4. Select it — the map should center on the correct location at zoom 16.
5. Verify no network request is fired when a coordinate string is detected.

## 📸 Previews

_Screenshots to be added._


urien (commented):
C'est super @Paul-AUB 
Est ce que tu pourais mettre un anneau au niveau du point > O. cela permettrait de matérialiser le centre de la carte et de voir s'il y a des entrées à proximité

urien (approved):
C'est parfait @Paul-AUB je le présente a la mission prospection lors de la réunion de ce soir

ClemRz (approved):

- **[coordinateConvert.js:190]** The `hasDMS` regex `/[°'"dms]|[NSEWnsew]/i` — the `i` flag makes `[dms]` also match uppercase `D`, `M`, `S`, so a coordinate string containing e.g. a trailing unit like `"45.1179m, 5.4786m"` would be misclassified as DMS when it's really decimal. This is an unlikely real-world scenario for this search bar, but consider anchoring the heuristic to more explicit DMS patterns (e.g. `°` paired with `'` or `"`, or cardinal letters at word boundaries) if precision matters here.

- **[GeocodingControl.jsx:384–396]** The 150ms `setTimeout` before `map.setView` is a pattern from the existing code for flushing React state, but the marker is added inside that timeout. If the user selects a coordinate result, immediately selects another one (within 150ms), `markerCleanupRef.current?.()` fires before the first timeout runs — meaning the first marker's creation callback will still execute and overwrite `markerCleanupRef.current`. Consider clearing the timeout handle too, or guarding with a flag. A possible approach:

  ```javascript
  markerCleanupRef.current?.();
  const handle = setTimeout(() => {
    // ... marker creation
  }, 150);
  markerCleanupRef.current = () => {
    clearTimeout(handle);
  };
  ```

  Then inside the timeout, reassign `markerCleanupRef.current` to the marker-removal function as you already do.

- **[useProjections.js:20 / GeocodingControl.jsx:376]** `useProjections()` returns `rawProjections ?? []` — when projections haven't loaded yet, this creates a new `[]` on every render, causing the `useEffect` (whose deps include `projections`) to re-execute on each render cycle until projections arrive. In practice, the effect early-exits when `query` is empty so it's not a bug, but it's wasteful. Consider memoizing the empty-array fallback in `useProjections` (e.g. a module-level `const EMPTY = []`) or using `useMemo`/`useRef` for stability. This is a pre-existing concern in the hook, not introduced by this PR, so no blocker — just worth knowing.

### Nitpicks (Optional)

- **[coordinateConvert.js:185]** The regex `/-?\d[\d.,]*/g` will match strings like `1,2,3` as a single token (e.g. `"1,234"` for thousands-separated input). Since the function then does `.replace(',', '.')`, only the first comma is replaced. This works fine for the intended use-case (decimal comma `45,1179`) but might produce unexpected floats for thousand-separated numbers. A comment clarifying that thousands separators are not supported would help future readers.

- **[coordinateConvert.test.js]** The test coverage is solid. Consider adding a test for negative latitude/longitude (southern/western hemisphere in decimal, e.g. `-33.8688, 151.2093`) to ensure `coordinate-parser` handles the sign correctly. Also a test with a comma as decimal separator (e.g. `45,1179 5,4786`) since that's common in French locale input.

- **[GeocodingControl.jsx:465–479]** The `allResults` memo includes `coordinateResult` in its dependency array and rebuilds the full array whenever it changes. Since `coordinateResult` changes on every keystroke (either to a parsed value or `null`), this might cause more rebuilds than strictly needed when the user is typing a normal place name. Negligible performance impact for an array of ~20 items, but noting for awareness.

PR #1390, opened 2026-06-11
fix(massif): filter entrances by massif ID instead of name

Paul-AUB:
# fix(massif): filter entrances by massif ID instead of name

## 🤔 What

- On massif pages, the entrances list is now filtered by `massifs.id` (integer) instead of `massifs.name` (tokenized string)
- Removed the pre-flight `fetchFieldSearch` call that was used to resolve the massif name to its exact Typesense facet value
- The active filter chip still displays the human-readable massif name (via `valueLabels`) rather than the raw ID

## 🤷‍♂️ Why

Filtering by name was fragile: massifs sharing common words (e.g. several "causse de…" massifs) could produce ambiguous text matches, causing the wrong entrances to be displayed. The massif ID is already known from the URL params and is unambiguous.

Fixes #1387.

## 🔍 How

**`EntrancesList/index.jsx`**
- Replaced `initialFilter: { 'massifs.name': resolvedMassif }` with `initialFilter: { 'massifs.id': parseInt(massifId, 10) }`
- Dropped the async `fetchFieldSearch` effect and its `massifValue` state (one fewer network round-trip)
- Passed `valueLabels: { 'massifs.id': massif.name }` so the locked filter chip shows the massif name instead of its numeric ID — same pattern already used for the `iso3166` → region name display

**`EntrancesSearch.jsx`**
- Registered `'massifs.id'` in `FILTER_LABELS` and `initialFilterState`
- The massif name autocomplete is now hidden when `massifs.id` is locked (in addition to when `massifs.name` is locked)

`massifs.id` is indexed as `int32[], facet: true` in the Typesense entrances schema, so no backend change is required.

## 🔗 Related API PR

- Context: GrottoCenter/grottocenter-api#1646
- Backend fix: GrottoCenter/grottocenter-api#1651

## 🧪 Testing

1. Navigate to any massif page that has entrances (e.g. a massif with a name shared with other massifs like "causse d'…")
2. Verify the entrances list shows only entrances belonging to that specific massif
3. Verify the "Massif: [name]" chip appears locked in the search form

## 📸 Previews

N/A — no visual change (the filter chip already displayed the massif name before).

### DIFF
--- a/packages/web-app/src/components/appli/AdvancedSearch/EntrancesSearch.jsx
+++ b/packages/web-app/src/components/appli/AdvancedSearch/EntrancesSearch.jsx
@@ -54,6 +54,7 @@ const FILTER_LABELS = {
   // iso3166 is the ISO 3166-2 subdivision code used as a locked filter when navigating from a
   // country/region page. It maps to the same "Region" label as the freeform `region` field.
   iso3166: 'Region',
+  'massifs.id': 'Massif',
   'massifs.name': 'Massif',
   region: 'Region',
   'cave.name': 'Network name',
@@ -75,6 +76,7 @@ const initialFilterState = {
   iso3166: '',
   region: '',
   county: '',
+  'massifs.id': null,
   'massifs.name': '',
   city: '',
   // postalCode is intentionally absent: the entrances API endpoint does not support postal code filtering
@@ -167,7 +169,7 @@ const EntrancesSearch = ({ initialFilter = {}, lockedFilter = [], valueLabels = 
               value={filterState.country}
             />
           )}
-          {!lockedFilter.includes('massifs.name') && (
+          {!lockedFilter.includes('massifs.name') && !lockedFilter.includes('massifs.id') && (
             <SearchTextAutocomplete
               ressourceType={searchEntity}
               ressourceField="massifs.name"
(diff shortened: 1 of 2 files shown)


urien (approved):
Cela fonctionne parfaitement.
Merrci @Paul-AUB

ClemRz (approved):
Clean, well-scoped fix. Switching from text-based massif name matching to the unambiguous integer ID is the right call — eliminates the fragile tokenized-string matching and removes a network round-trip in the process.

The implementation correctly mirrors the existing `iso3166` pattern (locked filter + `valueLabels` for human-readable chip display). Verified that:

- `parseInt(massifId, 10)` is safe here since the branch is guarded by `massif` being truthy (which requires a successful fetch)
- `fetchFieldSearch` is still used in `SearchElements.jsx`, so only the now-unused import was removed
- `ADVANCED_SEARCH_TYPES` had no other reference in the file
- `'massifs.id'` correctly excluded from `filterableKeys` (same pattern as `iso3166` — locked-only filters don't need to be in that list)
- The `ActiveFilterChips` component iterates `filterState` directly, so the locked chip renders as expected with the human-readable name via `valueLabels`

- **[EntrancesSearch.jsx:79]** The initial value for `'massifs.id'` in `initialFilterState` is `null`, while string-based filters use `''`. This is consistent with how other non-string filters (sliders, booleans) are handled, so it works correctly — just noting this is intentional since `ActiveFilterChips` checks `value === null` to skip rendering the chip when no filter is active.

### Nitpicks (Optional)

- **[EntrancesList/index.jsx:55]** Minor: `parseInt(massifId, 10)` could theoretically produce `NaN` if someone manually navigates to `/ui/massifs/abc/entrances`. The existing guard (`massif` must be truthy) makes this unreachable in practice since the API fetch would fail. Not blocking, but a comment noting the guard's role might help future readers.

PR #1393, opened 2026-06-11
feat(importObservations): add scientific observations CSV import wizard

ClemRz:
# Import scientific observations from CSV

## 🤔 What

Add a multi-step wizard that lets authenticated users import scientific observations from CSV files produced by data loggers.

- 6-step wizard: Upload → Device & Sensors → Column Mapping → Validation → Context → Submit
- Device search/creation with sensor configuration management
- Flexible timestamp parsing (datetime, date+time, year/month/day components, elapsed seconds)
- Timezone-aware UTC conversion
- Profile export/import for repeatable imports
- Integration into Entry and Network Science tabs
- New route `/ui/observations/import`

## 🤷‍♂️ Why

GrottoCenter needs to support importing environmental monitoring data (temperature, CO2, humidity, etc.) collected by instruments placed in caves. This enables researchers to store observation time-series alongside cave metadata.

## 🔍 How

- **Utilities layer**: CSV parser, encoding detector, number normalizer, timestamp builder, sampling interval detector, profile manager — all with unit + property-based tests
- **Redux layer**: Full action/reducer set for wizard state management (upload, device, column mapping, validation, context, submission)
- **UI layer**: One component per wizard step, orchestrated by a stepper shell with navigation guards
- **Integration**: Science tab added to Entry/Network pages with an "Import observations" button that deep-links to the wizard with a locked `caveId`
- **Cypress infrastructure refactored**: Shared commands (`loginAs`, `visitAuthenticated`, `checkPageLoaded`, `mockApiCatchAll`), fixture files, and mock helpers for reuse across specs

## 🔗 Related API PR

- https://github.com/GrottoCenter/grottocenter-api/pull/1656

## 🧪 Testing

- Unit tests: `yarn test` (all wizard utils, steps, and reducer)
- E2E: `yarn e2e:run` (requires dev server running — tests the full wizard flow with mocked API)
- Property-based tests cover timestamp parsing, number normalization, row resolution, sampling detection, and profile management

## 📸 Previews

N/A — wizard is behind authentication and admin-only Science tab.


Paul-AUB (approved):
This is an impressively well-structured PR: clean Redux architecture, extensive test coverage (unit, property-based, E2E), full i18n for 15 languages, and good navigation guards throughout. The utility layer is solid and the Cypress refactor is a clear improvement. Notes below are all non-blocking.

1. **[packages/web-app/src/components/appli/ImportObservationsWizard/index.jsx:63-75]** `isStep4NextDisabled` validates `licenseId`, `pointLabel`, and `caveId` but not `authorIds`. If a user deliberately removes the pre-filled author from the AuthorsSelect, they can reach SubmitStep with `context.authorIds = []` and only see a server-side rejection after uploading. Consider adding `if (!context.authorIds || context.authorIds.length === 0) return true;` — the API presumably requires at least one author for document creation.

2. **[packages/web-app/src/components/appli/Entry/index.jsx:465]** `<Science caveId={entrance?.cave?.id} />` passes `undefined` when the API returns an entrance without cave data — the prop is declared `isRequired`. Since this is inside `isAdmin &&` and renders inside a loaded entry, it's unlikely to happen in practice, but a defensive guard (`entrance?.cave?.id &&`) or changing the PropType to allow `undefined` would avoid PropTypes warnings. The Network component correctly passes the string `caveId` from `useParams()` — no issue there.

3. **[packages/web-app/src/components/appli/ImportObservationsWizard/utils/profileManager.js:183]** `deriveProfileFileName` uses `/[^A-Za-z0-9\-_]/g` which strips all accented characters (é, à, ç, É…). Given the app's international and French-first user base, a point named "Salle des Échos" becomes `SalledesEchos_profile.json` with the `É` dropped entirely, which can be surprising. Consider `/[^\w\-]/g` (which preserves Unicode word chars) or a transliteration step for common accented Latin chars.

4. **[packages/web-app/src/components/appli/ImportObservationsWizard/actions/Substance.js:7]** The `_dispatch` parameter in `searchSubstances` and `createSubstance` has an underscore prefix suggesting it is unused, but it is actually passed to `checkAuthStatus(_dispatch)`. Renaming it to `dispatch` would avoid misleading readers.

### Nitpicks (Optional)

5. **[packages/web-app/cypress/support/commands.js:15-70]** `loginAs` and `visitAuthenticated` both contain the full JWT payload construction inline. `visitAuthenticated` could call `cy.loginAs(authOptions)` (which stores the token in localStorage) and then `cy.visit(url)` instead of duplicating the token-building logic.

6. **[packages/web-app/src/components/appli/EntitiesForm/utils/MapMarkerSelector.jsx:81-93]** `AdditionalMarkers.propTypes` declares `positions: PropTypes.arrayOf(PropTypes.shape({}))` without `.isRequired`. The component is only rendered when `additionalPositions.length > 0`, so `positions` can never actually be undefined here — marking it `isRequired` would match the actual invariant and make the interface self-documenting.

7. **[packages/web-app/src/components/appli/ImportObservationsWizard/constants/quantityKinds.js, units.js, quantityKindUnitsMap.js]** The `TODO: Consider fetching these from an API endpoint to avoid drift` comments are appropriate given that these mirror the API's `t_quantity_kind` and `t_unit` tables, and the user testing already surfaced missing measurements (Al, NO3, Rn, etc.). The follow-up in GrottoCenter/grottocenter-api#1659 already tracks this — the inline TODOs are a useful breadcrumb until it lands.

Paul-AUB (approved):
Re-review of the changes since my previous approval (commits `78e87a8` and `5cf72b7`). All 7 items from the prior review are correctly addressed, and the new "mandatory coordinates in `pointOnly` mode" feature is clean and well-tested. Verified:

- ✅ `isStep4NextDisabled` now blocks on empty `authorIds` (and on missing coordinates in `pointOnly`)
- ✅ `<Science />` guarded by `entrance?.cave?.id` truthiness in `Entry/index.jsx`
- ✅ `deriveProfileFileName` now NFD-normalizes + strips diacritics before sanitizing
- ✅ `Substance.js` `_dispatch` → `dispatch`
- ✅ `visitAuthenticated` now delegates to `cy.loginAs()`
- ✅ `AdditionalMarkers.positions` marked `.isRequired`
- ✅ `CoordinateFormSection` correctly consumes the `required` prop

One non-blocking observation below.

1. **[packages/web-app/src/components/appli/ImportObservationsWizard/steps/ContextStep.jsx:417-418]** Profile import can leave `pointOnly` coordinates hidden and uneditable. `unknownCoordinates` now defaults to `true` in `initialState`, but `exportProfile`/`importProfile` never persist it (profileManager.js:55-85, 146-159), and `SET_CONTEXT` merges via `{ ...state.context, ...action.context }` (ImportWizardReducer.js:196). So importing a `pointOnly` profile yields `locationMode: 'pointOnly'` + `unknownCoordinates: true` (retained default). Then `showCoordinates = showPoint && !context.unknownCoordinates` is `false`, hiding the coordinate form, while `coordinatesRequired` hides the "unknown coordinates" checkbox too — the user can neither see nor edit the (mandatory) coordinates. Data still submits because the lat/long from the profile are applied and `isStep4NextDisabled` passes, but the UX is confusing on a round-trip. Consider `const showCoordinates = showPoint && (coordinatesRequired || !context.unknownCoordinates);` so `pointOnly` always shows the form regardless of the stale `unknownCoordinates` flag. (Note: switching mode via the radio already sets `unknownCoordinates: false`, so only the profile-import path is affected — the existing "always show in pointOnly" test passes `unknownCoordinates: false` explicitly and doesn't cover this case.)

PR #1394, opened 2026-06-13
fix(notifications): handle missing notificationType verbs in formatMessage

ClemRz:
# fix(notifications): handle missing notificationType verbs in formatMessage

## 🤔 What

- Fix crash in `NotificationMenuItem` when `formatMessage` is called with an empty `id`
- Add support for `PERMANENT_DELETE` and `RESTORE` notification types
- Add defensive fallback for any future unknown notification types

## 🤷‍♂️ Why

The API returns `notificationType.name = "PERMANENT_DELETE"` (and can return `"RESTORE"`), which the frontend switch statement didn't handle. This left `verb` as an empty string, and `formatMessage({ id: '' })` throws:

```
Error: [@formatjs/intl] An `id` must be provided to format a message.
```

Introduced in `12cf35dd` when the translation normalization refactored verb interpolation into a separate `formatMessage` call.

## 🔍 How

- **`formatNotification.js`**: Added `PERMANENT_DELETE` and `RESTORE` cases. The `default` branch now falls back to `notificationType.name.toLowerCase()` instead of `''`, so future unknown verbs render as-is without crashing.
- **`NotificationMenuItem.jsx`**: Added guards (truthy check before calling `formatMessage`) and `defaultMessage` fallbacks so missing translation keys display the raw string instead of throwing.
- **Translation files**: Added `permanently_deleted` key to all 15 language files. `restored` already existed.

## 🔗 Related API PR

None — backend already sends `PERMANENT_DELETE` and `RESTORE`.

## 🧪 Testing

1. Log in as a user who has notifications with `PERMANENT_DELETE` type
2. Open the notification menu — no console error, notification renders with "permanently deleted" verb
3. Verify other notification types (`UPDATE`, `CREATE`, etc.) still render correctly

### DIFF
--- a/packages/web-app/public/lang/ar.json
+++ b/packages/web-app/public/lang/ar.json
@@ -1617,6 +1617,7 @@
   "Perform an advanced search on all the fields. Results displayed will have at least one field matching your request.": "أجرِ بحثاً متقدماً في جميع الحقول. ستحتوي النتائج المعروضة على حقل واحد على الأقل يطابق طلبك.",
   "Periodical issue": "عدد دوري",
   "Permanently delete": "حذف نهائي",
+  "permanently_deleted": "تم الحذف نهائيًا",
   "Person": "شخص",
   "Personal information": "المعلومات الشخصية",
   "Persons": "أشخاص",
--- a/packages/web-app/public/lang/bg.json
+++ b/packages/web-app/public/lang/bg.json
@@ -1617,6 +1617,7 @@
   "Perform an advanced search on all the fields. Results displayed will have at least one field matching your request.": "Извършете разширено търсене във всички полета. В показваните резултати ще имат поне едно поле, отговарящо на вашата заявка.",
   "Periodical issue": "Периодично издание",
   "Permanently delete": "Изтрий окончателно",
+  "permanently_deleted": "окончателно изтрито",
   "Person": "Личност",
   "Personal information": "Лична информация",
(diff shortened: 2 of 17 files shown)


Paul-AUB (approved):

1. **[formatNotification.js:127-129]** The `default` branch silently falls back to `notificationType.name.toLowerCase()` for unknown types. Consider adding a `console.warn` so future unhandled notification types surface in development before they reach production:
   ```js
   default:
     console.warn(`[formatNotification] Unknown notificationType: ${notificationType.name}`);
     verb = notificationType.name ? notificationType.name.toLowerCase() : '';
   ```
   Per project conventions, `console.warn` is acceptable (only `console.log` is flagged by ESLint).

2. **[NotificationMenuItem.jsx:65-70]** The `entityType ?` guard is always truthy given the current `formatNotification` implementation — it always returns a non-empty string (either the entity type or `'unknown'`). The guard is harmless and makes the component more defensive against future regressions, but it may give a false sense that `entityType` can be falsy. A code comment explaining the intent would clarify why it's there.

### Nitpicks (Optional)

3. **[NotificationMenuItem.jsx:65]** The `defaultMessage: entityType` fallback means a translation key like `'unknown'` would render as-is (the word "unknown") rather than something user-friendly. This is pre-existing behavior for the unknown-entity path — just worth keeping in mind if that edge case ever needs localization.

PR #1395, opened 2026-06-13
fix(coordinates): handle S/W cardinals and French O (Ouest) notation

Paul-AUB:
## 🤔 What

- Fix coordinate parsing: South (`S`) and West (`W`) cardinals were not negating the decimal value correctly
- Add support for French `O` (Ouest) notation as an alias for `W`
- Fix fast-path bypass that caused coordinates with cardinals (e.g. `22.9° S, 64.5° W`) to be returned positive
- On mobile, open revision history in the same tab instead of a new one (used `useOpenLink` hook)
- Stabilize `useProjections` empty-array return to avoid unnecessary re-renders

## 🤷‍♂️ Why

Coordinates entered with `S`/`W` cardinals (e.g. `22.9179° S, 64.5117° W`) were silently returned with wrong sign — the fast-path for two plain numbers short-circuited before the coordinate-parser could apply the directional sign. French users also commonly write `O` (Ouest) for West, which was not recognized at all.

The mobile history issue caused a blank new tab on mobile browsers because `Link` with `target="_blank"` doesn't respect `useOpenLink`'s same-tab logic.

## 🔍 How

**Coordinate fix (`coordinateConvert.js`):**
- Extracted `isNegativeDir(sign, dir)` helper that accepts `S`, `W`, and `O`/`o` as negative directions
- Extended both DMS and DDM regexes to accept `[NSEWOnsewo]` instead of `[NSEWnsew]`
- Added `normalizeWest(input)` that replaces standalone `O`/`o` with `W` before passing to `coordinate-parser`
- Changed the fast-path condition: plain numbers with cardinal letters now fall through to `coordinate-parser`, which correctly pairs each cardinal with its adjacent number
- Added 3 new unit tests covering `S`/`W`, `O` (Ouest) in decimal, and `O` in DMS format

**Mobile history fix (`UtilityFunction.js`, `DocumentDetails/index.jsx`):**
- Replaced `<Button component={Link} to={...} target="_blank">` with `onClick={() => openLink(url)}` using the existing `useOpenLink` hook
- Same pattern applied to the document details snapshot action

## 🔗 Related API PR

_N/A — frontend-only change._

## 🧪 Testing

1. Open an entrance edit form and enter `22.9179° S, 64.5117° W` — latitude should be negative (~-22.9)
2. Enter `22.9179° S, 64.5117° O` — same result (French notation)
3. Enter `22° 55' 4.4" S, 64° 30' 42.1" O` — DMS with Ouest, both values negative
4. On mobile, click the revision history button — should open in the same tab

## 📸 Previews

_No visual change — bug fix only._

### DIFF
--- a/.agents/skills/code-review/SKILL.md
+++ b/.agents/skills/code-review/SKILL.md
@@ -185,4 +185,3 @@ Report back to the user that the review was submitted, including the PR URL.
 - **Be thorough but respectful.** Critique the code, not the author. Use phrases like "Consider..." or "This might..." rather than "You should..." or "This is wrong."
 - **Reference specific files and line numbers** whenever possible so the author can locate issues quickly.
 - **Check steering files first** — don't flag something as a convention violation unless it actually violates the project's documented conventions.
-- If the diff exceeds 500 added/deleted lines, focus on the most critical files first (business logic, security-sensitive code, public API surfaces) and note this in the review.
--- a/packages/web-app/src/components/appli/Entry/Snapshots/UtilityFunction.js
+++ b/packages/web-app/src/components/appli/Entry/Snapshots/UtilityFunction.js
@@ -1,9 +1,9 @@
 import React from 'react';
 import { useIntl } from 'react-intl';
-import { Link } from 'react-router-dom';
 import { Tooltip, Button } from '@mui/material';
 import HistoryIcon from '@mui/icons-material/History';
 import PropTypes from 'prop-types';
+import useOpenLink from '../../../../hooks/useOpenLink';
 import {
   CommentSnapshots,
   DocumentSnapshots,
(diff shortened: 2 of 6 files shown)


ClemRz (approved):
Good fix — the root cause analysis is solid (fast-path short-circuiting before `coordinate-parser` could apply directional sign), and the tests confirm the corrections. The `normalizeWest` regex is well-bounded (only standalone `O` gets replaced), and the `useOpenLink` migration is the right pattern for consistent mobile/desktop behavior.

- **[coordinateConvert.js:218]** The `/i` flag on `/[NSEWnsew]/i` is redundant since both cases are already explicit in the character class. Consider dropping it for clarity: `/[NSEWnsew]/`. Not a bug, just a readability thing.

### Nitpicks (Optional)

- **[.agents/skills/code-review/SKILL.md]** This line removal (the 500-line diff guidance) seems unrelated to the coordinate/mobile fix. Consider splitting it into a separate commit or PR for a cleaner git history. Not blocking.

PR #1396, opened 2026-06-13
tech(vite): migrate CRA to Vite 8, Vitest and Storybook 10

Paul-AUB:
## 🤔 What

Replace Create React App (`react-scripts` 5.0.1) with **Vite 8** across the entire `packages/web-app` package.

- Build tool: CRA → **Vite 8** (Rolldown/oxc-based bundler)
- Unit tests: Jest → **Vitest** (189 tests / 31 files)
- Storybook: v8 + webpack5 → **v10 + `@storybook/react-vite`** (CSF3 format)
- Code-splitting: all 46 routes converted to `React.lazy` + `<Suspense>`
- Env vars: `REACT_APP_*` / `process.env` → `VITE_*` / `import.meta.env` (typed via `vite-env.d.ts`)
- Compression: Brotli + Gzip on all `dist/` assets
- Bundle analyzer: `dist/stats.html` generated on every build

closes #1392 

## 🤷‍♂️ Why

`react-scripts` has been officially deprecated since February 2025: no more Webpack security patches, unpatched transitive CVEs, and no React 19 RSC support. Vite 8 is the current standard for React SPAs and is a prerequisite for the upcoming React Router v8 and Storybook 9+ migrations.

## 🔍 How

**Key non-obvious decisions:**

- `vite.config.mjs` uses `.mjs` extension — `rollup-plugin-visualizer` is ESM-only and fails to load as `.js`
- **`leaflet-draw` shim** (`src/shims/leaflet-draw.js` + regex alias `/^leaflet-draw$/`): `leaflet-draw` is a side-effect-only CJS IIFE with no default export; `react-leaflet-draw` does `import Draw from 'leaflet-draw'`, which Rolldown rejects at build time. The shim runs the side effect and re-exports `L.Draw`.
- **27 `.js` files containing JSX renamed to `.jsx`**: Vite 8/Rolldown does not honor the `esbuild.loader: 'jsx'` override for `.js` files (unlike Vite 7/esbuild). Renaming aligns with the existing `react/jsx-filename-extension` TODO.
- **`window.intlBootstrap`** explicitly assigned in `index.html` and `.storybook/preview-head.html` so the ESM module bundle (deferred) can read the locale bootstrap global.
- **`eslint-config-react-app` reinstalled** as a devDep: it was bundled with `react-scripts` and suppresses hundreds of strict Airbnb rules on the web-app package. Removing it without restoring the config would break `yarn lint`.
- **`server.warmup`** configured to pre-transform all page and appli components at dev-server startup, eliminating first-navigation latency caused by Vite's on-demand file transformation.
- **Storybook 15 stories rewritten**: the audit incorrectly reported zero `.stories.js` files — 15 `_stories.js` files using the removed `storiesOf` API and `addon-knobs` were found and rewritten to CSF3.
- **`leaflet/dist/leaflet.css` moved to `App.jsx`** (eager): with Vite's CSS code-splitting, the CSS follows its JS chunk. Importing it only in lazy-loaded routes caused the map tiles to render with broken positioning on first visit.

**Bundle impact:**

| Metric | Before (CRA) | After (Vite 8) |
|---|---|---|
| Initial JS bundle | 2 373 KB | 825 KB |
| Gzip | 707 KB | 243 KB |
| Reduction | — | ~65% |

## 🔗 Related API PR

None — frontend-only change.

## 🧪 Testing

```bash
yarn test              # 189 tests / 31 files — all green (Vitest)
yarn build             # outputs dist/ with .br/.gz assets + dist/stats.html
yarn build-storybook   # Storybook 10 build — passes
yarn lint              # exit 0
yarn start             # dev server on :3000, ready in ~800ms
```

Manual checks still needed before merge:
- [ ] Map page — Leaflet tiles + D3 hexbin + leaflet-draw polygon editing
- [ ] Account page — authenticated route (lazy chunk)
- [ ] ApiDoc `/ui/api/1` — swagger-ui-react rendering
- [ ] Locale switching — FR/EN + RTL (`ar`, `he`)
- [ ] Document upload form

## 📸 Previews

Run `yarn build` then open `packages/web-app/dist/stats.html` to inspect the bundle composition.


Paul-AUB:
See also https://github.com/GrottoCenter/grottocenter-front/issues/1392#issuecomment-4699937135

ClemRz (approved):

- **[packages/web-app/src/components/appli/Entry/Snapshots/component/RestoreSnapshot.jsx:133-148]** The `restoreSnapshot` function dispatches an async Redux action then immediately shows the success dialog and reloads the opener. The dispatch hasn't resolved yet at that point — if the API call fails, the user sees "Restore completed" even though it didn't work. Consider awaiting the dispatch result (or chaining on its `.then()`) before showing the success dialog. This was a pre-existing issue but the PR touches this code to fix the `window.opener` crash, so it's worth noting.

- **[packages/web-app/.storybook/preview.jsx:12]** `import messages from './en.json'` — this file is present and serves as the Storybook translation source, which is great. However, this file will drift from the app's `public/lang/en.json` over time. Consider importing `../public/lang/en.json` directly (Vite resolves JSON imports natively) to keep Storybook and the app in sync without needing to maintain a duplicate file.

- **[packages/web-app/src/setupTests.js:1-5]** The comment still references "jest-dom" and "jest matchers". A minor cleanup opportunity — update the comment to mention Vitest since the file now runs under the Vitest test runner.

- **[packages/eslint-config/package.json]** The peerDependency on `eslint-plugin-storybook` was removed, but the root `package.json` still depends on `eslint-plugin-storybook@^0.12.0` and the ESLint config extends `plugin:storybook/recommended`. Version 0.12.x was designed for Storybook 8 — consider verifying it works correctly with the CSF3/SB10 patterns used here (or pin to `>=0.12` to document the minimum). If Storybook 10 support is included in 0.12.x, that's fine; just double-check it doesn't produce false positives on the new CSF3 object-export format.

- **[packages/web-app/vite.config.mjs:61]** The `testTimeout: 30000` is generous globally (30s for all tests). If only property-based tests need extra time, consider configuring the timeout per-file via `vi.setConfig({ testTimeout: 30000 })` in those files (which the `MapMassif.property.test.jsx` already does). Having a high global timeout can mask genuinely slow/stuck tests.

### Nitpicks (Optional)

- **[packages/web-app/src/helpers/polygonUtils.test.js:5]** Comment still says "Jest hoists vi.mock()" — should say "Vitest hoists vi.mock()".

- **[packages/web-app/index.html:162-186]** The inline `<script>` block in `index.html` uses `var` for `RTL_LOCALES`. Not a blocker (it's intentional legacy inline script), but `const` would be more consistent with the rest of the block that already uses `const`.

- **[packages/web-app/src/App.jsx]** The `future` flags (`v7_startTransition`, `v7_relativeSplatPath`, etc.) are React Router v7 future flags that enable v8-like behavior. Worth a comment noting these are intentional preparation for the React Router v8 migration, so future readers don't wonder why "v7" flags exist in a v7 app.

- **[packages/web-app/src/components/appli/Massif/MapMassif.property.test.jsx:108]** `vi.setConfig({ testTimeout: 30000 })` is called at the top of a `describe` block. With `testTimeout` already at 30000 globally in `vite.config.mjs`, this is now redundant. Could be removed for clarity, or kept if the plan is to lower the global timeout later.

urien (approved):
Tout fonctionne,
2 autres personnes ont fait des tests sans trouver de pb

PR #1398, opened 2026-06-16
feat(advancedSearch): add geo export formats for entrance searches

ClemRz:
# Add geo export formats for entrance searches

## 🤔 What

- Add GeoJSON, GPX, and KML export options alongside CSV for entrance search results
- Introduce an `ExportFormatDropdown` component that replaces the "Export to CSV" button when viewing entrances
- Non-entrance entity types retain the existing CSV-only export button unchanged
- Closes #1397

## 🤷‍♂️ Why

Cavers need to export entrance data in geo-aware formats (GeoJSON, GPX, KML) to use in mapping applications and GPS devices. The backend already supports (or will support via GrottoCenter/grottocenter-api#1677) a `format` query parameter on the export endpoint — this PR exposes that capability in the UI.

## 🔍 How

- **Shared config** (`src/conf/exportFormats.js`): Single source of truth for supported formats (array + derived `Set` for validation)
- **Action** (`actions/Advancedsearch.js`): `downloadAdvancedSearchResults` accepts a `format` parameter, validates it against `VALID_EXPORT_FORMATS`, appends `?format=` to the export URL, and uses the format as the file extension. Invalid formats fall back to CSV.
- **ExportFormatDropdown** component: MUI Button + Menu with 4 format options. Handles disabled state with tooltip (row limit exceeded). Text truncates gracefully on mobile via overflow ellipsis.
- **DesktopEntityTable**: Conditionally renders the dropdown for `entrances`, retains the CSV button for other entity types. Derives `visibleColumns`, `exportColumns`, and `exportColumnLabels` as plain constants (no redundant function wrappers).
- **SearchResults**: Passes `format` through to `downloadAdvancedSearchResults`.

## 🔗 Related API PR

- GrottoCenter/grottocenter-api#1677 (backend `?format=` support)

When `format=csv` is passed (the default), behavior is identical to the current implementation — this PR can be merged before or after the API PR.

## 🧪 Testing

- Unit tests for `downloadAdvancedSearchResults`: URL format param, filename extension, fallback for invalid formats
- Property-based tests: valid formats produce correct URL + filename; invalid formats always fall back to CSV
- Component tests for `ExportFormatDropdown`: menu items, selection callback, disabled tooltip, menu close
- Component tests for `DesktopEntityTable`: dropdown for entrances, CSV button for others, format passthrough

All tests pass locally: `yarn test --testPathPattern="(Advancedsearch.test|ExportFormatDropdown.test|DesktopEntityTable.test)"`

### DIFF
--- a/packages/web-app/src/actions/Advancedsearch.js
+++ b/packages/web-app/src/actions/Advancedsearch.js
@@ -1,5 +1,6 @@
 import fetch from 'isomorphic-fetch';
 import { advancedSearchUrl, advancedSearchExportUrl } from '../conf/apiRoutes';
+import { VALID_EXPORT_FORMATS } from '../conf/exportFormats';
 import { checkAndGetStatus } from './utils';
 
 export const FETCH_ADVANCEDSEARCH = 'FETCH_ADVANCEDSEARCH';
@@ -58,8 +59,10 @@ export const downloadAdvancedSearchResults = async ({
   filter,
   matchAllFields = true,
   columns,
-  columnsName
+  columnsName,
+  format = 'csv'
 }) => {
+  const safeFormat = VALID_EXPORT_FORMATS.has(format) ? format : 'csv';
   const data = {
     query,
     entity,
@@ -74,7 +77,9 @@ export const downloadAdvancedSearchResults = async ({
     body: JSON.stringify(data)
   };
 
-  const blob = await fetch(advancedSearchExportUrl, requestOptions)
+  const exportUrl = `${advancedSearchExportUrl}?format=${encodeURIComponent(safeFormat)}`;
+
+  const blob = await fetch(exportUrl, requestOptions)
     .then(checkAndGetStatus)
     .then(response => response.blob())
     .catch(errorMessage => {
@@ -86,7 +91,7 @@ export const downloadAdvancedSearchResults = async ({
   const url = window.URL.createObjectURL(blob);
   const a = document.createElement('a');
   a.href = url;
-  a.download = `Grottocenter_search_export_${Math.trunc(Date.now() / 1000)}.csv`;
+  a.download = `Grottocenter_search_export_${Math.trunc(Date.now() / 1000)}.${safeFormat}`;
   document.body.appendChild(a);
   a.click();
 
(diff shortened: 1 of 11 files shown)


Paul-AUB (approved):

1. **[src/actions/Advancedsearch.test.js:47-50]** The `exportFormats` module is mocked with hardcoded values, even though the comment says "use the real values." Since `exportFormats.js` has no side effects, there is no need to mock it — the real module can be imported directly. As written, if a new format is added to `EXPORT_FORMATS`, these tests will silently pass while the mock drifts out of sync with the real validation set. Consider removing the `jest.mock('../conf/exportFormats', ...)` call entirely and letting the real module be used.

2. **[src/components/common/EntityTable/DesktopEntityTable.test.jsx:57-110]** No test covers the disabled state propagation: when `nbTotalRows > 10000` and `entityType === 'entrances'`, `ExportFormatDropdown` should receive `disabled={true}`. The tooltip and disabled button are tested in `ExportFormatDropdown.test.jsx`, but the integration path from `nbTotalRows` → `disabled` prop is not. Consider adding:
   ```js
   it('renders ExportFormatDropdown as disabled when nbTotalRows exceeds limit', () => {
     renderTable({ entityType: 'entrances', onCSVDownload: jest.fn(), nbTotalRows: 10001 });
     expect(screen.getByRole('button', { name: /Export/i })).toBeDisabled();
   });
   ```

3. **[src/components/common/EntityTable/DesktopEntityTable.jsx:549]** The prop `onCSVDownload` now handles all export formats (GeoJSON, GPX, KML) for entrances, making the name misleading. A rename to `onExport` or `onDownload` would better reflect the expanded semantics. This is non-blocking for this PR but worth tracking as a follow-up since the same prop name propagates through `EntityTable` and `SearchResults`.

### Nitpicks (Optional)

4. **[src/conf/exportFormats.js:1-18]** The file contains two multi-line `/** ... */` JSDoc blocks. The project convention (per AGENTS.md) is "no multi-line comment blocks — one short line max, only when the WHY is non-obvious." The exported names `EXPORT_FORMATS` and `VALID_EXPORT_FORMATS` are self-explanatory; the comments can be removed or collapsed to a single line each.

5. **[src/actions/Advancedsearch.test.js:152-155, 183-186]** Same convention note: the property-based tests have multi-line `/** ... */` doc blocks before the `it()` calls. These describe the property being tested (which the test name already communicates) and can be dropped or shortened.

Paul-AUB (approved):

PR #1409, opened 2026-06-19
fix(entrance): reduce client-side memory pressure during bulk editing (#1405)

Paul-AUB:
## 🤔 What

Fixes client-side memory pressure causing UI freezes during bulk entrance editing (reported by BTH in #1405).

- Skip redundant `fetchEntrance` in `EntranceEdit` when Redux already holds fresh data for that entrance
- Skip `PUT /caves/:id` when cave fields are unchanged on submit
- Skip `PUT /entrances/:id` when entrance fields are unchanged on submit (both saves 0 requests if nothing changed at all)
- Add debounce timer cleanup on unmount in `SearchTextAutocomplete`
- Cap `SearchTextAutocomplete` cache to 50 entries (FIFO eviction)
- Cap `MobileEntityList` accumulated rows to 600 (prevents unbounded DOM growth on mobile "Load more")

closes #1405

## 🤷‍♂️ Why

A moderator reported intermittent browser freezes lasting tens of seconds to several minutes when bulk-editing Belgian entrance names via advanced search. Azure logs confirmed the freeze was 100% client-side (2-minute gap with zero outgoing requests while the server remained healthy).

Root causes: excessive Redux dispatches per edit cycle triggering cascading re-renders, an unbounded autocomplete cache growing over long sessions, and a stale closure risk from the debounce timer — all compounding on a 4 GB RAM machine where Chrome was already at 65% memory.

## 🔍 How

**EntranceEdit — skip redundant fetch:**  
`EntranceEdit` was always dispatching `fetchEntrance` on mount, even when navigating from the entrance view page which had just loaded the same data. `FETCH_ENTRANCE_LOADING` clears `state.entrance.data`, causing a visible spinner flash. Now the fetch is skipped if `state.entrance.data.id` already matches the requested `entranceId`.

**EntranceForm — skip unchanged PUTs:**  
On submit, the form now compares cave and entrance fields against their original values (`caveValues` / `entranceValues` props). Only fields that actually changed trigger PUT requests:
- cave unchanged + entrance unchanged → 0 PUTs, immediate redirect
- cave unchanged + entrance changed → 1 PUT (`/entrances/:id`)
- cave changed → 2 PUTs (`/caves/:id` + `/entrances/:id`) as before

**SearchTextAutocomplete — debounce cleanup:**  
Added `useEffect(() => () => clearTimeout(debounceTimer.current), [])` to prevent a setState call on an unmounted component when the user navigates away mid-debounce.

**SearchTextAutocomplete — cache cap:**  
The `cacheRef` object grew indefinitely since keys were never evicted. Now capped at 50 entries using FIFO eviction (oldest key deleted when limit is reached before inserting a new one).

**MobileEntityList — row accumulation cap:**  
`allRows` grew without bound across "Load more" clicks. Now capped at 600 rows; the "Load more" button is hidden when the cap is reached, prompting the user to refine their search.

## 🔗 Related API PR

None.

## 🧪 Testing

1. Open Advanced Search → filter Belgium (BE) → 7818 results
2. Open DevTools Network, paste in Console:
   ```js
   const _f = window.fetch;
   window.__reqs = [];
   window.fetch = (...a) => { window.__reqs.push((typeof a[0]==='string'?a[0]:a[0]?.url)); return _f(...a); };
   ```
3. Navigate to an entrance → click Edit → change only a flag (e.g. hasBat) → Save  
   → verify only 1 PUT `/entrances/:id` fires (no PUT `/caves/:id`)
4. Navigate to an entrance → click Edit → save without changing anything  
   → verify 0 PUTs fire, redirect happens immediately
5. Navigate to an entrance → click Edit → change the name → Save  
   → verify 2 PUTs fire (cave + entrance), as before
6. On mobile viewport: run "Load more" 30+ times → verify DOM card count stops at 600

## 📸 Previews

N/A — no visual changes.


ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:156-188]** The "nothing to save" path (when `caveUnchanged && entranceUnchanged`) silently does nothing — no dispatch, no user feedback. Because `onSubmit` resolves without throwing, `react-hook-form` sets `isSubmitSuccessful = true`, which renders `FormProgressInfo` with `isLoading=false` and `isError=false`, immediately triggering a redirect. While this "works" (user goes back to the view page), there's no indication that nothing was saved. More critically, if a future change to Redux state causes `entranceLoading` or `caveLoading` to be `true` at that moment (e.g., from a previous unfinished request), the user would see an infinite spinner. Consider explicitly navigating back with a toast notification ("No changes detected") or at minimum adding a brief comment explaining why the silent redirect is intentional:
  ```suggestion
      if (caveUnchanged && entranceUnchanged) {
        // No fields changed — onSubmit resolves successfully which triggers
        // FormProgressInfo with isLoading=false, causing an immediate redirect.
        return;
      } else if (entityType === ENTRANCE_AND_CAVE && !caveUnchanged) {
  ```

- **[packages/web-app/src/components/appli/AdvancedSearch/SearchElements.jsx:272-275]** The FIFO eviction relies on `Object.keys()` preserving insertion order. While this is guaranteed in modern engines for non-integer string keys, the cache key is a JSON string which may start with a `{` character — not an issue. However, `Object.keys()` order is **not** guaranteed to be insertion-order for integer-like keys (e.g., if `JSON.stringify` produces something parseable as an integer, though unlikely here). More practically, this isn't truly FIFO — if a cached entry is accessed again (cache hit path on line 262), it's not moved to the "end," so a frequently-used entry could still be evicted before a rarely-used one. This is acceptable as a simple cap, but the PR description calls it "FIFO eviction" which is slightly misleading. Consider using a `Map` instead, which guarantees insertion order and makes the eviction logic clearer:
  ```suggestion
      const keys = Object.keys(cacheRef.current);
      if (keys.length >= 50) {
        delete cacheRef.current[keys[0]];
      }
  ```
  Or for true FIFO with `Map`:
  ```js
  // Initialize: const cacheRef = useRef(new Map());
  if (cacheRef.current.size >= 50) {
    const oldest = cacheRef.current.keys().next().value;
    cacheRef.current.delete(oldest);
  }
  cacheRef.current.set(cacheKey, hits);
  ```

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:156-188]** The comparison logic is brittle and tightly coupled to the shape of both `caveValues`/`entranceValues` props and the output of `makeCaveData`/`makeEntranceData`. If any field is added to the form in the future, a developer must remember to update this comparison block too, or the optimization silently skips the PUT for that field. Consider extracting this into a helper (e.g., `isDeepEqual` or a dedicated `hasEntranceChanged(formatted, original)` function) co-located with the transformers, or using a shallow comparison utility from a library already in the project (Ramda's `equals` is available).

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:179]** The altitude comparison has a false-positive edge case when altitude is exactly `0`: `(entranceValues.altitude ? Number(entranceValues.altitude) : null)` evaluates to `null` when `entranceValues.altitude` is `0` (falsy), but `entranceDataFmt.altitude` would be `0` (since the form string `"0"` is truthy in the transformer). This means altitude=0 is always detected as "changed." Not a data-loss bug, but it defeats the optimization for caves at sea level. Consider:
  ```suggestion
        (entranceDataFmt.altitude ?? null) === (entranceValues.altitude != null ? Number(entranceValues.altitude) : null) &&
  ```

- **[packages/web-app/src/pages/EntityEdit/Entrance/EntranceEdit.jsx:19]** The `entranceDataId` selector extracts a primitive value, which is good for avoiding unnecessary re-renders. However, there's a subtle concern: if the user navigates from entrance A → edit entrance A → saves (which triggers `UPDATE_ENTRANCE_SUCCESS` but doesn't update `state.entrance.data`) → navigates back → navigates to edit entrance A again, the data could be stale (server might have updated it). This is acceptable for the bulk-editing scenario, but worth documenting with a comment explaining the tradeoff.

- **[packages/web-app/src/components/common/EntityTable/MobileEntityList.jsx:95]** The magic number `600` deserves a brief comment explaining the rationale (e.g., `// ~600 cards ≈ 30 "Load more" clicks × 20 rows/page, capped to prevent DOM bloat`). This helps future maintainers understand whether the number can be tuned.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:155-157]** The empty `if` block (`if (caveUnchanged && entranceUnchanged) { // nothing to save }`) reads awkwardly. An early return is clearer:
  ```suggestion
      if (caveUnchanged && entranceUnchanged) return;
      if (entityType === ENTRANCE_AND_CAVE && !caveUnchanged) {
        dispatch(updateCaveAndEntrance(caveData, entranceDataFmt));
      } else {
        dispatch(updateEntrance(entranceDataFmt));
      }
  ```

- **[packages/web-app/src/components/appli/AdvancedSearch/SearchElements.jsx:272]** Minor: `>= 50` means the cache actually stores 50 entries before evicting. The PR description says "Cap to 50 entries" which is correct, but if the intent is "max 50 entries at any time," consider `> 49` or add a comment clarifying that the 51st insert triggers eviction of the 1st, keeping the max at 50.

urien (approved):
Tout fonctionne parfaitement.
Merci @Paul-AUB

ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/pages/EntityEdit/Entrance/EntranceEdit.jsx:22-28]** The skip-fetch `useEffect` has `entranceDataId` in its dependency array. When the IDs don't match and `fetchEntrance` dispatches, the reducer sets `data: undefined` (`FETCH_ENTRANCE_LOADING`), which changes `entranceDataId` from the old ID to `undefined`. This triggers the effect again — `String(undefined) !== String(entranceId)` is still true — causing a **double fetch**. The original code (always fetch, no condition) only fired once. A guard ref prevents the re-trigger:

  ```suggestion
  const fetchedForRef = useRef(null);

  useEffect(() => {
    if (String(entranceDataId) === String(entranceId)) return;
    if (fetchedForRef.current === entranceId) return;
    fetchedForRef.current = entranceId;
    dispatch(fetchEntrance(entranceId));
  }, [entranceId, dispatch, entranceDataId]);
  ```

  This keeps the skip-fetch optimisation intact while preventing the double-dispatch when navigating to a _different_ entrance.

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:155-192]** The field-by-field comparison is ~30 lines tightly coupled to the shapes of `caveValues`/`entranceValues` AND the output of `makeCaveData`/`makeEntranceData`. If a field is added to the form or transformer in the future, a developer must remember to update this block too — otherwise the optimisation silently skips the PUT for that field. `Ramda.equals` is already available in the project; a helper like `hasChanged(formatted, original, keys)` co-located with the transformers would make the comparison self-maintaining and easier to test in isolation.

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/index.jsx:183-184]** The coordinate comparison lines are correct but non-obvious. A short inline comment would help future readers:
  ```javascript
  // Coordinates are omitted from entranceDataFmt when the entrance is sensitive
  // and the user is non-admin, so treat undefined as "no change intended".
  ```

### Nitpicks (Optional)

- **[packages/web-app/src/pages/EntityEdit/Entrance/EntranceEdit.jsx:23]** The comment says "the entry view page" — per project domain vocabulary, use "entrance" not "entry".

ClemRz (approved):
## Overall

Solid performance PR. The problem diagnosis is thorough, the optimizations are targeted, and each change is individually safe. The skip-fetch logic in `EntranceEdit`, the conditional PUT in `EntranceForm`, and the cache/DOM caps all address real client-side memory pressure vectors without introducing regressions in the happy path.

A few observations below — nothing blocking, mostly about maintainability.

---

### `EntranceEdit.jsx` — skip-fetch guard

The `fetchedForRef` + `entranceDataId` dependency approach works but relies on two overlapping guards to prevent double-dispatch. A simpler alternative would be a single condition with a ref:

```js
useEffect(() => {
  if (String(entrance?.id) === String(entranceId)) return;
  dispatch(fetchEntrance(entranceId));
}, [entranceId, dispatch]);
```

If the concern is that `entrance` might be undefined mid-flight (during loading), you already handle that with `isStale` further down. Adding `entranceDataId` as a dep means Redux selector changes trigger re-evaluation of this effect even when `entranceId` hasn't changed — the ref catches that, but it's an extra layer of complexity for the next reader.

Not a blocker — just a readability note. If you want the ref guard for extra safety, a brief inline comment explaining the double-guard pattern would help.

---

### `transformers.js` — brittle field enumeration

`hasCaveChanged` and `hasEntranceChanged` enumerate every single field. If someone adds `externalUrl` or `accessDescription` to the form next sprint and forgets to add it here, the "skip unchanged PUT" optimization silently drops that update.

Two options to make this safer:

1. **Derive the field list from a shared constant** (e.g., `ENTRANCE_COMPARISON_FIELDS`) that both the comparison functions and PropTypes reference.
2. **Shallow object comparison helper** — normalise both sides to the same shape/types, then `JSON.stringify` compare. Heavier, but immune to drift.

At minimum, a `// IMPORTANT: keep in sync with EntranceEdit.entranceValues` comment would help future contributors.

---

### `EntranceForm` — implicit redirect on no-op submit

When nothing changed, `onSubmit` returns early (no dispatch), `isSubmitSuccessful` flips to `true`, and `FormProgressInfo` immediately redirects because `isLoading=false && isError=false`. This is correct but non-obvious — it relies on the fact that `FormProgressInfo`'s redirect effect fires synchronously on first render when those conditions are already met.

The inline comment you added is good. Consider additionally logging or showing a toast ("No changes detected") rather than silently redirecting — it gives the user feedback that their action was acknowledged rather than wondering if the save actually happened.

---

### `MobileEntityList.jsx` — silent cap

When the 600-row cap is hit, the "Load more" button disappears silently. The PR description mentions "prompting the user to refine their search" but there's no visible message in the code. Adding a small notice like:

```jsx
{!hasMore && allRows.length >= MAX_ACCUMULATED_ROWS && (
  <Typography variant="caption" color="text.secondary" textAlign="center">
    <Translate>Refine your search to see more results</Translate>
  </Typography>
)}
```

would close the gap between intent and implementation.

---

### `SearchElements.jsx` — cache eviction

The `Map`-based FIFO eviction is clean. One minor note: `cacheRef.current.keys().next().value` relies on `Map` insertion-order iteration — which is spec-guaranteed, so this is correct. Good call switching from object to `Map`.

---

### Minor

- The debounce cleanup `useEffect(() => () => clearTimeout(debounceTimer.current), [])` is correct and prevents the stale-timer setState-after-unmount. Appreciated.
- Coordinate comparison using `String()` is safe given `normelizeCoordinate` only replaces commas in strings and returns non-strings as-is. If the API ever returns coordinates with different precision than what `react-hook-form` preserves, you'd get false positives — but that's unlikely in practice.

---

**Verdict:** Approve with minor suggestions. The perf wins are meaningful for the reported use case, and the code is well-commented.

PR #1411, opened 2026-06-19
refactor(login): UI/UX rework of the login modal

Paul-AUB:
# refactor(login): UI/UX rework of the login modal

## 🤔 What

Reworks the login modal (desktop + mobile) for a cleaner, more conventional sign-in experience, and fixes several accessibility and functional gaps along the way.

- **Submittable form**: the email/password fields and the submit button now live in a single `<form>`, so pressing **Enter** submits and password managers behave correctly.
- **Accessibility**: `<label>`/input pairs are properly associated (`htmlFor`/`id`), `autoComplete="email"`/`"current-password"` and `autoFocus` added, `aria-describedby` wired to inline error messages.
- **Inline, per-field validation** (email / password) instead of a stacked list of generic error alerts.
- **Server errors are now displayed**: invalid credentials / server errors were stored in Redux but never shown to the user — they now appear as an `Alert` in every mode (login, password reset, not-verified).
- **Primary action placement**: the full-width **Log in** button sits directly under the fields; the secondary **Sign up** action (outlined, full-width) is clearly separated below a divider.
- **Consistent desktop/mobile layout**: vertically centered content on mobile full-screen, no detached footer CTA.
- **Visual polish**: stronger title, properly positioned close button, normal-case "Forgot password?" link.

## 🤷‍♂️ Why

The previous modal had real issues: the submit button was outside the `<form>` (Enter didn't work), labels weren't associated with inputs, autofill hints were missing, wrong credentials produced **no visible feedback**, and on mobile the full-screen dialog left a large empty area with a detached CTA. The goal is to align the login experience with current UI/UX best practices and fix the accessibility/functional regressions.

## 🔍 How

- `LoginForm` is now a **presentational** component (renders a `div`, not a form); the single `<form onSubmit>` lives in `Login.jsx` and wraps all dialog content variants.
- Validation split into small `validateEmail` / `validatePassword` helpers feeding a `fieldErrors` object; a change-handler factory removes duplication.
- Server-error display derives from `state.login.error`, hidden on dialog open and on keystroke to avoid stale errors. The `Login - ` / `Forgot password - ` message prefixes were removed at the source (`actions/Login.js`) so the view translates a clean i18n key directly (no fragile regex stripping).
- `StandardDialog` gains an opt-in `centerContentMobile` prop and no longer renders an empty actions bar when no `actions` are passed; close button repositioned.
- Two new i18n keys (`An unknown error occurred.`, `A server error occurred…`) added and translated across all 14 language files, then sorted.

## 🔗 Related API PR

_None._

## 🧪 Testing

1. Open the login modal (desktop and mobile widths).
2. Press **Enter** in a field → submits.
3. Submit with empty/invalid email or empty password → inline per-field errors.
4. Submit wrong credentials → error `Alert` is shown (previously silent).
5. Verify the password-reset and "account not verified" flows still display their messages, button, and now any server error.
6. Check tab order, label association, and autofill from a password manager.

## 📸 Previews

image

image


ClemRz (approved):

- **[packages/web-app/src/components/appli/Login.jsx:70]** `formatMessage({ id: authState.error.message })` relies on the error message string being a valid i18n key. This works for the known cases (`"Invalid email or password."`, `"An unknown error occurred."`, `"A server error occurred, please try again later..."`) because those keys were added. However, in `postForgotPassword`, the `errorMessage` can be set to arbitrary server response text (e.g., `json.message || json.error || text`). When that happens, `formatMessage` will emit a `MISSING_TRANSLATION` warning and fall through to showing the raw string — which is acceptable but might show untranslated server-side English text to non-English users. Consider adding a `defaultMessage` fallback:
  ```suggestion
  const serverError =
    !isServerErrorHidden && authState.error?.message
      ? formatMessage({
          id: authState.error.message,
          defaultMessage: authState.error.message
        })
      : '';
  ```
  This makes the fallback explicit and self-documenting.

- **[packages/web-app/src/components/appli/Login.jsx:73-74]** `makeFieldChangeHandler` creates new function references on every render because it's not memoized. Since these are passed to `LoginForm` (which receives them as `onEmailChange`/`onPasswordChange`), the child re-renders on every parent render. This isn't a performance concern today given the component's simplicity, but wrapping with `useCallback` would be more idiomatic for the pattern:
  ```javascript
  const handleEmailChange = useCallback(value => {
    setEmail(value);
    setFieldErrors(prev => ({ ...prev, email: '' }));
    setIsServerErrorHidden(true);
  }, []);
  ```

- **[packages/web-app/src/components/common/StandardDialog/index.jsx:80]** The `$centerMobile` prop line exceeds 80 characters. Consider splitting for readability:
  ```suggestion
        <DialogContent
          $scrollable={scrollable}
          $centerMobile={centerContentMobile}>
  ```

### Nitpicks (Optional)

- **[packages/web-app/src/components/common/LoginForm/index.jsx:29]** The `FieldsWrapper` uses a hardcoded `gap: 12px` pixel value. Per project conventions, prefer `theme.spacing(n)` with integer multiples (e.g., `gap: ${({ theme }) => theme.spacing(1)}` for 8px, or keep the raw `12px` if intentional). Since `12px` is 1.5× the default spacing unit and the steering file says to use raw pixel values for non-standard spacing, this is acceptable — just flagging for awareness.

- **[packages/web-app/src/components/appli/Login.jsx:175]** `LoginButtonMessage` is defined as a component inside the render body of `Login`. This means it gets re-created every render, which loses React reconciliation benefits. It works fine here since it's lightweight, but moving it outside `Login` (or converting it to a simple function returning JSX that's called rather than rendered as `<LoginButtonMessage />`) would be slightly cleaner.

- **[packages/web-app/src/actions/Login.js]** Good cleanup removing the `Login - ` / `Forgot password - ` prefixes from error messages. The error messages are now clean i18n keys. Nice improvement.

Overall this is a solid refactor — the form semantics (native `<form onSubmit>`), accessibility improvements (label association, `autoComplete`, `aria-describedby`), and server error visibility are all meaningful UX wins. The code is well-structured and the translation keys are properly sorted across all language files.

urien (approved):
Merci beaucoup @Paul-AUB

PR #1415, opened 2026-06-21
feat(Footer): add BI link, best practices section and import dashboard link

Paul-AUB:
# feat(Footer): add BI link, best practices section and import dashboard link

## 🤔 What

- Add a **"Statistics (BI)"** link to `bi.grottocenter.org` (Superset) in the footer **Technical** column — closes #1413
- Add a dedicated **"Best practices"** footer column grouping:
  - **Bats protection** (existing wiki link, moved out of the Technical column)
  - **Cave conservation** (new wiki link) — closes #1408
- Rebalance the footer grid for an even column distribution (brand `md:4` + four equal `md:2` columns)
- Add a discreet **BI dashboard** link on the CSV import result page, shown after a successful import
- Add the related i18n keys in all 15 language files

## 🤷‍♂️ Why

- Users had no entry point to the new BI / Superset site (#1413).
- The homepage only linked to bat-protection good practices; underground conservation guidance now has its own pages and deserves a visible, clearly-labelled link, in a section distinct from the technical icons (#1408).
- After importing data, surfacing the BI dashboard helps contributors immediately visualise their contribution.

## 🔍 How

- New external links in `conf/externalLinks.js`: `biLinks`, `caveConservationLinks` (the latter via the existing `generateLinks` helper for FR/EN/fallback).
- `Footer.jsx`: new "Best practices" column with labelled text links (more accessible than icon-only), bats link moved there, and a new "Statistics (BI)" icon in the Technical column.
- New `public/images/icons8/statistics.svg` icon built to match the existing icons8 family (rounded `#D7CCC8` badge with the chart shape cut out in transparency).
- `ImportCSV/ImportSteps/Step4.jsx`: conditional link to the BI dashboard rendered only when `resultImport.total.success > 0` (or duplicates), kept low-key so it doesn't compete with the download buttons.
- i18n keys added to `en.json`, propagated and translated across all languages, then sorted.

## 🔗 Related API PR

_None._

## 🧪 Testing

- Open the homepage and check the footer: the **Best practices** column (Bats protection / Cave conservation) and the **Statistics (BI)** icon in the Technical column; verify links open the right wiki / BI pages.
- Check responsive layout on mobile and desktop (even column distribution).
- Run a CSV import to completion and confirm the BI dashboard link appears on success.

## 📸 Previews

image


ClemRz (changes requested):
### Issues (Must Fix)

- **`src/assets/icons/index.js:29`** The `scientificObservationIcon` export breaks alphabetical ordering — it's placed after `touristicIcon` but should come before `stayOnTrailIcon`. The SKILL.md for create-icon even documents that registration should follow alphabetical order.

- **`src/hooks/useOpenBi.js`** The hook uses raw `fetch()` directly, bypassing the project's standard error-handling utilities (`checkAuthStatus` from `actions/utils.js`). This means a 401 response (expired token) won't trigger the global logout/reload flow that every other authenticated API call uses. If the user's token expires between opening the page and clicking the BI link, they'll get a generic "Unable to open" toast instead of being redirected to log in again. Consider using the existing auth-aware fetch pattern or at minimum handling the 401 case explicitly (e.g. dispatch logout + show login dialog).

- **`src/conf/externalLinks.js:94`** The `biLinks` export is declared but never imported or used anywhere. The `useOpenBi` hook reads `REACT_APP_BI_URL` directly via `BI_BASE_URL`. Either remove the unused export or refactor the hook to import `biLinks['*']` from `externalLinks.js` as the single source of truth (the issue spec suggested this pattern, matching `oaiLinks` / `z3950Links`).

- **`src/pages/homepage/Footer.jsx`** The `useOpenBi` hook is instantiated on the homepage footer, which is visible to all users (including anonymous visitors). Since the BI link requires authentication, consider adding a tooltip or visual indicator that login is required — currently an anonymous user clicking the icon gets a login dialog that may feel unexpected from a footer technical-links section where every other link simply opens a URL.

- **`src/hooks/useOpenBi.js:59`** The `window.open('', BI_TAB_NAME)` call opens a blank tab immediately to avoid popup blockers. This is clever, but if the user is unauthenticated, the tab is opened and immediately closed (`biTab?.close()`). Some browsers flash a blank tab briefly — this could feel jarring. Consider only calling `window.open` after confirming auth status (the tradeoff being potential popup blocking after the async token fetch, which you've noted via the comment).

- **`src/hooks/useOpenBi.unit.test.js`** The test mocks `notistack` directly (`useSnackbar`), but the hook uses `useNotification` (which wraps notistack). If `useNotification` ever changes its internal implementation, the test could silently pass while the actual behavior breaks. Consider mocking `./useNotification` instead of `notistack` for a more robust unit boundary.

- **`src/components/appli/ImportCSV/ImportSteps/Step4.jsx`** The `<Link component="button" disabled={isOpening}>` — MUI's `Link` component doesn't natively support a `disabled` prop. The opacity style handles the visual side, but the button may still be clickable (depending on MUI version). Consider adding `pointer-events: none` when disabled, or use a plain `<Button variant="text">` which properly supports `disabled`.

### Nitpicks (Optional)

- **`.agents/skills/create-icon/SKILL.md`** This is a useful agent skill, but it's unrelated to the BI link / best-practices feature. The PR author noted it's bundled for convenience — consider whether this should be a separate commit/PR for clean git history.

- **`src/components/appli/Entry/index.jsx` / `Network/index.jsx`** Replacing `ScienceOutlinedIcon` with `BiotechIcon` is a visual change to an existing feature (the "Science" tab icon on entrance/network pages). This isn't mentioned in the PR title or description — might be worth a note for reviewers testing the change.

- **`src/pages/homepage/Association.jsx`** Same icon swap (ScienceOutlined → Biotech) in the association goals section — also undocumented in the PR description.

- **Translation key `"Statistics"`** The English value is `"Statistics (BI)"` with the `(BI)` suffix baked into every translation. If the product name ever changes (or you want to drop the acronym), all 15 files need updating. Consider making the key value just `"Statistics"` and appending `" (BI)"` in the component if needed — though this is a minor concern.

urien (approved):
Cela fonctionne bien. Merci a tous les 2, cela va simplificier les choses
merci de corriger le lien vers le wiki / langues avant de publier

ClemRz (approved):

- **`src/hooks/useOpenBi.js:107-112`** The auto-resume `useEffect` that fires after the user logs in will call `openBi()` → `window.open('', BI_TAB_NAME)`, but since it's running inside a `useEffect` (not a user gesture), the browser's popup blocker will likely prevent the tab from opening. The user would have to click the BI icon a second time after logging in. This mirrors the `useAuthNavigate` pattern which calls `navigate(to)` — but navigation doesn't trigger popup blockers while `window.open` does. Consider either (a) showing a toast with a clickable link as fallback when the popup is blocked, or (b) accepting the limitation and documenting it in the JSDoc so future maintainers know the auto-resume is best-effort for the new-tab case.

- **`src/pages/homepage/Footer.jsx` / translation files** The i18n key `"Wiki page for bats"` is now orphaned — it was used as `title` and `alt` for the old bats icon in the Technical column, but that icon was replaced by the BI button. The key still exists in all 15 language files but is no longer referenced in any source file. Consider removing it (or running `yarn translations:sync-with-en` to flag orphans).

### Nitpicks (Optional)

- **Bundled unrelated changes** The `BiotechIcon` swap (Entry, Network, Association pages) and the `scientific-observation` custom icon + entityConfig rename are self-contained and independent from the BI link / best-practices scope. The author noted this in PR comments — just flagging it for git history clarity. If you do further work on this branch, consider separating these into their own commit.

- **`src/assets/icons/scientific-observation.svg:25`** There's a duplicated comment line: `<!-- BiotechIcon — microscope body and arm (MUI @mui/icons-material/Biotech) -->` appears twice in a row. Harmless but easy to clean up.

PR #1416, opened 2026-06-22
Feat/1403 duplicate cave detection

dawoldo:
# feat(entrance): real-time duplicate detection on the entrance creation form

## 🤔 What

Helps users avoid creating duplicate cave/entrance entries by surfacing existing similar entrances in real time, while filling in the creation form. Two complementary informational mechanisms:

- **Name-based suggestions** — as you type the entrance/cave name, an autocomplete dropdown lists existing entrances with similar names (max 10). Selecting one opens a confirmation dialog ("It's the same cave" → opens the existing entrance / "It's a different cave" → dismiss).
- **Location-based detection** — as you enter coordinates, existing entrances within a configurable radius (default **1 km**) are plotted on the form's map as distinct markers, with a name tooltip on hover and the full entrance card on click.

Both are non-blocking (never disable fields or the submit button, never auto-fill anything) and only appear in creation mode, not edit mode.

## 🤷‍♂️ Why

Duplicate entrances/caves are a recurring data-quality problem. The check happens at the natural moment — during creation — so the user can self-correct before submitting, rather than us cleaning up duplicates afterwards. Reuses existing backend APIs, so no schema or endpoint changes are required.

## 🔍 How

Reuses existing backend endpoints unchanged — `POST /api/v1/search` (`entities: ['entrances']`) and `GET /api/v1/geoloc/entrances` (bounding-box params).

- **Data layer** (no UI):
  - `helpers/boundingBox.js` — converts a center point + radius (km) into `sw/ne` lat/lng, handling the pole singularity (`cos(lat) → 0`).
  - `actions/Geoloc.js` — `fetchNearbyEntrances(bounds)`, a raw fetch kept **out of the shared `state.map` Redux slice** so it never disturbs the main map.
  - `hooks/useNameDuplicateSuggestions.js` / `hooks/useNearbyEntrances.js` — local-state hooks (no Redux), 300 ms debounce, stale-response guards, silent error handling. The name hook reuses `fetchQuickSearchRaw` and caps results at 10 client-side (API returns 20).
- **UI**:
  - `NameSuggestionDropdown.jsx` — MUI `Popper` under the name field (wraps both `cave.name` and `entrance.name`) + the confirmation `Dialog`. Closes on click-away/Escape **without** touching form state.
  - `MapMarkerSelector.jsx` — nearby entrances reuse the canonical `EntrancePopup`; rendered as distinct **red circle markers** vs. the central new-entrance pin (satisfies the "visually distinguish" requirement). Also sets `pointer-events: none` on the decorative center pin so marker hover/click events reach the markers underneath.
- **Config**: tunable constants centralised in `conf/config.js` — `DUPLICATE_DETECTION_RADIUS_KM`, reusing existing `AUTOCOMPLETE_DEBOUNCE_DELAY` / `AUTOCOMPLETE_MIN_CHARACTERS`.

Isolation guarantee: the feature touches neither `state.quicksearch` (global search bar) nor `state.map`, and submit/validation logic is untouched.

## 🧪 Testing

1. Log in and open **Add an entrance** (creation mode).
2. **Name suggestions**: type ≥2 characters of an existing entrance name → dropdown appears (loading spinner, then up to 10 results with region/country) → click a result → confirmation dialog. Verify "same cave" navigates to the existing entrance and "different cave" just closes. Confirm the dropdown never modifies your typed value, and Escape/click-away closes it.
3. **Nearby entrances**: enter coordinates within ~1 km of existing entrances → red circle markers appear on the map → hover shows the name, click shows the full entrance card. Clear/invalidate the coordinates → markers disappear.
4. Confirm both features are **absent in edit mode** and that the submit button is never blocked by either.
5. Edge cases: <2 chars hides the dropdown; out-of-range coordinates show no markers; API errors fail silently without blocking the form.

## 📸 Previews

image


ClemRz (commented):
## Overall

Clean, well-scoped feature. The isolation choices (no Redux slice, local hooks, informational-only UI) are sound and match the stated design goals. Good reuse of existing endpoints and components (`EntrancePopup`, `fetchQuickSearchRaw`, `useMarkers`). A few observations below.

---

### `actions/Geoloc.js` — unused default export

```js
export const fetchNearbyEntrances = bounds => ...

export default fetchNearbyEntrances;
```

The hook imports the named export. The `default` export is unused and adds ambiguity. Consider dropping it to keep one canonical import path.

---

### `helpers/boundingBox.js` — new file in deprecated folder

The `helpers/` folder is marked deprecated in the architecture docs (prefer hooks or `util/`). Since `computeBoundingBox` is a pure utility with no React dependency, `src/util/` would be a better home. Non-blocking, but flagging for consistency.

---

### `hooks/useNearbyEntrances.js` — stale key after coordinate change

```js
const debouncedKey = useDebounce(
  hasValidCoords ? `${lat},${lng}` : '',
  AUTOCOMPLETE_DEBOUNCE_DELAY
);
```

If the user blanks out a coordinate field (making `hasValidCoords` false) and then re-enters the *same* coordinates that were last debounced, `debouncedKey` transitions `"x,y" → "" → "x,y"`. Because the key string equals the previous non-empty value, the debounce hook *might* skip the effect depending on timing. In practice the 300 ms window makes this unlikely, but you could guard against it by including a generation counter in the key or clearing `entrances` eagerly in the same branch that sets `hasValidCoords = false`.

---

### `MapMarkerSelector.jsx` — the diff removes the import of `EntranceMarker` but the post-merge file still imports it

Looking at the diff, line 15 replaces the import:

```diff
-import { EntranceMarker } from '../../../common/Maps/common/Markers/Components';
+import { EntrancePopup } from '../../../common/Maps/common/Markers/Components';
```

But the current *base* file (that I read) still has the `EntranceMarker` import on line 15, separate from the `AdditionalMarkers` section. Please double-check the merged result doesn't leave a dead import (the diff looks correct — just flagging in case of rebase drift).

---

### `NameSuggestionDropdown.jsx` — minor UX/a11y notes

1. **Keyboard navigation** — The `Popper` list items are `ListItemButton`, which is good, but there's no `aria-live` region or `role="listbox"` on the container. Screen reader users won't know results appeared. Adding `role="listbox"` to the `<List>` and `role="option"` to each `ListItemButton` would close this gap with minimal effort.

2. **`onInput` + `onFocus` to re-open** — If the user dismisses with Escape and then keeps typing, the dropdown reappears via `onInput`. That feels intentional; just confirming it's by design.

3. **Smart quotes in i18n keys** — `"It's a different cave"` / `"It's the same cave"` use typographic apostrophes (curly `'`). The rest of the codebase mostly uses straight ASCII apostrophes. Not a bug, but could trip up searches or key lookups if another contributor adds the key with a straight quote. Worth normalizing.

---

### `EntranceDetail.jsx` — prop drilling `latitude` / `longitude`

The parent (`index.jsx`) passes `lat` and `lng` as props. Since `EntranceDetail` already receives `control`, it could `useWatch` the coordinates internally (same as `NameSuggestionDropdown` does with `name`). That would remove two props and their associated PropTypes. Low priority but would reduce coupling.

---

### Translation keys

Only `en.json` is touched. Per project convention the other 14 language files should also get the new keys (even if machine-translated). If this is intentionally deferred to a follow-up or automated sync, that's fine — just confirming intent.

---

### Missing tests

No unit or property-based tests were added for:
- `computeBoundingBox` (pure function, easy to test)
- `useNameDuplicateSuggestions` / `useNearbyEntrances` (hook logic with debounce + stale-response guards)

Given the project's testing conventions and the SDD requirement for property-based tests, coverage for at least `computeBoundingBox` would be worthwhile.

---

None of these are merge blockers. Nice feature overall.

ClemRz (commented):
## Overall

Solid, well-scoped feature. The architecture — local-state hooks that stay completely isolated from the shared Redux slices, non-blocking UI, and a clean creation-only guard via `enabled` — is well thought out. The bounding-box utility is nicely unit-tested with property-based tests. The code is readable and follows project conventions. A few items to tighten up before merge:

---

## Issues

### 1. `useNearbyEntrances` — missing cleanup on unmount

If the component unmounts while a fetch is in-flight, the stale-response guard (`requestId !== requestIdRef.current`) prevents a state update on the _next_ request, but on unmount the ref is garbage-collected and React will still warn about `setEntrances` being called on an unmounted component. Consider an `isMounted` ref or an AbortController:

```js
useEffect(() => {
  // ...
  const abortController = new AbortController();
  fetchNearbyEntrances(bbox, { signal: abortController.signal })
    .then(...)
    .catch(...);
  return () => abortController.abort();
}, [debouncedKey, radiusKm]);
```

(Same applies to `useNameDuplicateSuggestions`.)

### 2. `NameSuggestionDropdown` — `sx={{ mt: 0.5, ... p: 1.5 }}`

The architecture steering says:
> Non-integer values with `theme.spacing` — use integer multiples only, or raw pixel values for non-standard spacing.

`mt: 0.5` and `p: 1.5` use non-integer multipliers passed to MUI's `sx` shorthand (which calls `theme.spacing` under the hood). Use `mt: '4px'` / `p: '12px'` or round to the nearest integer (`mt: 1` / `p: 2`).

Same applies to `gap: 0.5` and `gap: 0.75` in the legend Box inside `MapMarkerSelector`.

### 3. Translations — several keys left untranslated

Multiple keys are added with the English fallback in non-English files:
- "An entrance with a similar name already exists…"
- "Is this the same cave?"
- "It's a different cave" / "It's the same cave"
- "Searching for similar entrances…"
- "See the entrance"
- "View full details"
- "Existing entrances with a similar name:"

Only "Existing nearby entrances" appears translated across files. The others just repeat the English string. I understand translations are AI-generated, but ideally at least the major languages (fr, es, de, it, pt) should carry actual translations rather than English placeholders.

### 4. `fetchNearbyEntrances` in `actions/Geoloc.js` — double `.json()` possible?

```js
fetch(makeUrl(getMapEntrancesUrl, bounds))
  .then(checkAndGetStatus)
  .then(response => response.json());
```

`checkAndGetStatus` already calls `response.json()` in some code paths (depending on how the project's `checkAndGetStatus` is written). Double-check that you're not calling `.json()` on an already-parsed body. If `checkAndGetStatus` returns the parsed JSON (as `checkAndGetStatus` does for success responses in other action files), you'd get a `TypeError: response.json is not a function`.

### 5. `useNearbyEntrances` — `generation` in deps

The `useEffect` that sets `generation` depends on `hasValidCoords`, but `hasValidCoords` is derived from `lat`, `lng`, `enabled`, and `zoomedInEnough`. If `enabled` flips from true → false → true with the same coords, the generation bump correctly forces a re-fetch. Good design — just calling it out as verified.

### 6. `MapMarkerSelector` — `onZoomChange` prop could cause infinite re-render

`ZoomReporter` has `onZoomChange` in the `useEffect` deps array. If the parent passes an unstable callback (e.g., an inline arrow `onZoomChange={z => setMapZoom(z)}`), this will loop. In `EntranceDetail.jsx` you pass `setMapZoom` directly (a stable setter), which is fine. But for robustness, consider wrapping the dep in a ref inside `ZoomReporter`:

```js
const cbRef = useRef(onZoomChange);
cbRef.current = onZoomChange;
useEffect(() => { cbRef.current(map.getZoom()); }, [map]);
```

### 7. Minor: `"See the entrance"` key is defined in translations but not used in the diff

I don't see this key referenced anywhere in the new code. Is it leftover from an earlier iteration?

---

## Nits (non-blocking)

- `NameSuggestionDropdown` — the `aria-live="polite"` on the Paper is good for accessibility, but the `role="listbox"` on the `<List>` should be paired with `aria-activedescendant` or keyboard arrow navigation for full ARIA compliance. Not a blocker, but worth a follow-up.
- `boundingBox.js` — the exported `default` (`export default computeBoundingBox`) is redundant since you also have the named export and the hooks import the named one. Consider dropping the default to avoid ambiguity.
- The `EntrancePopup` inside `AdditionalMarkers` reuses the canonical popup — nice touch for consistency.

---

## Summary

The feature is well-isolated, non-blocking to the form, and addresses a real data-quality problem. Main actionable items are the non-integer spacing values (project rule violation), the potential unmount memory-leak pattern in the hooks, and verifying `checkAndGetStatus` doesn't double-parse JSON. Once those are addressed this is good to merge.

urien (approved):
Cela fonctionne très bien, pas de retour des membres de l'association
Merci @dawoldo

PR #1417, opened 2026-06-22
feat(entrance): add network description link

Paul-AUB:
# feat(entrance): add network description link

## 🤔 What

On the entrance page, when the entrance belongs to a network (a cave with 2+ entrances), show a callout in the Descriptions panel that links to the network and its description count.

## 🤷‍♂️ Why

Network descriptions are stored on the cave, not on the individual entrance. Previously, nothing indicated their existence when viewing an entrance — users had no path to find them from the entrance page.

## 🔍 How

- `EntryPage` fetches the network cave when `entrance.cave.entrances.length > 1`, reads `descriptions.length` from Redux state, and passes `networkDescriptionsCount` down
- `Entry` forwards `networkId`, `networkName`, and `networkDescriptionsCount` to the `Descriptions` panel
- `Descriptions` renders a callout with two inline `MuiLink` buttons: one opening `/ui/caves/<id>` (network page) and one opening `/ui/caves/<id>#description` (anchor to descriptions), via `useOpenLink`
- The panel auto-expands when network descriptions exist, even if the entrance itself has none
- i18n keys `network.descriptions.callout` and `network.descriptions.count` added to all lang files

## 🔗 Related API PR

None.

## 🧪 Testing

1. Open an entrance that belongs to a network (cave with 2+ entrances)
2. The Descriptions panel should show a callout mentioning the network name and description count
3. Click the network link → navigates to `/ui/caves/<id>`
4. Click the descriptions count link → navigates to `/ui/caves/<id>#description`
5. Open an entrance that does **not** belong to a network → no callout

## 📸 Previews

image

### DIFF
--- a/packages/web-app/public/lang/ar.json
+++ b/packages/web-app/public/lang/ar.json
@@ -1651,6 +1651,8 @@
   "Network name": "اسم الشبكة",
   "Network properties": "خصائص الشبكة",
   "network_count": "عدد الشبكات",
+  "network.descriptions.callout": "هذه المدخل جزء من {networkLink} التي تحتوي أيضاً على {descriptionsLink}.",
+  "network.descriptions.count": "{count, plural, one {وصف واحد} other {# أوصاف}}",
   "networks": "شبكات",
   "Networks": "الشبكات",
   "Networks and caves": "الشبكات والكهوف",
--- a/packages/web-app/public/lang/bg.json
+++ b/packages/web-app/public/lang/bg.json
@@ -1651,6 +1651,8 @@
   "Network name": "Пещерна система",
   "Network properties": "име на Пещерна система",
   "network_count": "брой мрежи",
+  "network.descriptions.callout": "Този вход е част от {networkLink}, която също има {descriptionsLink}.",
+  "network.descriptions.count": "{count, plural, one {едно описание} other {# описания}}",
   "networks": "мрежи",
   "Networks": "Мрежи",
   "Networks and caves": "Мрежи и пещери",
--- a/packages/web-app/public/lang/ca.json
+++ b/packages/web-app/public/lang/ca.json
@@ -1651,6 +1651,8 @@
   "Network name": "Nom del sistema de coves",
   "Network properties": "Propietats del sistema de coves",
   "network_count": "nombre de xarxes",
+  "network.descriptions.callout": "Aquesta entrada forma part de {networkLink} que també té {descriptionsLink}.",
+  "network.descriptions.count": "{count, plural, one {una descripció} other {# descripcions}}",
   "networks": "xarxes",
   "Networks": "Xarxes",
   "Networks and caves": "Xarxes i coves",
--- a/packages/web-app/public/lang/de.json
+++ b/packages/web-app/public/lang/de.json
@@ -1651,6 +1651,8 @@
   "Network name": "Systemname",
   "Network properties": "Systemeigenschaften",
   "network_count": "Netzwerkanzahl",
+  "network.descriptions.callout": "Dieser Eingang ist Teil von {networkLink}, das auch {descriptionsLink} hat.",
+  "network.descriptions.count": "{count, plural, one {eine Beschreibung} other {# Beschreibungen}}",
   "networks": "Netzwerke",
   "Networks": "Netzwerke",
   "Networks and caves": "Netzwerke und Höhlen",
--- a/packages/web-app/public/lang/el.json
+++ b/packages/web-app/public/lang/el.json
@@ -1651,6 +1651,8 @@
   "Network name": "Όνομα δικτύου",
   "Network properties": "Ιδιότητες δικτύου",
   "network_count": "αριθμός δικτύων",
+  "network.descriptions.callout": "Αυτή η είσοδος ανήκει στο {networkLink} που έχει επίσης {descriptionsLink}.",
+  "network.descriptions.count": "{count, plural, one {μία περιγραφή} other {# περιγραφές}}",
   "networks": "Δίκτυα",
   "Networks": "Δίκτυα",
   "Networks and caves": "Δίκτυα και σπήλαια",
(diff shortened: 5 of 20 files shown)


ClemRz (changes requested):
### Issues (Must Fix)

- **[packages/web-app/src/pages/Entry/index.jsx:18]** Shared `state.cave` conflict — The `EntryPage` now calls `fetchCave(networkCaveId)` which writes into `state.cave`. This is the same Redux slice used by the `NetworkPage` (`pages/Network/index.jsx`). Since both pages are never rendered simultaneously (different routes), this isn't a runtime collision *per se*, but the cave reducer also handles description mutations (`POST_DESCRIPTION_SUCCESS`, `DELETE_DESCRIPTION_SUCCESS`, etc.) intended for the *currently viewed cave page*. If a user is on the entrance page and adds a description to that entrance, the cave reducer will still react to those description actions against `state.cave` — which now holds the *network* cave, not the entrance's own entity. This could produce unexpected stale or corrupted state in `state.cave`.

  Consider one of these alternatives:
  1. **Don't use `fetchCave` / `state.cave`** — instead, derive the network descriptions count from data already present in `state.entrance` (if the API already returns `cave.descriptions` on the entrance endpoint), or add a lightweight selector/action that only fetches the count without polluting the shared cave slice.
  2. **Use a dedicated slice** (e.g., `state.networkDescriptions`) so the entrance page's fetch doesn't interfere with the cave page's state.

- **[packages/web-app/src/components/appli/Entry/index.jsx:425-426]** Duplicated network condition — The expression `entrance.cave?.entrances?.length > 1` is evaluated twice (once for `networkId`, once for `networkName`). Consider computing this once:
  ```javascript
  const isNetwork = entrance.cave?.entrances?.length > 1;
  // ...
  networkId={isNetwork ? entrance.cave?.id : undefined}
  networkName={isNetwork ? entrance.cave?.name : undefined}
  ```
  The same pattern is duplicated in `pages/Entry/index.jsx` (line 26). A small helper or a single derived variable at the page level could DRY this up and pass it down.

- **[packages/web-app/src/components/appli/Descriptions/index.jsx:89]** The `<Divider sx={{ mb: 1 }} />` at the top of the callout creates a visual separator at the very beginning of the content area (before any other content). Since `ScrollableContent` already wraps content in `CardContent` with its own padding, this leading divider may look odd — especially when the panel auto-expands with no entrance descriptions below. Consider whether this divider is needed, or move it below the callout to separate it from the description list / "no descriptions" alert.

- **[packages/web-app/src/components/appli/Descriptions/index.jsx:44]** The `openLink` hook (and its import) is only used when `hasNetworkDescriptions` is true, but it's always instantiated (calls `useNavigate` internally). This is fine for hooks rules compliance, but worth noting that the hook is a no-op for non-network entrances. No action needed — just confirming the hooks rules are respected.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/Descriptions/index.jsx:74-75, 82-83]** The PR removes the trailing space before `>` in JSX closing tags (e.g., changing `}\n          >` to `}>`). While this passes lint, it's an unrelated formatting change that makes the diff noisier. Prettier should handle this consistently — these changes are fine to keep, just noting they're cosmetic.

- **[packages/web-app/src/pages/Entry/index.jsx:26]** `networkCaveId` is declared between the two `useEffect` hooks. Placing derived values above the effects they feed is good practice — this reads well. Minor style note: `const networkCaveId` could use a comment explaining the "entrances > 1 means it's a network" domain rule for future readers unfamiliar with the vocabulary.

- **[packages/web-app/public/lang/en.json]** Translation keys use dot notation (`network.descriptions.callout`) while most existing keys use spaces or underscores. Both patterns exist in the file, so this is acceptable — just noting the inconsistency for awareness.

ClemRz (approved):

- **[packages/web-app/src/pages/Entry/index.jsx:25-28]** Flash of stale `networkDescriptionsCount` when navigating between two network entrances — When the user navigates from entrance A (network with 5 descriptions) to entrance C (different network with 2 descriptions), the sequence is: `FETCH_ENTRANCE_SUCCESS` updates `data` → component re-renders with the new `isNetwork=true` but the old count (5) → the second effect fires and fetches the correct count (2). During that brief window, the callout displays the stale count from the previous network.

  Consider resetting the count when `entranceId` changes:
  ```javascript
  useEffect(() => {
    dispatch(fetchEntrance(entranceId));
    // Reset stale network count from a previous entrance
    dispatch({ type: FETCH_NETWORK_CAVE_DESCRIPTIONS_COUNT_SUCCESS, count: 0 });
  }, [entranceId, dispatch]);
  ```
  Or alternatively, add a `RESET_NETWORK_DESCRIPTIONS_COUNT` action type for clarity.

- **[packages/web-app/src/actions/Cave/GetNetworkCaveDescriptionsCount.js:23-25]** The `.catch()` block silently swallows non-auth errors. If the cave endpoint returns a 404 or 500, the callout simply won't appear — which is acceptable graceful degradation. However, there's no way to distinguish "the network has 0 descriptions" from "the fetch failed" since both result in count=0. Consider at minimum logging a `console.error` for non-auth failures so debugging is possible:
  ```javascript
  .catch(error => {
    // Graceful degradation: callout won't show
    console.error('Failed to fetch network descriptions count', error);
  });
  ```

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/Descriptions/index.jsx:89]** The `<Divider sx={{ mb: 1 }} />` renders as the very first element of the `content` prop, immediately below the panel header. From the screenshot it looks fine, but since `ScrollableContent` already provides structural separation via the card header, you might consider whether a simple `sx={{ mb: 1 }}` on the `<Typography>` alone is sufficient without the leading `Divider`.

- **[packages/web-app/src/components/appli/Descriptions/index.jsx:74-75, 82-83]** Cosmetic JSX formatting changes (removing space before `>` in closing tags). These are fine and pass lint — just noting they're unrelated to the feature.

urien (approved):
Cela fonctionne super bien et c'est très pratique. Merci @Paul-AUB

PR #1421, opened 2026-06-26
feat(map): track and display explored entrances (API #1702)

Paul-AUB:
## 🤔 What

Track and display "explored entrances" at the entrance level (instead of cave level), and show them on the map.

- Migrate explored tracking from cave-level to entrance-level (aligns with API PR #1702)
- Display explored entrances as green pins on the main map for logged-in users (toggle in data control legend)
- Persist the "show explored" filter preference via localStorage
- Show a dedicated explored-entrances map on Account and Person profile pages
- Remove the "explored" button from the Network page (exploration is per-entrance, on each Entrance page)
- Fix mobile card selection UX in search forms: tap-to-toggle with checkbox + primary color tint (was broken — no selection possible on mobile)
- Rename all internal references from "caves" to "entrances" for clarity
- Update user-facing strings and all 14 language files

Closes #1420

## 🤷‍♂️ Why

The API (PR #1702) moves the explored relationship from `caves` to `entrances`, giving each explored entrance its own latitude/longitude. This eliminates the N+1 fetch problem (previously required one `GET /caves/:id` per explored network to resolve coordinates), makes the data model more precise, and enables the map to display all explored entrances without any additional requests.

## 🔍 How

- New actions: `linkExploredEntrance` / `unlinkExploredEntrance` (`PUT/DELETE /entrances/:id/cavers/:id`)
- `useExploredEntrances` hook: reads `person.exploredEntrances` directly (lat/lng included in payload), no N+1 fetching
- `ExploredEntrancesMap`: standalone Leaflet map with `BoundsFitter` (debounced `fitBounds` on first load)
- `ExploredOverlay`: green SVG pin markers with cluster support; `BoundsFitter` component for auto-centering
- `RelatedCaves`: branched on `isOrganization` — cavers use entrance-level link/unlink, organizations keep cave-level (unchanged)
- Mobile selection: `MobileEntityCard` wrapped in `React.memo`, `handleToggle` stable via `useCallback([], [])` + `onSelectedRef` pattern to avoid cascade re-renders; `visibleColumns` memoized in `EntityTable`

## 🔗 Related API PR

https://github.com/GrottoCenter/grottocenter-api/pull/1702

> ⚠️ The staging site will not work correctly until API PR #1702 is merged and deployed. The old cave-caver endpoints (`PUT/DELETE /caves/:id/cavers/:id`) no longer exist in the new API.

## 🧪 Testing

1. Log in with an account that has explored entrances
2. **Main map** — open data control legend → check "My explored entrances" → green pins appear on explored entrances, no network requests spam; reload → preference is remembered
3. **Account page** → Activities tab → "Explored entrances" section: map centered on explored points + flat list; add/remove an entrance via search form ("Mark as explored" button)
4. **Person profile page** → Activities tab → same section visible (read-only for other users)
5. **Entrance page** → "Add to my explored entrances" / "Remove from my explored entrances" button toggles correctly
6. **Network page** → no explored button
7. **Organization page** → cave linking unchanged
8. **Mobile** → search form: tap a result card to select it (checkbox appears, card tints), tap again to deselect; "Mark as explored" button enables when ≥1 card selected
9. Log out → explored pins disappear; view another user's profile → only their entrances shown


ClemRz (changes requested):
### Issues (Must Fix)

- **[Person/Person.jsx:286]** `RelatedCaves` is rendered without `isCaveSearchVisible` and `onToggleCaveSearch` props, but both are declared `isRequired` in the component's PropTypes. This fires PropTypes warnings in dev and could become a runtime error if `onToggleCaveSearch` were ever called unconditionally (it's called inside `onSubmitForm` without a null guard). Consider either making these props optional in `RelatedCaves` PropTypes (with sensible defaults) or passing them from Person:
  ```suggestion
  <RelatedCaves
    exploredEntrances={person.exploredEntrances}
    entityId={person.id}
    isOrganization={false}
    canManageCaves={false}
    onRefresh={handleRefresh}
    isCaveSearchVisible={false}
    onToggleCaveSearch={() => {}}
    userId={person.id}
  />
  ```

- **[RelatedCaves/RelatedCaves.jsx:172]** The `submitLabel` is unconditionally set to `"Mark as explored"` regardless of `isOrganization`. When the organization path uses `SearchCaveForm`, the button will read "Mark as explored" instead of the previous "Associate" label, which is semantically wrong for organizations (they "associate" caves, not "explore" them). Consider branching on `isOrganization`:
  ```suggestion
  <SearchCaveForm
    onSubmit={onSubmitForm}
    submitLabel={
      isOrganization ? null : formatMessage({ id: 'Mark as explored' })
    }
  />
  ```
  Passing `null` will fall through to the default `"Associate"` label via the existing `submitLabel ?? formatMessage({ id: 'Associate' })` logic in `SearchCaveForm`.

- **[DataControl.jsx:457]** `hasExploredData` is typed as `PropTypes.bool` but the hook returns `null` as a distinct third state (data not yet loaded). The component logic correctly uses `=== false` comparisons, but the PropType is inaccurate. Consider `PropTypes.oneOf([true, false, null])` to match the actual tri-state semantics.

- **[ExploredOverlay.jsx:84]** The tooltip content interpolates `m.name` directly into an HTML string without escaping. This matches existing patterns in `Markers.jsx` (where `entranceTip` returns the raw name), but in this case the HTML context is explicit (`<span>......${m.name}</span>`). If an entrance name ever contains `<` or `&` characters, the tooltip could break or render unexpectedly. Low risk since entrance names come from authenticated community contributors, but a simple escape function would harden it.

- **[useExploredEntrances.js]** Both `useExploredEntrances` and `useExplored` independently read from and dispatch to the same `state.person` slice. If the main map hook fetches person data for the logged-in user, and the user simultaneously navigates to another person's profile (which also dispatches `fetchPerson`), they'll overwrite each other. This is a pre-existing architectural limitation (single `state.person` slot), not introduced by this PR, but worth noting since the explored overlay makes this shared state visible on more pages simultaneously.

### Nitpicks (Optional)

- **[useLocalStorage.js:22-23]** The `serRef`/`desRef` are initialized but never updated after mount. If a caller were to change `serialize`/`deserialize` between renders (unlikely but possible), the refs would hold stale values. This is fine for the current usage where options are always inline arrow functions defined in the same render, but adding `serRef.current = serialize; desRef.current = deserialize;` at the top of the hook body would make it fully resilient.

- **[MobileEntityList.jsx:189]** The `eslint-disable-next-line` comment is acceptable given the `onSelectedRef` pattern, but the project convention asks for a written justification in the preceding comment. The existing comment ("Notify parent only when selection actually changes, not on mount") provides context about *what* the effect does but doesn't explicitly justify *why* the dep is excluded. Consider adding: `// onSelectedRef is a ref — intentionally excluded from deps`.

ClemRz (approved):
### Re-review: Previously raised issues

All four items from the previous review have been addressed:

1. **`isCaveSearchVisible` / `onToggleCaveSearch` props** — Now optional with JS defaults (`false` / `noop`). Person page no longer triggers PropTypes warnings. ✅
2. **`submitLabel` for organizations** — Acknowledged as intentional (organizations also "explore" entrances). The author's rationale is clear and the PR comment documents it. ✅
3. **`hasExploredData` PropTypes** — Changed to `PropTypes.oneOf([true, false, null])`, matching the tri-state semantics. ✅
4. **HTML escaping in tooltip** — Added `&`, `<`, `>` escaping via `HTML_ESCAPES` map before interpolation. ✅
5. **eslint-disable justification** (nitpick) — Comment now explicitly explains why `onSelectedRef` is excluded from deps. ✅

- **[useLocalStorage.js:22-23]** The `serRef`/`desRef` refs are initialized once but never updated on subsequent renders. This is fine for the current usage (all callers pass inline arrow functions that are semantically stable), but if a caller ever passed a changing serializer, the stale ref would silently produce wrong results. A single line `serRef.current = serialize; desRef.current = deserialize;` after the ref declarations would make the hook fully robust — low priority since current call sites are safe.

- **[apiRoutes.js:305-309]** `linkExploredEntranceUrl` and `unlinkExploredEntranceUrl` produce the exact same URL string. This is technically correct (PUT vs DELETE is determined by the action creator), but having two identical functions might confuse future readers. A single `exploredEntranceUrl(entranceId, caverId)` used in both action creators would remove duplication — though this is more a matter of taste and not blocking.

### Nitpicks (Optional)

- **[ExploredOverlay.jsx:84]** The `HTML_ESCAPES` map only escapes `&`, `<`, `>`. The `"` and `'` characters are also potentially dangerous in HTML attribute contexts (though here the name is placed in element content, not an attribute). Current coverage is sufficient for this use case.

---

Overall this is a solid follow-up push. The previously raised concerns are all resolved cleanly. The feature implementation is well-structured: `useExploredEntrances` hook is reusable, the `ExploredOverlay` component is decoupled, and the `useLocalStorage` hook is a nice addition for persisting map preferences. LGTM.

PR #1423, opened 2026-06-27
fix(map): center explored entrances map reliably on load

### LINKED ISSUE #1422: Personnal explored map doesn't always load
At first load, sometimes the map is focused in the ocean

Image

Need to investigate to knwo what's going on. 
Can be "fixed" by F5/reload the page.

Paul-AUB:
## 🤔 What

Fix the explored-entrances map on the Account and Person profile pages
initializing on wrong coordinates (open ocean) until a full page refresh (F5).
The bug was intermittent.

Closes #1422

## 🤷‍♂️ Why

The map mounts inside a tab and a collapsible (`ScrollableContent`), so its
Leaflet container is often still zero-sized or mid-animation at mount time.
`CustomMapContainer` corrects the size asynchronously via a `ResizeObserver` →
`invalidateSize`.

`BoundsFitter` ran `fitBounds` exactly once, on a blind 300 ms timer, behind a
permanent one-shot guard. When the timer fired before the container reached its
final size, `fitBounds` computed against a stale/zero viewport, landed on the
wrong center+zoom (≈ 0,0), and the one-shot guard made it permanent until reload.
Whether the timer won or lost the race against the resize explained the
intermittency.

Aggravating factor: the map received no initial `center`/`zoom`, so before the
first successful fit it sat at Leaflet's ≈ 0,0 default (open ocean).

## 🔍 How

- **Rewrite `BoundsFitter`** (`ExploredOverlay.jsx`): drop the blind timer and
  the permanent one-shot guard. `fit()` only runs once `map.getSize()` is
  non-zero, and re-fits on every Leaflet `resize` (emitted by
  `CustomMapContainer` via `invalidateSize` as the container settles) until the
  user interacts (`dragstart`/`zoomstart`). Our own `fitBounds` moves are ignored
  via an `isProgrammaticRef` guard (reliable because `animate: false` makes the
  events synchronous). This converges to the correct framing regardless of the
  mount / data-arrival / resize ordering.
- **Derive the initial center** (`ExploredEntrancesMap.jsx`) from the explored
  points' bounding-box center instead of a hardcoded region, so the map is
  always initialized on the user's actual data — anywhere in the world.

## 🔗 Related API PR

_None — frontend-only fix._

## 🧪 Testing

1. Log in with an account that has explored entrances
2. Go to the Account page → Activities tab → "Explored entrances" section
3. The map must center on the explored entrances on first load, **without** a
   refresh — repeat a few times (it was intermittent)
4. Same check on a Person profile page
5. Verify an account whose entrances are outside Europe also centers correctly
6. Pan/zoom the map, then toggle fullscreen — the map should not fight the user
   after interaction

## 📸 Previews

_Before: map showed open ocean until F5. After: centered on the explored
entrances on first load._

### DIFF
--- a/packages/web-app/src/components/common/Maps/MapClusters/ExploredEntrancesMap.jsx
+++ b/packages/web-app/src/components/common/Maps/MapClusters/ExploredEntrancesMap.jsx
@@ -11,12 +11,18 @@ import useExploredEntrances from './useExploredEntrances';
 // (never a hardcoded region). BoundsFitter then refines the exact zoom/padding.
 const getInitialCenter = points => {
   if (points.length === 0) return [20, 0]; // defensive; map renders only with data
-  const lats = points.map(p => p.latitude);
-  const lngs = points.map(p => p.longitude);
-  return [
-    (Math.min(...lats) + Math.max(...lats)) / 2,
-    (Math.min(...lngs) + Math.max(...lngs)) / 2
-  ];
+  // Single pass — avoids spreading a potentially large array onto the call stack.
+  let minLat = Infinity;
+  let maxLat = -Infinity;
+  let minLng = Infinity;
+  let maxLng = -Infinity;
+  points.forEach(({ latitude, longitude }) => {
+    if (latitude < minLat) minLat = latitude;
+    if (latitude > maxLat) maxLat = latitude;
+    if (longitude < minLng) minLng = longitude;
+    if (longitude > maxLng) maxLng = longitude;
+  });
+  return [(minLat + maxLat) / 2, (minLng + maxLng) / 2];
 };
 
 const ExploredEntrancesMapInner = ({ points }) => (
@@ -58,7 +64,7 @@ const ExploredEntrancesMap = ({ userId }) => {
       <CustomMapContainer
         wholePage={false}
         center={initialCenter}
-        zoom={5}
+        zoom={5} // placeholder until BoundsFitter fits to the points
         dragging={!isMobile}
         scrollWheelZoom={false}>
         <ExploredEntrancesMapInner points={points} />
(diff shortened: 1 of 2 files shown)


ClemRz (approved):

- **[ExploredEntrancesMap.jsx:13-14]** `Math.min(...lats)` / `Math.max(...lats)` spreads the entire array onto the call stack. If a user has explored a very large number of entrances (10k+), this could hit the JS engine's argument limit and throw a RangeError. In practice, cavers likely won't hit this, but a safe `reduce`-based approach is trivial and future-proof:
  ```suggestion
  const getInitialCenter = points => {
    if (points.length === 0) return [20, 0];
    let minLat = Infinity, maxLat = -Infinity, minLng = Infinity, maxLng = -Infinity;
    for (const { latitude, longitude } of points) {
      if (latitude < minLat) minLat = latitude;
      if (latitude > maxLat) maxLat = latitude;
      if (longitude < minLng) minLng = longitude;
      if (longitude > maxLng) maxLng = longitude;
    }
    return [(minLat + maxLat) / 2, (minLng + maxLng) / 2];
  };
  ```

### Nitpicks (Optional)

- **[ExploredOverlay.jsx:80]** `map.on('dragstart zoomstart', stopAutoFit)` — while the space-separated syntax works in Leaflet, the individual-event form (`map.on('dragstart', stopAutoFit); map.on('zoomstart', stopAutoFit);`) is slightly more explicit and mirrors the cleanup in the return function. Not blocking — just a readability preference.

- **[ExploredEntrancesMap.jsx:59]** `zoom={5}` is a reasonable initial zoom, but since `BoundsFitter` will override it almost immediately, consider adding a brief inline comment like `{/* placeholder until BoundsFitter fits */}` to clarify intent for future readers. Very minor.

---

Overall this is a well-reasoned fix. The root cause analysis is solid — the race between `fitBounds` and the container reaching its final size is a classic Leaflet-in-tabs pitfall. The `isProgrammaticRef` guard for distinguishing own moves from user moves (leveraging `animate: false` synchronicity) is a clean pattern. The dual-layer approach (initial center from data + BoundsFitter refinement) means the map never flashes at `[0, 0]`.

Nice work 👍

PR #1428, opened 2026-06-28
fix(descriptions): fix French gender agreement in "no description" message

### LINKED ISSUE #1425: Reprise de texte
Lorsqu'il y a une indication précisant qu'il n'y a pas description, le texte en Français est a reprendre

Image

Paul-AUB:
fix(descriptions): fix French gender agreement in "no description" message

## 🤔 What

Fix the French (and Spanish) translation of the "no description" alert in the `Descriptions` component so it correctly matches the grammatical gender of each entity type.

- Before: `"Il n'y a actuellement aucune description pour ce entrée."` ❌ (wrong gender + wrong article)
- After: `"Il n'y a actuellement aucune description pour cette entrée."` ✅

## 🤷‍♂️ Why

The previous i18n key `descriptions.none` used a single `{entityType}` placeholder, which forced a single article (`ce`) for all entity types. In French, the article depends on grammatical gender:

- `cette entrée` (feminine)
- `cette cavité` (feminine)
- `ce massif` (masculine)

Same issue existed in Spanish (`esta`/`este`).

Closes #1425.

## 🔍 How

Replaced the single `descriptions.none` key with three entity-specific keys:

- `descriptions.none.entrance`
- `descriptions.none.cave`
- `descriptions.none.massif`

The component now uses `` `descriptions.none.${entityType}` `` as the message id — no more runtime parameter injection for the entity name.

All 15 language files updated. Languages that didn't differentiate by entity type (ar, bg, ca, de, el, he, id, it, ja, nl, pt, ro) keep the same text for all three keys.

## 🔗 Related API PR

N/A

## 🧪 Testing

Navigate to an entrance, cave, or massif that has no description and verify the alert text is grammatically correct in French.

## 📸 Previews

Before (wrong):

image


urien (approved):
Merci @Paul-AUB , c'est parfait

ClemRz (approved):

- **[packages/web-app/public/lang/it.json, de.json, pt.json, nl.json]** The Italian, German, Portuguese, and Dutch translations currently use a short generic phrase (e.g., "Nessuna descrizione", "Keine Beschreibung") identical for all three keys. These languages also have grammatical gender. Consider providing full sentences matching the style of fr/es/en for a better user experience — though this is not blocking since the current translations are grammatically correct on their own (they just don't mention the entity type). This can be addressed in a follow-up.

### Nitpicks (Optional)

- **General** Nice, minimal change. The approach of splitting into entity-specific keys (`descriptions.none.{cave,entrance,massif}`) is cleaner than trying to handle grammatical gender through ICU message syntax. Good use of the constrained `entityType` prop to build the key dynamically.

PR #1431, opened 2026-06-28
fix(map): restore display of embedded maps

Paul-AUB:
# fix(map): restore display of embedded maps

## 🤔 What

- Fixes embedded maps (`wholePage={false}`) no longer rendering — e.g. the "Explored entrances" map on the personal/account page, as well as the Country, Region and Organization maps.
- Only the white placeholder area was shown, with no actual Leaflet map.

## 🤷‍♂️ Why

A recent change (#1429, _"adapt map height to adjacent content"_) replaced the map `Wrapper`'s definite `height: 400px` with `height: 100%`.

For embedded maps the parent container has no defined height, so:

- `height: 100%` on the `Wrapper` resolves to `auto`
- the inner Leaflet `<MapContainer>` (`height: 100%`) then resolves against an indefinite parent and **collapses to 0**
- only the white `min-height: 400px` box remains, with no visible map

The full-page map kept working because its `height: calc(100dvh …)` is a definite value.

## 🔍 How

In `MapContainer.jsx`, made the `Wrapper` a flex column so the layout height is driven reliably in both cases:

- `Wrapper`: added `display: flex; flex-direction: column;` (keeping `height: 100%; min-height: 400px`)
- inner `<MapContainer>`: `height: '100%'` → `flex: '1 1 auto', minHeight: 0`

Result:

- **Embedded map, no parent height** → flex column falls back to `min-height: 400px`, the map fills it.
- **Parent with a real height (or full-page)** → the `Wrapper` fills it and the map grows with it, preserving #1429's "adapt to adjacent content" goal.
- Consumers passing an explicit `style.height` (e.g. `MapMarkerSelector`, `MapLine`) still override via `...style`.

## 🔗 Related API PR

_None._

## 🧪 Testing

1. Open the personal/account page and check the "Explored entrances" map renders.
2. Check the Country, Region and Organization detail maps still render.
3. Check the full-page map (MapClusters) is unchanged.

## 📸 Previews

_To be added._

### DIFF
--- a/packages/web-app/src/components/common/Maps/common/MapContainer.jsx
+++ b/packages/web-app/src/components/common/Maps/common/MapContainer.jsx
@@ -15,6 +15,8 @@ const Wrapper = styled('div', {
   shouldForwardProp: prop => !prop.startsWith('$')
 })(
   ({ theme, $wholePage }) => `
+  display: flex;
+  flex-direction: column;
   width: 100%;
   height: 100%;
   min-height: 400px;
@@ -149,7 +151,7 @@ const CustomMapContainer = ({
   return (
     <Wrapper $wholePage={wholePage}>
       <MapContainer
-        style={{ height: '100%', width: '100%', ...style }}
+        style={{ flex: '1 1 auto', minHeight: 0, width: '100%', ...style }}
         wholePage={wholePage}
         center={center}
         zoom={zoom}


ClemRz (approved):

- **[MapContainer.jsx:153]** When consumers pass an explicit `style.height` (e.g., `MapMarkerSelector` passes `height: '40dvh'`), it ends up on the same element as `flex: '1 1 auto'`. In a flex column parent, the explicit `height` does override the flex sizing, so this works — but the intent is slightly opaque. Consider adding a brief inline comment explaining why `...style` at the end is intentional (it lets consumers override the flex sizing with a fixed height). This is purely a readability nit and not blocking.

### Nitpicks (Optional)

- **[Branch name]** Minor typo in the branch name: `fix/personnal-map-issue` → "personnal" should be "personal". Not blocking since it's ephemeral.

---

Clean, minimal fix. The flex column approach correctly solves the `height: 100%` collapse on embedded maps while preserving the full-page behavior from #1429. The `minHeight: 0` on the inner container is a nice touch to prevent the flex item's implicit min-height from interfering. LGTM.

PR #1440, opened 2026-07-02
feat(organization): improve UI/UX of responsible organizations for geo entities

Paul-AUB:
# feat(organization): improve UI/UX of responsible organizations for geo entities

## 🤔 What

Follow-up on #1385 that reworks the "Responsible organizations" section of a geographic entity (country / region / massif) and fixes several UI/UX issues. Closes #1436.

- Rename the section to **"Responsible organizations"** with an explanatory subtitle distinguishing managing organizations from all organizations active within the area.
- Restyle the main action button: wording **"Associate"**, secondary (orange) color, association icon (organization icon + link badge, matching the Documents "Associate" logic), moved into the section header.
- Rename the modal to **"Create or associate an organization"**; make the info hint reliable (see How); Cancel button switched to outlined.
- Display associated organizations as **cards**, reusing the shared `EntitiesList` / `OrganizationCard` (same component as the Person page), with responsive layout and a proper remove action. `OrganizationCard` now also handles the deleted/redirect state.
- Fold the section by default when empty (same logic as the Descriptions section) instead of showing an expanded empty state — applied to both "Responsible organizations" and "More information" (`StatisticsDataDashboard`).
- Fix systemic gutter/alignment issues surfaced while doing the above (see How).

## 🤷‍♂️ Why

The inline UI shipped in #1385 had several rough edges reported in #1436: confusing wording, a mis-styled/mis-placed action button, plain-text links instead of cards, and an always-expanded empty section. While aligning the dialog buttons, two systemic theme-level inconsistencies were found that affected the whole app, so they were fixed at the source rather than patched locally.

## 🔍 How

**Feature (issue #1436)**
- `OrganizationAssociation/index.jsx` now owns its `ScrollableContent` (title, subtitle, header action button, `defaultExpanded={!isEmpty}`) and renders organizations through `EntitiesList type="organization"` with `onItemRemove`. Country/Region/Massif drop their redundant `ScrollableContent` wrapper.
- `EntitiesListItem.jsx`: `OrganizationCard` extended to render deleted organizations (strikethrough + "Deleted" chip, link to successor when present).
- `AssociationForm.jsx`: the "a new organization will be created" hint no longer depends on the fuzzy search result (Typesense matches descriptions, so "no result" was unreliable). It is now derived purely from the selection state — shown whenever text is typed and no existing organization is selected. `selectedOrg` holds only real, existing organizations.

**Systemic alignment fixes (theme)**
- `grottoTheme.js`: `MuiFormControl` padding `4px` → `4px 0` (keep vertical rhythm, remove the horizontal inset that offset every field from its container gutter). Added a `MuiDialogActions` override so dialog buttons share the 24px gutter of `DialogContent` (MUI defaults them to 8px).
- `FormContainers.jsx`: `FormRow` gains a `gap` (the inter-field spacing previously relied on the removed `FormControl` horizontal padding); `FormActionRow` buttons no longer use hardcoded/mismatched margins so they align to the field gutter.
- `PublicationDatePicker.jsx`: removed a local `4px` compensation that is now handled globally.
- `SearchDocumentForm/index.jsx`: reordered the action buttons so the primary "Associate" is on the right (was inverted with "Reset").
- `StatisticsDataDashboard/index.jsx`: `defaultExpanded={hasData}` so the empty section stays folded.

## 🔗 Related API PR

_None — this is frontend only, relying on the endpoints already delivered with #1385._

## 🧪 Testing

1. On a country / region / massif page, open the "Responsible organizations" section.
2. As a logged-in user, click **Associate**: search an existing organization (rendered as a card once added) or type a new name (the info hint stays visible and stable while typing) and create it.
3. Remove an association from a card.
4. On an entity with no organizations, confirm the section is folded by default; same for "More information" when there is not enough data.
5. Check dialog button alignment (e.g. the associate modal) and the create/edit forms (Entrance, Document…) for correct field spacing and gutter alignment.

## 📸 Previews

_To be added._


ClemRz (commented):
## Review Summary

Well-structured PR that meaningfully improves the UX of the "Responsible organizations" section. The move from inline text links to `OrganizationCard` via `EntitiesList`, the `ScrollableContent` wrapping with `defaultExpanded={!isEmpty}`, and the systemic theme fixes are all solid choices that align with existing patterns. Good job reusing shared components rather than reinventing them.

A few items to consider before merge:

---

### 1. Dead translation keys left behind

The following i18n keys are no longer referenced in code after this PR:

- `"Set organization"`
- `"No existing organization found. A new one named \"{name}\" will be created."`
- `"View successor organization"`
- `"Add/Edit organization"`

They should be removed from all 15 language files to avoid dead weight and confusion for future translators. You can batch-remove them or leave it for a follow-up, but worth tracking.

---

### 2. `AssociationForm` info hint is now always visible

Previously the hint (`Alert severity="info"`) only appeared when `options.length === 0 && !isSearching`. Now it shows whenever `!selectedOrg && inputValue.trim().length > 0` — i.e., even while results are loading or after results are displayed.

This is by design per the PR description ("derived purely from the selection state"), but the UX implication is that a user who has typed text and sees a list of matching organizations also sees "…or a new one named X will be created" simultaneously. This might be fine (it's informational), but it could also be confusing since the user already has choices in front of them. Consider whether hiding it when `options.length > 0` would reduce noise — or at least verify this feels right in testing.

---

### 3. `MuiFormControl` padding change — broad blast radius

Changing `padding: '4px'` → `padding: '4px 0'` at the theme level affects **every** `FormControl` across the app. The PR accounts for the knock-on in `FormRow` (adding `gap`) and `PublicationDatePicker` (removing local compensation), but the app has 75+ forms.

This is explicitly called out in the description and it's the right fix (horizontal padding on `FormControl` shouldn't exist), but it warrants a quick visual scan of a few other form-heavy pages (Entrance edit, Document create, Cave edit) during testing to catch any spacing regression.

---

### 4. Minor: `MuiDialogActions` padding formula

```js
padding: `${paddingUnit}px ${paddingUnit * 3}px ${paddingUnit * 2}px`
```

With `paddingUnit = 8`, this gives `8px 24px 16px` — matching `DialogContent`'s horizontal gutter (24px) which is correct. Just noting that the top padding (8px) is less than the bottom (16px), which creates asymmetric spacing between the last content element and the buttons vs. buttons and dialog bottom edge. This is usually desirable (less whitespace above actions), but verify it looks balanced in modals with very little content.

---

### 5. `OrganizationCard` — `id` PropType broadened to `string | number`

Good catch that IDs can come as either type from the API. Just verify that `EntitiesList`'s `onItemRemove(e.id)` still works correctly downstream — the `organizations.find(o => o.id === id)` in `AssociationSection` does a `===` comparison, so if the API returns a number but `onItemRemove` receives it as a string (or vice versa), the find would fail silently. Worth a quick check.

---

### 6. Nit: `FormContainers.jsx` — unused styled import removed but `MuiButton` not aliased

The styled `Button` (with `margin: theme.spacing(2)`) was removed and replaced with direct `MuiButton` usage + flex `gap`. Clean approach. Just confirming `MuiButton` is already imported at the top of that file (it is, from the existing imports).

---

### Overall

Approve with the suggestion to clean up dead i18n keys (item 1) and verify the broad theme changes haven't caused visual regressions elsewhere (item 3). The logic changes in `AssociationForm` and `OrganizationCard` are sound and well-commented.

urien (approved):
Comme utilisateur tout est parfait, merci @Paul-AUB

PR #1441, opened 2026-07-02
fix(documents): allow submission without a file and align accepted file formats

Paul-AUB:
# fix(documents): allow submission without a file and align accepted file formats

## 🤔 What

Two related fixes on the document submission/edit form:

- **No file is mandatory anymore** (closes #1424). Any document type can now be
  submitted or edited without attaching a file — this notably unblocks changing
  a document's type (e.g. from "Topographic Data" to "Topographic Drawing")
  which previously kept the validate button greyed out.
- **A confirmation dialog** ("Submit without a file?") is shown when a
  file-expected type (Image, Topographic Drawing, Moving Image, Sound, Physical
  Object, Map, Dataset) is submitted with no file, so users don't forget by
  mistake.
- **The document type selector is now marked as required** (`*`).
- **The accepted file formats now match reality** (closes #1438):
  - the file picker's `accept` is built from file **extensions** too, not only
    MIME types — so backend-supported formats whose MIME type the browser
    can't resolve (e.g. `gpx`, and the Therion `th`/`th2`/`thconfig`/`lox`/`xvi`
    formats) are now actually selectable;
  - the Sound and Moving Image types no longer advertise formats the backend
    rejects (`aac`, `flac`, `wav`, `ogg`, `opus`, `m4a`, `aiff`, `wma`, `mkv`,
    `mov`, `ogv`, `webm`, `wmv`).

## 🤷‍♂️ Why

- #1424: a moderator could not correct a document's type because topographic
  documents were forced to have a file. The community agreed that no file should
  be mandatory for any type, with a warning instead of a hard block.
- #1438: the advertised list of accepted extensions did not match what the
  backend actually supports — `gpx` files are accepted by the backend but were
  not offered in the picker, while several audio/video formats were offered but
  rejected on upload.

## 🔍 How

- `Provider.js`: removed the `FILE_REQUIRED_TYPES` constraint from
  `checkFormValidation`. The licensing/authorization fields are now only
  required when at least one **non-deleted** file is present
  (`f.state !== IS_DELETED`), fixing the case where removing a file left the
  form unsubmittable.
- `index.jsx`: `onFormSubmit` opens a confirmation dialog when a file-expected
  type has no active file, using the same `IS_DELETED`-aware count.
- `documentTypeHelpers.js`: added `isFileExpected`; corrected the Sound and
  Moving Image entries of `DOCUMENT_TYPE_ACCEPT` to the backend-supported
  formats only.
- `AddFileForm/index.jsx`: the `accept` attribute now combines MIME types and
  dotted extensions.
- `DocumentTypeSelect.jsx`: added a required label above the type cards.
- i18n: added `No file attached — please check that it is not an oversight.`,
  `Submit without a file?`, `Submit anyway` (reusing `Cancel`) and translated
  them across all 14 language files.

## 🔗 Related API PR

None. All changes are frontend-only; the backend `file-formats` table already
contains `gpx` and the Therion formats.

## 🧪 Testing

1. Edit a "Topographic Data" document, change its type to "Topographic Drawing"
   without a file → the validate button stays enabled, and a confirmation dialog
   appears on submit.
2. Edit an Image document, remove its existing file → licensing fields
   disappear, the button stays enabled, and the confirmation dialog appears on
   submit.
3. On any document type, open the file picker → `gpx` (and Therion formats for
   Topographic Data) are now selectable.
4. For Sound / Moving Image, the picker only offers backend-supported formats.

## 📸 Previews

_Confirmation dialog when submitting without a file:_


ClemRz (approved):

- **[packages/web-app/src/utils/documentTypeHelpers.js:TOPOGRAPHIC_DATA config]** The PR description states that `gpx` files are now selectable for Topographic Data, but `gpx` is not present in the `extensions` array of `DOCUMENT_TYPE_ACCEPT[TOPOGRAPHIC_DATA]`. The new code correctly adds dotted extensions to the `accept` attribute, but since `acceptConfig?.extensions` takes priority over `backendExtensions`, `gpx` still won't appear in the Topographic Data picker. Consider adding `'gpx'` to the `TOPOGRAPHIC_DATA` extensions list (and an appropriate MIME entry like `application/gpx+xml`) if the intention is to make it selectable for that document type specifically.

- **[packages/web-app/src/components/appli/EntitiesForm/Document/index.jsx:~258]** There's a double blank line between the closing `</Dialog>` and the error messages block. Minor formatting inconsistency — one blank line would be cleaner.

- **[packages/web-app/src/components/appli/EntitiesForm/Document/index.jsx:Dialog]** Consider adding an `aria-describedby` to the `<Dialog>` pointing to the `<Alert>` content, so screen readers announce the warning message when the dialog opens. Currently the `DialogTitle` will be announced but the important warning context in `DialogContent` may be missed by assistive technology.

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/EntitiesForm/Document/formElements/DocumentTypeSelect.jsx:217]** `ml: 0.5` in the `sx` prop is fine (MUI multiplies by the spacing factor), but the project steering notes a preference for integer spacing values. `'4px'` would be an acceptable alternative if you want to be extra consistent with the convention, though this is harmless either way.

- **[packages/web-app/src/utils/documentTypeHelpers.js:TOPOGRAPHIC_DATA mime]** The `mime` field for `TOPOGRAPHIC_DATA` embeds dotted extensions directly (`.th,.th2,.thconfig,.lox,.dxf,.cxf,.dpt,.xyz,.xvi,.zip,.gz`) — this was a pre-existing workaround. Now that the new code in `AddFileForm` explicitly builds dotted extensions from the `extensions` array, these embedded extensions in the MIME string are redundant. Not a blocker, but a cleanup opportunity for a follow-up.

urien (approved):
C'est parfait,
Merci @Paul-AUB

PR #1442, opened 2026-07-02
feat(map): reveal network entrances on hover and refine map popups

Paul-AUB:
# feat(map): reveal network entrances on hover and refine map popups

## 🤔 What

Two related improvements to the general map:

- **Network → entrances association**: hovering (or tapping on touch) a network marker now reveals which entrances belong to it, even when the entrances layer is hidden. The overlay draws:
  - blue **spokes** from the network centroid to each of its entrances,
  - highlighted **entrance markers** (blue core + halo) that stand out from the brown caves,
  - a discreet **convex-hull footprint** of the network.
- **Map popup refresh** (all types — entrances, caves/networks, organizations):
  - compact title (info-card size) instead of an oversized `h5` heading,
  - title no longer collides with the Leaflet close button,
  - links aligned with the app palette (brown, `primary.main`) instead of the browser-default blue,
  - removed the coordinates row from entrance and organization popups (no added value there).

## 🤷‍♂️ Why

The API now returns each network's entrances, but nothing on the map showed which entrances made up a network — and the answer had to hold even when the user chose not to display the entrances layer. A hover/tap reveal is the standard, low-clutter pattern for this.

Along the way, the shared popups had UX issues surfaced during testing: a giant title, a title running under the `×`, and links in browser-default blue that clashed with the app's brown/orange palette (and, now, with blue being used for the network highlight).

## 🔍 How

- New `useNetworkHighlight` hook manages a temporary Leaflet `LayerGroup`; the convex hull is computed with `d3-polygon`'s `polygonHull`. No extra API call — it uses the `entrances[]` already returned per network.
- `useMarkers` gained optional `onMarkerOver/Out` and `onPopupOpen/Close` handlers (all default `null`, backward-compatible with its other callers). `Markers.jsx` wires them: hover shows the highlight, and it stays pinned while a popup is open so touch devices (no hover) work via tap.
- Highlight styles live in `MapClusters/constants.js`; blue is the app's network semantic (networks heatmap).
- Popup title styled via a global `.map-popup-title` class in `App.css` (scoped specificity beats MUI's generated class) rather than `sx`, because popups are serialized with `renderToString` into detached HTML where emotion's hashed classes aren't injected. Same reasoning for the brown link color rule.
- Coordinates rows removed from `EntrancePopup`/`OrganizationPopup` (still reachable via the map right-click menu and the entity detail pages).

## 🔗 Related API PR

https://github.com/GrottoCenter/grottocenter-api/pull/1712

## 🧪 Testing

1. Open the general map and zoom in until network markers appear.
2. Hover a network marker (desktop) → spokes + blue entrances + faint hull appear, including entrances even if the entrances layer is toggled off.
3. Tap a network marker (mobile/touch) → same reveal while the popup is open.
4. Open entrance/cave/organization popups → compact title, no overlap with the `×`, brown links, no coordinates row.

## 📸 Previews

image
image
image
image
image

### DIFF
--- a/packages/web-app/src/App.css
+++ b/packages/web-app/src/App.css
@@ -72,6 +72,35 @@ code {
   margin: 2px 0;
 }
 
+/*
+ * Compact popup title (map info-card style, not an h-heading).
+ * Scoped with the parent so specificity (0,2,0) reliably beats MUI's single
+ * generated font-size class (0,1,0), regardless of stylesheet injection order.
+ * rem keeps it responsive to the user's root font-size (accessibility).
+ */
+.map-popup-property .map-popup-title {
+  font-size: 1.7rem;
+  font-weight: 600;
+  line-height: 1.3;
+  /* MUI renders subtitle1 as an <h6>, which inherits Skeleton's global
+     `h1..h6 { margin-bottom: 2rem }`. Override it here (scoped) rather than
+     changing the global heading margin app-wide. */
+  margin-bottom: 1rem;
+  /* Reserve room for Leaflet's absolutely-positioned close button (top-right),
+     so a long title wraps instead of running under the × */
+  padding-right: 5px;
+}
+
+/*
+ * Popup links (GCLink renders a bare <a> with no app styling, so in the
+ * detached popup HTML it falls back to the browser-default blue). Align with
+ * the app convention: MUI <Link> defaults to palette.primary.main (brown[700]).
+ * Also avoids clashing with blue, which now means "network highlight".
+ */
+.leaflet-popup-content a {
+  color: #5d4037;
+}
+
 .map-popup-property > img,
 .map-popup-property > svg {
   margin-right: 0;
--- a/packages/web-app/src/components/common/Maps/MapClusters/Markers.jsx
+++ b/packages/web-app/src/components/common/Maps/MapClusters/Markers.jsx
@@ -42,10 +42,7 @@ const Markers = ({
   // no hover), and only hide on mouseout when nothing is pinned.
   const pinnedNetworkRef = useRef(null);
 
-  const handleNetworkOver = useCallback(
-    network => showHighlight(network),
-    [showHighlight]
-  );
+  const handleNetworkOver = showHighlight;
   const handleNetworkOut = useCallback(() => {
     if (pinnedNetworkRef.current == null) hideHighlight();
   }, [hideHighlight]);
--- a/packages/web-app/src/components/common/Maps/MapClusters/Markers.jsx
+++ b/packages/web-app/src/components/common/Maps/MapClusters/Markers.jsx
@@ -1,8 +1,9 @@
-import React, { useEffect } from 'react';
+import React, { useCallback, useEffect, useRef } from 'react';
 import { includes, values } from 'ramda';
 import PropTypes from 'prop-types';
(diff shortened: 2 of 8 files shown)


ClemRz (approved):

- **[useNetworkHighlight.js:43]** When the network marker itself has no valid coordinates (e.g., `network.latitude` is `undefined`), the spokes will be drawn from `[undefined, undefined]`. The `hasCoords` guard only filters entrances but not the network centroid. Consider guarding early:
  ```javascript
  const center = [network.latitude, network.longitude];
  if (typeof network.latitude !== 'number' || typeof network.longitude !== 'number') return;
  ```

- **[useMarkers.js:65-68]** Event listeners (`mouseover`, `mouseout`, `popupopen`, `popupclose`) are attached to Leaflet markers but never explicitly removed — cleanup relies on `markerEl.remove()` disposing the element. This is fine for Leaflet (it cleans up listeners on remove), but if `createLeafletMarker` ever gets a new identity due to callback reference changes (e.g., during a re-render), existing markers in `markersRef` won't get their old listeners replaced since `updateMarkers` only adds *new* markers and removes *gone* ones. Currently safe because the `useCallback` deps are stable, but worth a code comment noting this assumption for future maintainers.

- **[Markers.jsx:49-53]** `handleNetworkOver` creates a new closure with `network => showHighlight(network)` — this is equivalent to just passing `showHighlight` directly. Consider simplifying to avoid the extra wrapper:
  ```javascript
  const handleNetworkOver = showHighlight;
  ```
  However, if the intent is to keep the signature explicit for readability or future extension, the current form is acceptable — just a minor style note.

- **[useNetworkHighlight.js:59-75]** Three separate `forEach` loops over the same `entrances` array is intentional (z-order layering: spokes → halos → cores). This reads well and the comment explains the why. For networks with many entrances (50+), a single pass with explicit `bringToFront()` calls could reduce iterations, but given the overlay is on-demand and networks rarely have dozens of entrances, this is fine for now.

### Nitpicks (Optional)

- **[App.css:97]** The hardcoded `color: #5d4037` for `.leaflet-popup-content a` works, but if the theme's `primary.main` ever changes, this will drift. A code comment referencing `brown[700]` (which it corresponds to) could help — though given it's already in a global CSS file divorced from the theme, this is inherent to the approach and the comment already explains it well.

- **[App.css:83]** `font-size: 1.7rem` — with the Skeleton base of `html { font-size: 62.5% }`, this maps to ~17px. That's reasonable for a popup title but sits between the Skeleton `h5` (18px) and `h6` (15px). No issue, just noting the effective computed size in case it's unintentionally larger than expected.

- **[OrganizationPopup.jsx]** After removing the `makeCoordinatesValue` import, the `latitude` and `longitude` fields remain in the `propTypes` definition (lines 27-28). Since they're no longer used by the component, consider removing them from propTypes to keep the contract accurate. Same applies to `EntrancePopup.jsx` (lines 64-65).

- **[utils.js:7-8]** `makeCoordinatesValue` is still exported but no longer imported anywhere in the Maps popup components. It *is* still used in snapshot comparison components (separate folder), so it can stay. Just noting it's not dead code.

Paul-AUB:
@urien Déployé en staging, peut être testé pour info.

urien (approved):
C'est une super fonctionnalité. Merci @Paul-AUB

PR #1449, opened 2026-07-06
feat(snapshot): manage renamed entrances

Paul-AUB:
# feat(snapshot): manage renamed entrances

## 🤔 What

- Display the cave name (`caveName`) in the entrance revision history timeline, instead of the entrance name, since the cave name is the value that can change over time for network entrances.
- Add a distinct visual marker (chip + timeline dot color) for synthetic rename snapshots (`isNameChangeSnapshot`).
- Rename the "Cave" label to "Network" in the entrance network snapshot body, since that property is only shown in network context.
- Fix a responsive layout issue on the timeline entry header (meta column and title/chips column now stack on mobile instead of overlapping/truncating).
- Add i18n keys (`Renamed`) across all supported languages.

Closes #1446

## 🤷‍♂️ Why

The backend (grottocenter-api PR #1718) now resolves the cave name temporally per snapshot, since a cave can be renamed independently of its entrances. The frontend needs to surface that per-snapshot `caveName` and clearly flag the snapshot where the rename happened, instead of always showing the (unchanged) entrance name.

## 🔍 How

- `AccordionSnapshot.jsx`: for `entrances` snapshots, the header title/diff now resolves from `caveName` (falling back to `name`/`title` if absent, so `null` values are handled gracefully) instead of always using the entrance name. Added a `Renamed` chip + secondary timeline dot color when `snapshot.isNameChangeSnapshot` is `true`. Reworked the header layout to stack responsively on small screens (`flexDirection: { xs: 'column', sm: 'row' }`) and reclaimed unused right-hand padding on `TimelineContent`.
- `EntranceNetworkSnapshots.jsx`: renamed the displayed label from `Cave` to `Network`.
- Added the `Renamed` translation key to `en.json` and all other language files, kept sorted and in sync via `yarn translations:sync-with-en` / `yarn translations:sort`.

## 🔗 Related API PR

grottocenter-api #1718

## 🧪 Testing

- `yarn lint` passes on the changed files.
- Manually verified in the running dev app (mocked API responses) that:
  - the cave name diff displays correctly across current/rename/initial timeline entries,
  - a `null` `caveName` falls back gracefully instead of showing "null",
  - the `Renamed` chip and marker only appear on `isNameChangeSnapshot` entries,
  - the header layout no longer overlaps/truncates on mobile widths.

## 📸 Previews

_Screenshots available on request — not attached to keep the PR focused on the diff._

### DIFF
--- a/packages/web-app/public/lang/ar.json
+++ b/packages/web-app/public/lang/ar.json
@@ -2040,6 +2040,7 @@
   "Remove sensitivity": "إزالة الحساسية",
   "Removing sensitivity from the massif will not automatically remove the sensitivity flag from individual entrances.": "إزالة الحساسية عن الكتلة الجبلية لن يؤدي تلقائياً إلى إزالة علامة الحساسية عن المداخل الفردية.",
   "Removing sensitivity from the massif will not automatically remove the sensitivity flag from individual entrances. Those flags must be managed separately. Do you want to continue?": "إزالة الحساسية عن الكتلة الجبلية لن يؤدي تلقائياً إلى إزالة علامة الحساسية عن المداخل الفردية. يجب إدارة هذه العلامات بشكل منفصل. هل تريد الاستمرار؟",
+  "Renamed": "أُعيد التسمية",
   "Repeat your password here.": "أعد كتابة كلمة المرور هنا.",
   "Report": "تقرير",
   "Report message": "الإبلاغ عن الرسالة",
--- a/packages/web-app/public/lang/bg.json
+++ b/packages/web-app/public/lang/bg.json
@@ -2040,6 +2040,7 @@
   "Remove sensitivity": "Премахване на чувствителност",
   "Removing sensitivity from the massif will not automatically remove the sensitivity flag from individual entrances.": "Премахването на чувствителността от масива няма автоматично да премахне флага за чувствителност от отделните входове.",
   "Removing sensitivity from the massif will not automatically remove the sensitivity flag from individual entrances. Those flags must be managed separately. Do you want to continue?": "Премахването на чувствителността от масива няма автоматично да премахне флага за чувствителност от отделните входове. Тези флагове трябва да се управляват отделно. Искате ли да продължите?",
+  "Renamed": "Преименувано",
   "Repeat your password here.": "Повторете паролата си тук.",
   "Report": "Доклад",
   "Report message": "Докладване на съобщение",
--- a/packages/web-app/public/lang/ca.json
+++ b/packages/web-app/public/lang/ca.json
@@ -2040,6 +2040,7 @@
   "Remove sensitivity": "Elimina la sensibilitat",
   "Removing sensitivity from the massif will not automatically remove the sensitivity flag from individual entrances.": "L'eliminació de la sensibilitat del massís no eliminarà automàticament la marca de sensibilitat de les entrades individuals.",
   "Removing sensitivity from the massif will not automatically remove the sensitivity flag from individual entrances. Those flags must be managed separately. Do you want to continue?": "L'eliminació de la sensibilitat del massís no eliminarà automàticament la marca de sensibilitat de les entrades individuals. Aquestes marques s'han de gestionar per separat. Vols continuar?",
+  "Renamed": "Reanomenat",
   "Repeat your password here.": "Repetiu aquí la vostra contrasenya.",
   "Report": "Informe",
   "Report message": "Denunciar missatge",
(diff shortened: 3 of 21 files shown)


ClemRz (changes requested):
### Issues (Must Fix)

- **[AccordionSnapshot.jsx:~195-201]** The "Restore" button is rendered for all non-current, non-document snapshots — including name-change snapshots (`isNameChangeSnapshot: true`). Since these synthetic entries have `null` for `latitude`, `longitude`, `altitude`, and `cave`, restoring one would dispatch an `updateEntrance` / `updateCaveAndEntrance` action with null coordinates and cave data, likely corrupting the entrance record or crashing. Consider hiding the restore button for name-change snapshots:
  ```suggestion
  {!isCurrent && snapshotType !== 'documents' && !isNameChange && (
  ```

- **[AccordionSnapshot.jsx:37-49]** For `snapshotTitle`, the fallback chain for the `entrances` type starts with `snapshot.caveName`. But `caveName` can be `null` for name-change snapshots where the cave name couldn't be resolved (as stated in the issue). In that scenario, the nullish coalescing correctly falls through to `snapshot.name`. However, for regular entrance snapshots, `caveName` will also be present and will take priority over `name`/`title`. If a regular snapshot has *both* `caveName` and `name`, the title will show the cave name rather than the entrance name — which may be the intended behavior but could be confusing when viewing an entrance-only (non-network) timeline. Consider whether this fallback should only apply to network entrances or always. A brief inline comment clarifying the intent would help future readers.

- **[EntranceNetworkSnapshots.jsx / EntranceCaveSnapshots.jsx]** For name-change snapshots, `latitude` and `longitude` are `null`. `Number(null)` evaluates to `0` (not `NaN`), so the coordinate guard `!(Number.isNaN(lat) && Number.isNaN(long))` passes and the component would render `0.0000, 0.0000`. This isn't introduced by this PR, but since it directly affects how name-change snapshots render in the collapsed body, it might be worth addressing here or in a follow-up.

### Nitpicks (Optional)

- **[AccordionSnapshot.jsx:36]** `const isNameChange = snapshot.isNameChangeSnapshot === true;` — the `=== true` is redundant given the PropType is `bool`. `!!snapshot.isNameChangeSnapshot` or just the truthy check would be equivalent and more idiomatic. Minor style preference.

- **[AccordionSnapshot.jsx:71]** `sx={{ pb: 2, pt: 0, pr: 0 }}` — removing right padding on `TimelineContent` is mentioned as "reclaiming unused right-hand padding" in the PR description. This could affect other snapshot types beyond entrances. If it looks fine in all contexts, no issue — just flagging it since it applies globally.

ClemRz (approved):

- **[AccordionSnapshotListPage.jsx:39-47]** The `previousMap` and `latestByGroup` computation doesn't account for name-change snapshots. If a rename snapshot happens to be the last item in a group, `latestByGroup[key]` will point to it — which means the current item's `previous` will be a rename snapshot (with `null` coordinates, cave, etc.). Similarly, a rename snapshot could become `previous` for the next real snapshot in `prevMap`. The fix in `AccordionSnapshotList.jsx` correctly skips rename snapshots for `prev` tracking, but the page variant doesn't have the same logic. Consider filtering rename snapshots from the `previous` chain here too, or adding a comment noting this is a known gap for a follow-up.

- **[AccordionSnapshotListPage.jsx:96-109]** The page variant doesn't pass `newName` to `AccordionSnapshot` for rename entries. This means on the "all snapshots" page, rename snapshots will render with `HighLightsChar oldText={snapshotTitle} newText={undefined}` — which shows the old name as plain text without highlighting the change. Since `AccordionSnapshotList` properly resolves `newName` from the next real snapshot, this page variant should do the same for consistent UX. Could be addressed here or in a follow-up.

- **[AccordionSnapshot.jsx:44-48]** The comment says "prioritize the cave/network name" but the fallback chain starts with `snapshot.name` (not `snapshot.caveName`). The PR body mentions that `caveName` should be displayed for entrance snapshots, but the code uses `snapshot.name` first. If the intent is to show the entrance's own name (since renames are handled by dedicated snapshots), the code is correct but the comment is misleading. Consider updating the comment to match the actual behavior — something like "For entrance snapshots, use the entrance name directly; cave/network renames are handled by dedicated rename snapshots."

### Nitpicks (Optional)

- **[AccordionSnapshot.jsx:40]** `const isEntranceType = snapshotType === 'entrances';` — this variable is clear but only used for title resolution and one conditional branch. Minor naming nit: `isEntranceSnapshot` might read slightly more naturally. No action needed.

- **[EntranceNetworkSnapshots.jsx:56]** `value={entrance.caveName}` — if `caveName` is `null` (the API can return it as such per the issue spec), this renders `null` as the property value. The `Property` component might handle that gracefully, but worth double-checking that it doesn't render the text "null" or an empty field without explanation.

urien (approved):
Tout fonctionne

Merci @Paul-AUB

PR #1451, opened 2026-07-07
feat(entrance): improve entrance creation and move pages

Paul-AUB:
## 🤔 What

Rework of the **entrance creation** form and the **move / detach entrance** pages, plus the extraction of shared form building blocks and hooks.

### Entrance creation & edition form

- Reorganized the form into clearer, self-contained sections: entrance identity (name, language), **physical characteristics** (coordinates + map positioning, altitude, discovery year), **attributes & hazards**, and **network** linking.
- Replaced the old up-front "The entrance is: the first entrance of a new cave / linked to an existing entrance or network" radio choice with a simpler flow: the user describes the entrance, then **optionally** opts into linking via a single checkbox.
- **Network linking (create mode)** — a checkbox reveals a cave/network search with a live outcome preview:
  - linking to a **single entrance** creates a **network of 2 entrances**;
  - linking to an **existing network** **extends** it (message shows the resulting entrance count).
- **Network membership (edit mode)** — a dedicated section shows whether the entrance is standalone or part of a network of N entrances, with a link to the network/cave and the detach/move actions.
- Cave characteristics shared by a network (depth, length, temperature, diving…) are shown as **locked/read-only** when the entrance belongs to a network, with an explanation that they are shared by all its entrances.
- **Attributes & hazards** are now compact toggle chips (bat presence, flooding / CO₂ / rockfall / pollution danger, clean-gear & stay-on-trail requirements, rules, touristic, sensitive access).

### Move / detach entrance pages

- The page is now **single-action**, driven by a `?mode=move|detach` query param deep-linked from the two edit-page buttons (no redundant in-page mode picker).
- A discreet inline link lets the user switch to the other operation ("Rather detach the entrance?" / "Rather link to an entrance or network?"); detach is only offered for a networked entrance.
- Reworked `Header`, `FormActions`, `DetachEntranceSection` and a much richer `OperationSummary` showing a **before → after** preview of the operation.
- Kept the safety warning when moving a **solo** entrance (its cave and cave-level data would be deleted).

### Shared components & hooks

- New reusable components with Storybook stories: `BoolToggleChip`, `NumberField`, `NetworkInlineLink`, `LinkOutcomeAlert`.
- New hooks: `useEntitySearch` (cave/network autocomplete search) and `useOtherEntranceName`.
- i18n keys added/renamed/removed across all 15 language files.

## 🤷‍♂️ Why

The previous entrance-creation flow forced a confusing decision **before** the user had entered anything: they had to declare whether the entrance was the first of a brand-new cave or linked to something existing, using vocabulary ("entrance / cave / network") that is hard to grasp up front. This reworks the experience so the user first positions and describes the entrance, and network linking becomes an **optional, explained** step with a clear preview of what will happen.

The move/detach pages suffered from the same lack of clarity: it wasn't obvious what a move or a detach would actually produce, nor when detaching even made sense. Aligning them on the same components and adding a before → after summary makes the outcome explicit and reduces the risk of destructive mistakes (e.g. moving a solo entrance and losing its cave data).

Extracting the repeated pieces (toggle chips, number fields, network links, search logic) into shared components/hooks removes duplication between the create, edit and move flows and makes them consistent.

## 🔍 How

- Split `EntranceDetail` / `CaveDetail` into focused sub-sections: `EntranceAttributes`, `NetworkLinkSection` (create mode), `NetworkMembershipSection` (edit mode), plus the coordinate/map selectors.
- `EntranceForm` keeps `entityType` (`ENTRANCE_ONLY` vs `ENTRANCE_AND_CAVE`) as the single source of truth for whether a network link is active, and clears any previously-selected network's cave data on every toggle to avoid stale "shared network" values lingering in the form.
- The searched network's name is held in parent state (not the shared `cave.name` RHF field) so it doesn't bleed into the entrance/cave name when switching back to "new cave" mode.
- Removed the old `EditTypeSelection` decision tree; simplified it to reflect the new flow.
- Extracted the cave/network autocomplete into `useEntitySearch` and reused it in both `CaveSelect` and `CaveAutoCompleteSearch`.
- `MoveEntranceToCaveForm` derives its mode from the URL, computes `sourceInNetwork` / `isSameCave` / `targetNbEntrances`, and renders either the move (search + summary + confirm) or the detach section.
- Added/updated unit tests for `MoveEntranceToCaveForm`, `DetachEntranceSection` and the new `OperationSummary`.

## 🔗 Related API PR

_None._

### Create an entrance

1. Open the entrance creation form. The submit button stays disabled until name, language **and** coordinates are provided.
2. Position the entrance by searching or clicking on the map; check latitude/longitude/altitude fields update accordingly.
3. Fill physical characteristics and toggle a few attributes/hazards chips.
4. **Without** linking: submit → a new cave + entrance is created and you're redirected to the entrance page.
5. **Link to a single entrance/cave**: check the network box, search and select an entrance → the outcome alert announces a **network of 2 entrances**; submit and verify the network is created.
6. **Link to an existing network**: select a network → the alert announces the network extended to N entrances; verify shared cave characteristics are shown locked. Submit and verify the entrance joined the network.
7. Toggle the network checkbox on/off and change the selection several times → confirm no stale cave data (depth/length/temperature) or stale network link remains.

### Edit an entrance

- Open a standalone entrance → membership section shows it's the only entrance of its cavity.
- Open an entrance belonging to a network → membership section shows the network, its entrance count, a link to it, and shared characteristics are locked.

### Move / detach

- From the edit page, use the move button → `?mode=move`: search a target cave/network, check the before → after summary, and the warning when the source is a **solo** entrance. Selecting the same cave shows the "select a different cave" error and keeps the confirm disabled.
- Use the detach button → `?mode=detach`: verify the detach summary and that detaching is only offered for a networked entrance.
- Use the inline switch link to toggle between move and detach.

### Automated

- `yarn lint`
- Unit tests: `MoveEntranceToCaveForm`, `DetachEntranceSection`, `OperationSummary`.

## 📸 Previews
image
image
image

image

image

image

image


ClemRz (approved):

- **[src/hooks/useOtherEntranceName.js:23-37]** The `useEffect` has no cleanup return when `enabled && caveId` is truthy and the fetch fires. This means if the component unmounts while the fetch is in-flight, the `.then()` will still call `setName()` on an unmounted component. The `requestIdRef` pattern handles *stale* responses (where the effect re-runs), but not unmounts. Consider adding an explicit cleanup return analogous to `useEntitySearch`:
  ```javascript
  return () => {
    requestIdRef.current += 1;
  };
  ```
  after the `fetch(...)` block (outside the `.then`), so stale-on-unmount is also covered.

- **[src/components/appli/MoveEntranceToCave/MoveEntranceToCaveForm.jsx:34-58]** The success navigation `useEffect` fires when `hasSubmitted && !loading && !apiError`. However, `hasSubmitted` is never reset to `false`. If the move fails (apiError is set), then the user fixes something and the error is cleared upstream (e.g., the reducer resets on a new dispatch), the effect would fire without a new submission. Consider resetting `hasSubmitted` when the apiError changes from a non-null value back to falsy, or toggling it off inside the success branch.

- **[src/components/appli/EntitiesForm/Entrance/CaveDetail.jsx:20-32]** The `validateTemperature` and `validateDistance` functions are recreated on every render. Since `formatMessage` is stable across renders, consider wrapping them in `useCallback` or extracting them outside the component (with `formatMessage` passed in) to avoid unnecessary re-validations when the parent re-renders. Not urgent, but good practice for a form that re-renders often.

- **[src/components/appli/EntitiesForm/Entrance/NetworkLinkSection.jsx:36]** When the user unchecks the network checkbox, `reset()` is called (which resets form validation errors) *after* `updateEntityType(ENTRANCE_AND_CAVE)`. However `updateEntityType` also calls `reset(...)` internally (in the parent's `handleUpdateEntityType`). This means `reset` is called twice in succession — once in the parent (clearing cave.id/depth/length/temperature/isDiving) and once here (resetting errors). The second call to `reset()` with no arguments may undo the parent's reset (resetting to `defaultValues` rather than the cleared values the parent just set). Consider removing the `reset()` call from this component and letting the parent handle all resets via `handleUpdateEntityType`.

- **[src/components/common/NetworkInlineLink.jsx]** The component uses `role="button"` on a `<span>` to stay inline. Consider adding `aria-label` or ensuring the rendered text is descriptive enough for screen readers. As-is, the text content (the network name) serves as the accessible name, which is fine. But if `label` can be empty or just an icon fallback, the element would have no accessible name.

### Nitpicks (Optional)

- **[src/hooks/useEntitySearch.js:78]** The `eslint-disable-next-line react-hooks/exhaustive-deps` is well-justified in the comment above. Just noting it's fine.

- **[src/components/appli/EntitiesForm/utils/NumberField.jsx:38-56]** `InputProps` and `InputLabelProps` are deprecated in MUI 7 in favor of `slotProps.input` / `slotProps.inputLabel`. The rest of the codebase still uses the legacy API, so no change needed now, but worth keeping in mind for a future cleanup pass.

- **[src/components/appli/MoveEntranceToCave/Header.jsx]** The `alt` text on the entrance icon (`CustomIcon` doesn't have explicit alt text) combined with just the entrance name means screen readers might not announce the context. Consider adding an `aria-label` on the link like `Move entrance: {name}`.

- **[src/components/appli/EntitiesForm/utils/BoolToggleChip.jsx:27-28]** `px: 1.5` and `borderRadius: 2` — `px` is `theme.spacing(1.5)` which uses a non-integer value. Per project conventions, only integer values should be used with `theme.spacing()`. However, `sx` shorthand (`px: 1.5`) maps to `theme.spacing(1.5)` = `12px` which is a valid pixel value — the convention in the steering file specifically calls out `theme.spacing(1.5)` as the pattern to avoid. This might be considered equivalent, or it might be OK since the `sx` shorthand is the idiomatic MUI approach. Up to you whether to enforce it or accept it.

- **[Translation files]** The key `"Attributes & hazards"` is added but the `EntranceAttributes` component uses `"Hazards & restrictions"` as its `FormSection` title. `CaveDetail` uses `"Characteristics"`. The `"Attributes & hazards"` key doesn't appear to be used in the new code — it might be intended for UI elsewhere or might be orphaned. Worth confirming it's referenced somewhere.

urien (approved):
C'est parfait @Paul-AUB , ce sera beaucoup plus clair pour les utilisateurs de Grottocenter.
Pas d'autre retour que celui de Bastien qui évoque un point a traiter ulétrieurement

PR #1452, opened 2026-07-07
feat(documents): use responsive image thumbnails from API

Paul-AUB:
# feat(documents): use responsive image thumbnails from API

## 🤔 What

- Consume the new `thumbnails` object (`small`/`medium`/`large` WebP variants) returned by the API on document file resources, in addition to the existing `completePath`.
- Render responsive `srcSet`/`sizes` on thumbnail cards (`ImageThumbnail`) instead of always loading the full-resolution image.
- Use the `large` variant in the full-screen lightbox (`ImageLightbox`), while keeping `completePath` for downloads (always full resolution).
- Gracefully fall back to `completePath` whenever `thumbnails` or an individual variant is `null` (non-image files, failed generation, uploads predating the backfill, or images smaller than a given variant's target width).

Closes #1447

## 🤷‍♂️ Why

Document-heavy pages (document detail, cave/entrance/massif "associated documents" sections, document lists) were always loading full-resolution images even for small thumbnail cards, which is wasteful — especially on mobile. The API now generates responsive WebP variants specifically to fix this.

## 🔍 How

- `types/document.type.js`: added a shared `ThumbnailsPropTypes` shape (`{small, medium, large}`, all optional/nullable) and added `thumbnails` to the `files` prop type.
- `components/common/DocumentsList/utils/imageUtils.js`: two new pure helpers reused everywhere images are rendered:
  - `getThumbnailSources(file)` → `{src, srcSet}`, building `srcSet` only from non-null variants and falling back to `completePath`.
  - `getLightboxSrc(file)` → `large` variant, falling back to `completePath`.
- `components/common/DocumentsList/ImageThumbnail.jsx`: accepts optional `srcSet`/`sizes` props, with a `sizes` default that mirrors its own `ThumbnailCard` breakpoint (100vw below `sm`, fixed 240px from `sm` up) — so the value lives next to the CSS it describes instead of being duplicated by callers.
- `components/common/DocumentsList/ImageLightbox.jsx`: full-screen image now uses `getLightboxSrc`; the download button still uses `completePath`.
- `components/common/DocumentsList/Files.jsx` and `pages/DocumentDetails/Section.jsx`: both thumbnail grids now go through `getThumbnailSources`.

No Redux/action/reducer changes were needed — API responses are stored as-is, so `thumbnails` reaches components without any normalization step.

## 🔗 Related API PR

https://github.com/GrottoCenter/grottocenter-api/pull/1719

## 🧪 Testing

- `yarn lint` passes.
- Manually verified in the browser Network tab that thumbnail cards request `small`/`medium` WebP variants (not `completePath`) and that resizing the viewport changes which variant loads.
- Verified the lightbox loads the `large` variant and that the download button still fetches `completePath`.
- Verified fallback behavior with files where `thumbnails` is `null` and where only some variants are `null`.

## 📸 Previews

_N/A — no visual change at the rendered size/quality, only network payload size changes._

### DIFF
--- a/packages/web-app/src/components/common/DocumentsList/Files.jsx
+++ b/packages/web-app/src/components/common/DocumentsList/Files.jsx
@@ -4,8 +4,13 @@ import { Button, ListItem, Grid } from '@mui/material';
 import { styled } from '@mui/material/styles';
 import ImageThumbnail from './ImageThumbnail';
 import ImageLightbox from './ImageLightbox';
-import { isImageFile, decodeFileName } from './utils/imageUtils';
+import {
+  isImageFile,
+  decodeFileName,
+  getThumbnailSources
+} from './utils/imageUtils';
 import { getFileIcon } from './utils/fileIcons';
+import { ThumbnailsPropTypes } from '../../../types/document.type';
 
 const FileListItem = styled(ListItem)`
   margin: 0;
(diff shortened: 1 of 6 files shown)


ClemRz (approved):

- **[packages/web-app/src/pages/DocumentDetails/Section.jsx:440-448]** `FilesSection.propTypes` doesn't include `thumbnails` in its file shape, even though the component now calls `getThumbnailSources(file)` which reads `file.thumbnails`. For consistency with `Files.jsx` (which was updated to include it), consider adding it here too:
  ```javascript
  FilesSection.propTypes = {
    files: PropTypes.arrayOf(
      PropTypes.shape({
        fileName: PropTypes.string,
        completePath: PropTypes.string,
        description: PropTypes.string,
        thumbnails: ThumbnailsPropTypes
      })
    )
  };
  ```
  (You'd also need to import `ThumbnailsPropTypes` from `'../../types/document.type'`.)

### Nitpicks (Optional)

- **[packages/web-app/src/components/common/DocumentsList/utils/imageUtils.js:113]** The `getThumbnailSources` JSDoc return type could document `sizes` since the issue originally envisioned returning it from the helper. Currently `sizes` lives in `ImageThumbnail` as a default prop, which is arguably a better design (collocated with the CSS it describes), so this is purely a documentation nit — no code change needed.

---

Overall this is a well-structured PR. The fallback logic is clean, the `srcSet`/`sizes` approach is correct, and the separation between thumbnail sources (helper) and sizing policy (component default) keeps things maintainable. The download button correctly keeps using `completePath` for full-resolution files.

urien (approved):
Tout fonctionne sur une connexion très limitée... Sauf quand je demande la photo d'origine... Elle arrive mais c'est long
Vous allez faire gagner de la bande passante @tout le monde.

PR #1454, opened 2026-07-10
fix(quick-search): do not propose single-entrance caves alongside their entrance

Paul-AUB:
# fix(quick-search): do not propose single-entrance caves alongside their entrance

## 🤔 What

Filter the quick search results so a single-entrance cavity is no longer listed twice.

- Entrances are always kept.
- Caves are kept only when they are networks (`nbEntrances > 1`).
- Other entities (organizations, massifs) are untouched.

## 🤷‍♂️ Why

A cavity with a single entrance was showing up twice in the quick search dropdown: once as its cave and once as its entrance. Since the two entries point to the same physical cavity, this is redundant and confusing. We only want to propose networks or entrances.

## 🔍 How

Added a memoized `filteredResults` in `QuickSearch.jsx` that removes cave results whose `nbEntrances` is not greater than 1, before passing the suggestions to the autocomplete. The filter is recomputed only when the server results change.

## 🔗 Related API PR

_None._

## 🧪 Testing

1. Open the quick search.
2. Type the name of a cavity that has a single entrance (e.g. "Sarrasins (Sialet des)").
3. Verify only the entrance is proposed, not a duplicate cave entry.
4. Type the name of a network (2+ entrances) and verify it still appears as a network.

## 📸 Previews

Before (bug) 
image

After (fixed)
image

### DIFF
--- a/packages/web-app/src/components/appli/QuickSearch.jsx
+++ b/packages/web-app/src/components/appli/QuickSearch.jsx
@@ -26,8 +26,9 @@ const QuickSearch = ({ hasFixWidth, onClose }) => {
   const debouncedInput = useDebounce(input, AUTOCOMPLETE_DEBOUNCE_DELAY);
 
   // A single-entrance cavity would appear twice (once as its cave, once as its
-  // entrance). Keep only networks (caves with 2+ entrances) and entrances so the
-  // list shows networks or entrances, never a cave redundant with its entrance.
+  // entrance). Keep only networks (caves with 2+ entrances) and all other entity
+  // types (entrances, organizations, massifs) so the list never shows a cave
+  // redundant with its entrance.
   const filteredResults = useMemo(
     () =>
       results.filter(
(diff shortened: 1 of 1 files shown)


ClemRz (approved):

- **[QuickSearch.jsx:34]** Minor robustness point: if the API ever fails to include `nbEntrances` on a cave result (e.g. a schema change or partial response), that cave would be silently hidden from results. The current behavior is consistent with how `Entity.jsx` already handles the same field, so this is fine as-is — just worth being aware of if the API changes. No action needed.

### Nitpicks (Optional)

- **[QuickSearch.jsx:30-37]** The comment is well written and clarifies the "why" nicely. Consider a tiny wording tweak: "Keep only networks (caves with 2+ entrances) **and all other entity types**" to make explicit that entrances, organizations, and massifs pass through untouched — though the code itself is self-documenting enough.

---

Clean, minimal change that solves the stated problem. The `useMemo` with a correct dependency array, the nullish coalescing fallback, and the filter logic all look good. The pattern is consistent with the existing `entityOptionForSelector` helper. Approved.

PR #1455, opened 2026-07-10
refactor(riggings): improve riggings UX for reading and editing

Paul-AUB:
# refactor(riggings): improve riggings UX for reading and editing

## 🤔 What

Full UX overhaul of the riggings (équipement) feature on entrance pages — both the read view and the edit form.

**Read view (`Entry/Riggings/`)**
- Add `RiggingSummary`: two chips below the rigging title showing obstacle count (ICU plural) and approximate total rope length with a rope icon (hidden if 0, prefixed with `~` when values could not all be parsed)
- Zebra striping on table rows (was applied to all rows before)
- Column widths set (obstacles 25 %, ropes 10 %, anchors 20 %, observations flexible)
- `white-space: nowrap` on header cells — prevents word-break on small screens; table scrolls horizontally
- Empty cells replaced with an em-dash `—` in `text.disabled` colour (skipped in diff/snapshot mode to preserve `HighLightsLine`)
- Fix typo `marginBotton` (had no effect)
- `NetworkInlineLink`: new inline component (icon + label) that stays on one line when injected into a `FormattedMessage` value

**Edit form (`EntitiesForm/Riggings/`)**
- Extract `ObstacleField`, `ObstacleRowActions`, `ObstacleCard` from the monolithic `index.jsx`
- Responsive layout: card stack on mobile (< md), compact table on desktop
- `useForm({ mode: 'onBlur' })` for earlier validation feedback
- Actions (move up/down/delete) always visible — no more hover-only discovery
- Auto-focus on the obstacle field of newly appended rows
- `AnchorToolbar`: absolutely-positioned button on the Anchors field that opens a popover with cardinal arrow insertion (↑ ↓ ← →) and an i18n anchor notation legend (Spit, Bolt, Piton, Natural, Soft, Drilled, Redirect)
- `DiscardChangesDialog`: reusable confirmation dialog wired on both the creation toggle and the edit toggle; prevents silent data loss when closing a dirty form

**Utilities**
- `utils/ropeLength.js` + `ropeLength.test.js`: best-effort rope length parser supporting `C60`, `c 25`, `C12,5`, `10m`, `10 m`, `25,5 m`, multi-value cells (`C10 + C25`); cells with only dashes/slashes are ignored (not counted as unparsable)
- `rope.svg` icon added to the asset barrel (`src/assets/icons/`)

## 🤷‍♂️ Why

The riggings section is one of the most consulted sections by cavers preparing a trip. The previous read view had no summary, 10 px font on mobile, no column widths, and no visual empty-cell treatment. The edit form was a single large file with hover-only actions, no mobile support, no dirty-state guard, and no way to insert caving notation symbols.

## 🔍 How

- `RiggingSummary` uses a best-effort regex parser (two alternation groups: C-prefix and bare `number + m`) and shows a `~` prefix when at least one non-empty, non-dash cell could not be parsed
- `ObstacleField` wraps a single `Controller` and conditionally renders `AnchorToolbar` via `position: absolute` on a wrapper `Box`, avoiding any layout shift
- `AnchorToolbar` uses `onMouseDown + preventDefault` to keep focus on the textarea while opening the popover; `onKeyDown` handles `Enter`/`Space` for keyboard accessibility
- `DiscardChangesDialog` is driven by `formState.isDirty` exposed via an `onDirtyChange` prop and a local `isFormDirty` state in the parent
- All new strings are i18n'd (17 new keys in `en.json`, translated into all 14 languages)

## 🔗 Related API PR

None — no API changes required.

## 🧪 Testing

1. Navigate to an entrance with riggings (e.g. `/ui/entrances/<id>`)
2. **Read view**: verify obstacle count chip, rope length chip with `~` prefix when anchors contain non-parsable values, zebra rows, nowrap headers on narrow viewport, em-dash in empty cells
3. **Diff/snapshot mode**: verify no em-dashes appear and `HighLightsLine` still works
4. **Edit — desktop**: open the rigging form, verify compact table, always-visible actions, auto-focus on new row, dirty-state dialog on cancel
5. **Edit — mobile** (< md): verify card layout, `Obstacle N` header, Obstacle + Rope side-by-side, Anchors full width, AnchorToolbar button centred on the field
6. **AnchorToolbar**: click the `EditNote` button, insert arrows, verify cursor position is preserved; open the legend popover and check i18n abbreviations match the current locale
7. Run `yarn test src/utils/ropeLength.test.js --watchAll=false` — all 9 tests must pass

## 📸 Previews

image
image
image
image
image


ClemRz (approved):

- **[ObstacleCard.jsx:28]** `mb: 0.1` uses a non-integer spacing value. Per project conventions, MUI's `theme.spacing` is array-based and expects integer multiples. The visual effect of `0.1` (0.8px) is essentially invisible. Consider removing it or using a raw pixel value if tiny spacing is genuinely needed:
  ```suggestion
  mb: 0
  ```

- **[ObstacleCard.jsx:42]** `Stack spacing={0.75}` is another non-integer spacing value (6px). If a 6px gap is intentional, this technically works with MUI's factor-based spacing, but it deviates from the project's integer-only convention. Consider using `spacing={1}` (8px) for consistency, or a `gap: '6px'` escape hatch on a `Box` if 6px is critical.

- **[AnchorToolbar.jsx:98]** `Divider sx={{ mb: 1.5 }}` — same convention note (12px). Consider `mb: 2` or `mb: '12px'`.

- **[EntitiesForm/Riggings/index.jsx:84]** `focusIndex` is never reset to `-1` after the newly appended row mounts. Because `autoFocus` only fires on mount this is unlikely to cause a visible bug today, but if the list is ever keyed differently or a row remounts (e.g., after a swap puts a different item at that index), it could unexpectedly steal focus. A simple `useEffect` that resets `focusIndex` to `-1` after the append would be more defensive:
  ```js
  useEffect(() => {
    if (focusIndex >= 0) setFocusIndex(-1);
  }, [fields.length]);
  ```

- **[ObstacleField.jsx:42–48]** `handleInsertChar` is recreated on every render inside the `render` prop. Because `AnchorToolbar` has its own state (the popover), this won't cause performance issues today, but extracting the handler with `useCallback` outside the render prop (using `rhfField.value` via a ref) would keep things stable if the component grows.

### Nitpicks (Optional)

- **[RiggingTable.jsx:55]** `HighlightedTableCell` is defined inside the module scope (good!) but uses a default parameter `data = ''` — worth noting this is a nice improvement over the previous `isRequired` prop. No action needed, just an acknowledgment.

- **[Riggings.jsx:91]** The `DiscardChangesDialog` is rendered even when `isDiscardDialogOpen` is `false` (it's always in the tree). This is fine since `StandardDialog` likely short-circuits rendering when `open={false}`, but if you ever want to ensure a clean unmount/remount cycle, you could gate it with `{isDiscardDialogOpen && <DiscardChangesDialog ... />}`.

- **[rope.svg]** The SVG uses `stroke="currentColor"` but the `RiggingSummary` renders it as an `` tag. Since `` doesn't inherit CSS `color`, `currentColor` resolves to the SVG default (black). This works for the current light-theme use case but won't adapt if the chip is rendered on a dark surface. If dark-mode support is planned, consider using the SVG inline (as a React component) or a dedicated dark variant.

---

Overall this is a solid PR — well-structured decomposition of a monolithic form into focused components, good i18n coverage across all 15 languages, thorough unit tests for the rope parser, and a thoughtful UX improvement with the discard-changes guard. The non-integer spacing values are the only recurring deviation from documented conventions. Nice work!

urien (approved):
Merci @Paul-AUB c'est un très beau travail

PR #1459, opened 2026-07-11
feat(map): add compass-based map rotation

Paul-AUB:
# feat(map): add compass-based map rotation

## 🤔 What

Adds a compass button on the main map that rotates the map to match the device's
real-world orientation (Device Orientation API), for easier field navigation
toward a cave entrance.

- New **compass control** (bottom-right) that toggles a "follow heading" mode.
- When active, the map rotates continuously so the direction the device faces
  points up; a two-tone needle (red tip = North) always indicates true North.
- A second tap resets the map to north-up.
- **Mobile/touch only**: hidden on desktop (visible in DevTools device emulation
  for testing). Graceful states for permission denied / no sensor.
- i18n keys added and translated across all 14 language files.

## 🤷‍♂️ Why

In the field, keeping North fixed at the top of the screen forces the user to
mentally map the on-screen orientation to the terrain. Rotating the map to the
device heading removes that friction when navigating to an entrance.

## 🔍 How

- Added the **`leaflet-rotate`** plugin (Leaflet 1.9 / react-leaflet 5 have no
  native rotation). Rotation is enabled only on the map that has the compass
  (`rotate={isCompassControl}`), so every other map is unchanged.
- New hook **`useDeviceOrientation`** encapsulates the sensor: iOS 13+ permission
  prompt, `deviceorientationabsolute`/`webkitCompassHeading` reading, a hard
  ~8 Hz rate cap + jitter threshold, and a "no compass data" timeout that
  surfaces an `unavailable` error instead of failing silently.
- New **`CompassControl`** component (MUI button in a Leaflet control): needle
  points North using a continuous, shortest-path rotation (no full-turn spin at
  the 0°/360° wrap).
- **Custom panes fix**: base tiles (`LayersControl`) and the d3-hexbin heatmaps
  (`useHeatLayer`) are created inside the `rotatePane` when rotation is active,
  otherwise they stay in the un-rotated frame and end up misplaced.
- **Performance**: `map.setBearing` is coalesced to at most one call per animation
  frame and suspended during zoom animations; the heavy hexbin heatmaps are
  hidden while following the compass (announced via a `compassfollowchange` map
  event).

## 🔗 Related API PR

_None._

## 🧪 Testing

1. Open the main map on a mobile device (HTTPS) — or desktop with DevTools device
   toolbar (Ctrl+Shift+M) + the **Sensors → Orientation** panel.
2. Tap the compass button (bottom-right).
   - iOS: accept the permission prompt.
   - Rotate the device / vary `alpha`: the map rotates and the red needle keeps
     pointing North; heatmaps hide while following.
3. Tap again: the map returns to north-up and heatmaps reappear.
4. On a real desktop (no touch), the button is not shown.
5. Denied permission / no sensor → the button turns red with an explanatory
   tooltip.

## 📸 Previews

_To add: compass button (inactive / following) and a rotated map screenshot._


urien (approved):
Test sur téléphone et sur ordinateur
C'est parfait

ClemRz (approved):
Great feature — the implementation is well thought out with careful attention to performance (rAF coalescing, jitter filtering, heatmap hiding during rotation) and edge cases (iOS permission flow, no-sensor timeout, zoom animation suspension). The code is clean, well-commented, and follows project conventions. A few observations:

- **[CompassControl.jsx:161]** The `aria-label` is always `"Follow compass heading"` regardless of state (active, error). For accessibility, consider making it dynamic to match the tooltip — screen reader users should hear "Reset to north" when active, or the error message when in error state.
  ```suggestion
  aria-label={formatMessage({ id: tooltipId })}
  ```

- **[MapContainer.jsx:12]** The `import 'leaflet-rotate'` is a global side-effect that patches `L.Map` for every map in the application (Country, Region, Organization detail maps, etc.). Since `rotate` defaults to `false` this shouldn't cause issues, but it does increase the bundle size for every page that includes *any* map. Consider lazy-loading it only for the MapClusters page if bundle size becomes a concern in the future.

- **[useDeviceOrientation.js:107]** The `stop()` function resets `lastHeadingRef` but not `lastEmitRef`. If a user toggles the compass off and back on within 120ms, the first heading event would be throttled. Very minor, but for completeness:
  ```suggestion
  const stop = useCallback(() => {
    clearNoDataTimer();
    removeListeners();
    setError(null);
    setHeading(null);
    lastHeadingRef.current = null;
    lastEmitRef.current = 0;
  }, [clearNoDataTimer, removeListeners]);
  ```

- **[CompassControl.jsx:30]** The `CompassNeedle` component is defined inline in the same file without `React.memo`. Since it receives a rapidly-changing `bearing` prop during rotation, it will re-render frequently anyway, so memoization wouldn't help here. But if future changes add more children or complexity to this component, extracting it with `memo` could be worth considering.

### Nitpicks (Optional)

- **[CompassControl.jsx:146]** The tooltip `open={error ? true : undefined}` — this forces the tooltip to stay open when there's an error, which is a nice UX touch for discoverability. Just noting that on mobile where the button is primarily used, permanently-open tooltips can sometimes overlap other controls. Might be worth testing on a small phone screen.

- **[useDeviceOrientation.js:50]** `const [isSupported] = useState(detectSupport)` — this pattern (lazy initializer) runs `detectSupport` only once, which is correct. Nice touch to avoid re-detecting on every render.

- **[hooks/index.js:27]** The export uses `export { default as useDeviceOrientation }` while the hook also has a direct import in CompassControl via relative path (`../../../../hooks/useDeviceOrientation`). Both approaches are fine and consistent with how other hooks in the project are consumed.

PR #1463, opened 2026-07-17
improvement(map): improve map markers, massif import and messages UX

Paul-AUB:
## 🤔 What

A batch of UI/UX improvements across the map, the massif geometry import, and the messaging pages.

**Map markers**

- **Network marker** redesigned: a blue hexagon carrying the three white entrance glyphs and their connecting path, replacing the previous reuse of the `Caves` PNG marker.
- **Organization marker** redesigned: an amber square balloon with a downward pointer, anchored on the pointer tip so the pin marks the exact location (was a 24×24 centered icon).
- **Entrance markers** slightly enlarged (small `6→8`, medium `10→11`, large `14→15`) so the smallest ones stay visible.
- **Quality filter legend** reordered to `Insufficient → Satisfactory → Good` instead of relying on `Object.values()` ordering.
- Fixed the `alt` text of `CoordinatesMarker`, which said `networkIcon`.

**Massif**

- The massif polygon no longer covers the entrance markers on the massif map.
- The geometry import modal now uses the shared `FileSelectorInput` (drag & drop, file chip, removal) instead of a bare hidden `<input type="file">`, reports parse errors inline instead of via a toast, labels the two ends of the simplification slider (_Precise_ / _Simplified_), and disables the confirm button until a file is actually loaded.
- `FileSelectorInput` gained a `multiple` prop (default `true`) so it can be used for single-file selection with correctly singularized labels.

**Messages**

- New message dialog opens fullscreen on mobile, with the textarea filling the leftover height.
- The conversation page height is now derived from the theme instead of hardcoded values, and the inbox drops its card chrome on mobile like a regular mail client.
- <kbd>Enter</kbd> no longer sends on devices with a virtual keyboard — it inserts a line break, and sending goes through the button only.
- Sending a message no longer closes the virtual keyboard.
- The `UserAvatar` component is extracted to `components/common/UserAvatar` and reused by both the AppBar and the inbox (was duplicated with a local `getInitials`).
- Fixed several React console errors (see _How_).

closes #1450

**Search**

- Restored the missing gap between the advanced search fields.

## 🤷‍♂️ Why

The map markers for networks and organizations were reusing generic icons, which made them hard to tell apart from the round entrance/cave markers — networks in particular were rendered with the `Caves` marker. The smallest entrance markers were also hard to spot at low zoom.

On the massif map, the polygon was drawn in the same Leaflet pane as the entrance markers and ended up covering them.

The messaging pages were built desktop-first: the compose dialog and the conversation view didn't adapt to mobile, and <kbd>Enter</kbd>-to-send made it impossible to write a multi-line message on a phone (no usable <kbd>Shift</kbd>+<kbd>Enter</kbd> on a virtual keyboard). They also produced a steady stream of React console errors.

## 🔍 How

**Markers** — `NetworkMarker` and `OrganizationMarker` are now inline SVG rendered through `renderToString` into a Leaflet `divIcon`, instead of `` tags pointing at `MapMarkersConfig`. The glyph paths are the verbatim tracings from `network.svg` / `organization.svg` (paths and transform matrices kept as-is), only recolored white and repositioned via a wrapping group transform. Colors come from the MUI palette (`blue[700]`, `amber[500]`) — blue is already the network semantic elsewhere on the map (heatmap, highlight overlay). The organization `viewBox` is extended to `100×130` to host the pointer, and `iconAnchor` is moved to the tip.

**Massif polygon z-order** — a dedicated `massifPolygonPane` is created at `zIndex: 350`, below the default `overlayPane` (400) where the entrance circle markers live. The pane component renders before `<GeoJSON>` so its `useEffect` runs first and the pane exists when the layer is added; the SVG renderer is bound to the same pane (the SVG renderer is still needed to avoid the 0×0 canvas bug when the tab is hidden).

**Enter-to-send** — keyed on `useMediaQuery('(pointer: coarse)')` rather than a breakpoint: a narrow desktop window still has a physical keyboard. `enterKeyHint` is set accordingly (`enter` vs `send`).

**Keyboard dismissal** — `onMouseDown={e => e.preventDefault()}` on the send button, so pressing it doesn't move focus out of the input.

**Console errors** — two root causes:

- `styled(Component)(({ $prop }) => …)` forwards every prop down to the DOM; MUI only filters props it knows. Added `shouldForwardProp: prop => !prop.startsWith('$')` to `StyledListItem`, `MessageBubble` and `MessageDate`.
- `FormHelperText` renders a `<p>`, which cannot contain a `<div>`. The helper-text `<Box>` is now `component="span"`.

**Messages page height** — `theme.spacing` is an array in `grottoTheme`, so `spacing(2)` is `8px`, not `16px`. The height calc derives each term from the theme (`appBarHeight`, `spacing`) and keeps `containerPb` in sync with `PageContainer`'s own `pb: 1`. Both `100vh` and `100dvh` are emitted, the former as a fallback.

## 🔗 Related API PR

_None — frontend only._

## 🧪 Testing

**Map**

- Open the world map and check that network (blue hexagon) and organization (amber square pin) markers are distinguishable from entrance/cave markers at several zoom levels, on both light and satellite tile layers.
- Verify the organization pin tip sits on the actual location.
- Check the quality filter legend order in the map data control.

**Massif**

- Open a massif with a polygon and entrances: the entrances must be visible above the polygon, and clicking one must still open its popup.
- Print / switch tabs and come back: the polygon must still render (0×0 canvas regression).
- In the massif form, import a `.geojson` and a `.zip` shapefile via both drag & drop and click. Try an invalid file: the error must appear inline in the modal. Remove the file and re-import the same one.

**Messages**

- On a phone (or with a coarse-pointer emulation): compose dialog must be fullscreen, <kbd>Enter</kbd> must insert a line break, and tapping send must not close the keyboard.
- On desktop: <kbd>Enter</kbd> sends, <kbd>Shift</kbd>+<kbd>Enter</kbd> inserts a line break.
- Open the browser console on the inbox and a conversation: no React warnings should appear.

## 📸 Previews

image

image

image


urien (approved):
Concernant le point de vue de l'utilisateur
tout est parfait
La gestion de la fiche d'équipement est très pertinente. Il serait intéressant d'avoir des retours d'expérience d'ici quelques temps de la part des utilisateurs et des personnes qui ont une réflexion sur la sécurité. Ils ont été sollicités et n'ont pas fait de retour

ClemRz (approved):
Thorough PR — the AppLink refactoring, massif pane z-ordering, Enter-to-send differentiation by pointer type, and the shouldForwardProp fixes are all well thought out. The translations are consistent across all 15 language files. A few items worth addressing:

- **[ShapefileImport.jsx:168]** `onError` is declared as a required prop and defined in PropTypes, but is never called inside the component — errors are now handled inline via the `parseError` state. This leaves dead code in both `ShapefileImport` and its caller (`PolygonMap.jsx` still passes `handleImportError`). Consider either removing the prop entirely (if inline errors are the new UX) or wiring it back in as a fallback for unexpected errors that escape the try/catch.

- **[PartnersCarousel.jsx:~78]** At the `PartnerItem` call site, `alt={name}` is passed but `PartnerItem` doesn't declare an `alt` prop — it uses `name` directly for the image alt text. The stray `alt` prop ends up as an HTML attribute on the rendered `<a>` tag. Consider removing `alt={name}` from the call site and adding `name={name}` (which is what the component already expects and uses).

- **[Messages/index.jsx:StyledCard]** The styled component returns a raw template literal string instead of an object. This works with Emotion, but it makes the code harder to lint and loses autocomplete in editors. This is a preexisting pattern in the codebase though, so feel free to keep it — just flagging in case it was unintentional.

- **[ConversationDetail.jsx:~260]** The focus fix (removing `disabled={isSending}` from the TextField) solves issue #1450 nicely. One edge case: since the input stays enabled during send, a user could modify `replyText` while the message is in flight. The `handleSend` guard prevents double-send, but the text could change after `replyText.trim()` is captured. The current code dispatches `replyText.trim()` synchronously and then clears it — which is fine because React state is captured in the closure. Just confirming this was considered.

- **[Welcome.jsx]** The sign-up CTA wraps a `<Button>` inside an `<AppLink>` (`<a>` wrapping a `<button>`). While browsers tolerate this, it's semantically invalid HTML (interactive content inside interactive content). Consider using `component={AppLink}` on the Button instead, as done elsewhere in this PR (e.g., `RandomEntryCard`, `AccountPage`):
  ```jsx
  <Button component={AppLink} to="/ui/signup" color="secondary" ...>
  ```

### Nitpicks (Optional)

- **[.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml]** and **[src/hooks/.eslintrc]** — these are pure line-ending normalization changes (CRLF → LF). Not blocking, but they add noise to the diff and git blame. If intentional (standardizing line endings), a `.gitattributes` rule would prevent them from recurring.

- **[PartnersCarousel.jsx:PartnerItem]** Since `PartnerVignette` is now a styled `AppLink`, clicking the image is handled by the link. The `name` prop on `PartnerItem` lacks `isRequired` in PropTypes even though it's needed for accessibility (it's the `alt` text). Consider marking it required.

- **[AppLink.jsx]** Minor: when both `to` and `href` are null/undefined (no prop passed), the component falls through to render a `<RouterLink to={null}>`. This would navigate to the current page. A defensive guard or PropTypes `isRequired` on at least one of `to`/`href` could prevent misuse — though current call sites always pass one or the other.

PR #1464, opened 2026-07-17
feat(ImportCsv): adapt CSV import to async API (polling for job progress)

Paul-AUB:
# feat(ImportCsv): adapt CSV import to async API (polling for job progress)

## 🤔 What

Adapts the entrances CSV import flow to the new asynchronous, queue-based `POST /entrances/import-rows` API contract.

- Add `jobStatusUrl` API route (`GET /api/v1/jobs/:batchId`)
- Split the import thunk in two: `importEntranceRows` (async, submits then polls) and `importDocumentRows` (unchanged, still synchronous — the documents endpoint wasn't touched by the API)
- Add a `pollJobStatus` thunk and a new `useJobPolling` hook (recursive `setTimeout`, not `setInterval`, so slow responses never overlap)
- Extend `ImportCsvReducer` with `isPolling`, `batchId`, `status`, `progress` to track the async job
- Update `Step4.jsx`: determinate progress bar, completion summary, links to the signed report URLs (replacing the in-memory CSV download buttons for the entrances flow only)
- Handle the new `IMPORT_COMPLETE` notification type in `formatNotification.js` / `notification.type.js`
- Add/translate i18n keys for the new states (submitted, processing, job failed, report downloads, report expiry) across all 15 language files
- Forward rest props (`data-testid`, etc.) through the shared `Alert` component
- Fix the reset-on-navigation flow (`Step2.jsx`) so returning to the file step, not leaving `Step4`, is what clears a completed/in-flight import — navigating away and back to `Step4` now resumes polling instead of losing the batch

## 🤷‍♂️ Why

Closes #1445. Backend PR [grottocenter-api#1708](https://github.com/GrottoCenter/grottocenter-api/pull/1708) converts `POST /entrances/import-rows` from a synchronous endpoint (which times out on large imports) into an async queue: it now returns `202` with a `batchId` instead of `200` with results, and progress must be polled from `GET /jobs/:batchId`. Without this change, entrances CSV import breaks as soon as the API PR is merged.

## 🔍 How

The real API contract was read directly from the backend implementation (not just the issue description), which surfaced a few things worth flagging to reviewers:

- `result.reportUrls.success` (singular) vs `result.summary.successes` (plural) — same response object, different pluralization. Handled explicitly in `Step4.jsx`.
- `result` stays `null` when `status === 'failed'` — the failure UI reads `progress.failures` instead.
- `progress.completedChunks` can reach `totalChunks` while `progress.processedRows < totalRows` when some chunks fail (failed chunks count toward `completedChunks` but not `processedRows`). The progress bar is forced to 100% on any terminal status so it doesn't visually stall.
- `useJobPolling` uses a recursive `setTimeout` (schedules the next tick only after the previous request resolves) rather than `setInterval`, to avoid overlapping requests / out-of-order progress updates if the API responds slowly under load — the exact scenario this async queue exists for.
- Documents import (`importDocumentRows`) was intentionally kept as a separate thunk rather than unified with the entrances one — the two flows no longer share anything beyond the initial POST, and keeping them separate means zero risk of regressing the still-synchronous documents path.

## 🐛 Backend bugs found while implementing this

Found while reading `CSVImportQueueService.js` / `NotificationService.js` directly (not blocking, but worth fixing on [grottocenter-api#1708](https://github.com/GrottoCenter/grottocenter-api/pull/1708) before/after merge):

1. **`IMPORT_COMPLETE` is missing from `NOTIFICATION_TYPES` and `actionVerbMap` in `NotificationService.js`.** The notification row is created (seeded as id 8), but anything routing through `actionVerbMap` will `throw Error('Unknown notification type: IMPORT_COMPLETE')`. The front handles this defensively (early-returns before the entity-detection loop in `formatNotification.js`, reusing the existing "unknown entity" `link: '#'` contract), but this should still be fixed API-side.
2. **No notification is emitted on the `failed` path** — only on `completed`. Polling is currently the only way for a user to learn their import batch failed.
3. **`completedChunks` counts failed chunks, `processedRows` doesn't.** A batch with failed chunks can reach `completedChunks === totalChunks` while `processedRows < totalRows`, so a naive `processedRows/totalRows` progress bar stalls below 100% even though the job is done. Worked around on the front (progress forced to 100% on any terminal `status`), but the root cause is in `getBatchProgress`.
4. **Naming inconsistency**: `result.reportUrls.success` (singular) vs `result.summary.successes` (plural) — same response object. Not a bug per se, but easy to introduce a client-side bug from, and worth a look.
5. **Swagger gap**: `JobBatchStatus.result` is documented as a bare `type: object` with no inner properties, so `reportUrls`/`summary` aren't discoverable from the spec — anyone codegening from swagger will miss them.

## 🔗 Related API PR

https://github.com/GrottoCenter/grottocenter-api/pull/1708

As noted above, the API PR should merge first (or alongside) for this to work end-to-end — until then, `import-rows` still behaves synchronously in production.

## 🧪 Testing

- Reducer unit tests cover the `SUBMITTED → PROGRESS → SUCCESS` and `SUBMITTED → PROGRESS → FAILURE` sequences, and confirm `batchId`/`progress` survive completion (unlike the previous reducer, which reset to `initialState` on success).
- `useJobPolling` unit tests cover the overlap regression (slow response + short interval never produces concurrent requests) and confirm no timer survives unmount.
- Full existing suite (603 tests) passes with no regressions.
- Manual/E2E verification against a live API was intentionally not done in this PR (would have required pointing at a real backend); recommend testing against the deployed API PR once merged, with an admin account, covering: large import progress, a batch with some failed chunks (progress bar reaching 100%), and navigating away from/back to Step4 mid-import.

## 📸 Previews

_Not included — recommend testing directly against a review/staging deployment once the API PR is merged._


ClemRz (approved):
Good work overall — the async polling approach is well-thought-out (recursive `setTimeout` over `setInterval` is the right call), the separation between entrance and document flows is clean, and the reducer/action coverage is solid. The PR description is exemplary — the backend bug notes and API contract observations are especially helpful for reviewers.

- **[useJobPolling.js:20]** When the poll returns a terminal status (`completed`/`failed`), the dispatched action sets `isPolling: false` in the store. But between that dispatch and the React re-render (which triggers the effect cleanup), the `.finally()` callback still fires with `cancelled === false` and schedules one more timeout. That timeout IS cleared by the effect cleanup moments later — so this works correctly — but it means there's always one "phantom" scheduled timeout that gets immediately cleared. Consider checking the Redux state or having `pollJobStatus` return a signal indicating whether polling should continue, and setting `cancelled = true` inside `.finally()` when the result is terminal. This would make the lifecycle more explicit and avoid relying on React's re-render timing. Not blocking — the current implementation is correct — but it would make the logic easier to reason about.

- **[useJobPolling.js:33-37]** `PropTypes` on a hook function don't do anything at runtime — React only checks PropTypes on component props. This is dead code. Consider removing it, or if the intent is documentation, a JSDoc `@param` block would be more appropriate for a hook.

- **[Step4.jsx:88-95]** The `isEntranceImport && importCsv.isLoading` block shows the "submitted" message while `isLoading` is true (between `IMPORT_ROWS_START` and `IMPORT_ROWS_SUBMITTED`). That's the brief window while the POST is in flight. The message says "has been submitted" which is slightly premature — it hasn't been accepted yet, just sent. Consider either showing a generic "Submitting..." spinner here (before the 202 is received) and only showing "submitted and is being processed" once `isPolling` becomes true, or just accepting the minor UX imprecision as-is. Low priority.

- **[Step4.jsx:119]** The raw error string `importCsv.error` is rendered directly into a `<Typography>`. For the entrance flow, this can be "The import job failed." (hardcoded English in the action creator). Consider using an i18n key here instead of the hardcoded English string from the action, so the error message respects the user's locale. One option: dispatch a key like `csvImport.jobFailed` as the error value and format it in the component.

- **[ImportCsv.js (pollJobStatus)]** If the network request succeeds but `response.json()` fails (malformed response body), the error falls through to the `.catch()` which dispatches `importRowsPollFailure`. This stops polling entirely (`isPolling: false`). A transient JSON parse error during a single poll tick probably shouldn't kill the entire polling loop — consider either retrying a few times before giving up, or only stopping on repeated failures. Not blocking for a first iteration, but worth considering as a follow-up hardening step.

- **[Alert.jsx]** The `...rest` spread is a pragmatic solution for forwarding `data-testid`, but it's worth noting this now forwards any unknown prop to the underlying MUI `Alert` DOM element, which could produce React console warnings for non-standard HTML attributes in other usage sites. An alternative is to only extract `data-testid` explicitly: `const { 'data-testid': testId, ...otherRest } = rest`. Low priority since the only current usage is `data-testid` which is a valid DOM attribute.

### Nitpicks (Optional)

- **[Branch name]** The branch is named `feautre/import-csv-async` — there's a typo ("feautre" instead of "feature"). Won't affect functionality but shows up in git history.

- **[ImportCsv.js:153]** The hardcoded error string `'The import job failed.'` could be extracted as a constant for consistency with the action type naming pattern. Minor readability preference.

- **[Step4.jsx]** The file is now 290+ lines. It remains readable, but if more states are added (e.g., partial failure UI, retry button), consider extracting the entrance-specific result rendering into a sub-component like `EntranceImportResult`.

ClemRz (approved):
.

urien (approved):
C'est parfait

Image

PR #1465, opened 2026-07-19
feat(documents): allow an organization to be an author

Paul-AUB:
# feat(documents): allow an organization to be an author

## 🤔 What

Documents can now be authored by **organizations** as well as **persons**, instead of persons only. Implements the front-only part of GrottoCenter/grottocenter-front#1419.

- New single, mixed author picker in the document form: one search field that queries **persons and organizations at once**, selected authors shown as chips.
- Inline creation of a new person **or** organization is kept behind the `+` panel (Person / Organization toggle only shown there).
- "At least one author" is now enforced in form validation (counts persons ∪ organizations).
- Every place that **displays** a document's authors now renders organization-authors next to person-authors:
  - Document details page (with a person / organization icon before each author)
  - Documents `EntityTable` "Author" column
  - `DuplicatesHandler` merge UI (schema + merge state + submitted payload)
  - Advanced-search "Organization author" filter
- `authorsGrotto` added to the document payload allowlist (`BASE_PAYLOAD_FIELDS`) so it is no longer silently stripped on submit.
- i18n keys added and translated across all 14 language files.

## 🤷‍♂️ Why

Many documents are actually authored by organizations (federations, clubs, speleo groups). Until now contributors worked around the person-only picker by creating "ghost cavers" (fake persons whose nickname is really an organization's name), which pollutes the persons directory and search. The backend already supports organization-authors through a parallel `authorsGrotto` (write) / `authorsOrganization` (read) collection, so this ships front-only.

## 🔍 How

- **Form state** (`EntitiesForm/Document/Provider.jsx`): added `authorsGrotto` to the default attributes, bridged the API read field `authorsOrganization` → internal write field `authorsGrotto` in `normalizeInitialValues`, and added the author-count validation rule. Same Provider drives both the create page and the edit route/modal, so both are covered.
- **Picker** (`formElements/AuthorsSection/`): a single MUI `Autocomplete` (multiple) backed by `useEntitySearch(['persons', 'organizations'])`; the two flat arrays `authors` / `authorsGrotto` stay the source of truth, the chip list is derived from them. Reuses the existing `CreateCaverPanel` and `CreateNewOrganization` for inline creation (the latter gained an optional `onCreateSuccess(org)` callback so it can be reused outside a single-value context field).
- **Payload**: `authorsGrotto` added to `BASE_PAYLOAD_FIELDS` in `documentTypeHelpers.js`; `buildFormData` already serializes it correctly.
- **Display sites**: mirrored the existing `DocumentDetails` merge pattern (persons + `authorsOrganization`) into the table column, duplicate-merge schema/state, and search filter.

## 🔗 Related API PR

No API change required to ship this — the backend already supports both author collections.

Two optional backend follow-ups are tracked in GrottoCenter/grottocenter-api#1729:
- Index `authorsGrotto` in Typesense (needed for the new "Organization author" search filter to actually return results).
- Rename the write field `authorsGrotto` → `authorsOrganization` for API contract symmetry, and show all org-authored documents (not only Collection ancestors) on the organization page.

## 🧪 Testing

1. Create a new document — one author row is seeded with the current user.
2. Search an organization in the author field, select it → chip appears with the organization icon.
3. Try clearing all authors → submit is blocked until at least one author (person or org) remains.
4. Create a brand-new organization from the `+` panel → it is attached to the document.
5. Save, open the document details → both the person and the organization appear under "Authors".
6. Edit an existing document that has organization-authors → they load correctly in the picker.
7. Check the documents table, the duplicate-merge UI, and the advanced-search "Organization author" filter show organization-authors.

## 📸 Previews

image

image

image

image


urien (approved):
Salut @Paul-AUB 
Nous avons eu un retour positif sur la nécessité d'accepter les organisations comme auteur de document
Le fonctionnement que tu proposes est parfait
Il y a tout de même une difficulté avec les documents pour lesquels ni auteur ni éditeur n'est disponible
De manière pratique j'utilise l'auteur Unknow

Image

Je propose qu'on conserve le fonctionnement que tu proposes et d'envisager une solution pour ces documents particulier dans un autre ticket. la solution actuelle étant toujours disponible

ClemRz (approved):

- **[AuthorsSection/index.jsx:137]** The `sx={{ mb: 0.5 }}` and `sx={{ gap: 0.25 }}` (line 162 of DocumentDetails) use non-integer spacing multipliers. The project convention states "no non-integer values with theme.spacing()" and MUI's sx shorthand internally calls `theme.spacing()` with these values. Consider using `mb: 1` (8px) or `mb: '4px'` for the 4px intent, and `gap: '2px'` or `gap: 0` / `gap: 1` in DocumentDetails. That said, this is a grey area — the convention was written with explicit `theme.spacing(1.5)` calls in mind, and MUI does support fractional multipliers in sx. I'll leave it to the maintainer's judgment.

- **[AuthorsSection/index.jsx:89]** The `sideActionEnabled` flag only becomes `true` after the first `isLoading` event (i.e., after typing ≥ 3 chars). This means users cannot create a brand-new person/organization via the `+` button without first searching. I see this mirrors the existing `AuthorsSelect` behavior exactly, so it's a pre-existing UX choice — but it's worth noting for a future improvement since the spec mentions being able to create from the `+` panel without a prior search.

- **[DocumentSearch.jsx:56,309]** The label `"Organization author"` is a hardcoded English string passed to `SearchTextAutocomplete`. This matches the existing pattern for all other labels in this file (all hardcoded English), so it's consistent — but now that the i18n key `"Organization author"` exists in all language files, it could be wrapped in `formatMessage({ id: 'Organization author' })` for proper localization. Same applies to `FILTER_LABELS` (line 56). This is a broader pre-existing tech debt, not something to block on.

- **[entitiesConfig.jsx:65]** The `documentAuthors` render function receives `(_value, doc)` — it relies on the EntityTable passing the full row object as the second argument. This works because the column config has `field: 'authors'` which provides `_value` as the cell value and `doc` as the row. Consider adding a brief JSDoc or inline comment explaining that the first argument is intentionally unused because this renderer merges two fields from the row object.

### Nitpicks (Optional)

- **[AuthorsSection/index.jsx:26]** Importing from `@/helpers/Entity` — the project steering marks `helpers/` as deprecated. The import is reading an existing export (`entityOptionForSelector`) so this isn't introducing new code into the deprecated folder, but if this helper is needed long-term it might be worth migrating it to `@/utils/` or a hook in a follow-up.

- **[DocumentDetails/index.jsx:162]** `gap: 0.25` produces a 2px gap. This is quite tight — worth checking visually that the icon and text link don't feel cramped on smaller screens.

---

**Overall**: Clean, well-structured PR. The field normalization between `authorsOrganization` (read) and `authorsGrotto` (write) is handled correctly. The `BASE_PAYLOAD_FIELDS` fix is critical and correctly placed. The duplicate-merge handler and entity table updates are thorough. I18n keys are added across all 15 language files and appear sorted. The `CreateNewOrganization` refactoring is backward-compatible (existing callers pass `contextValueName` so the `if (contextValueName)` guard works; the `organization` argument passed to `onCreateSuccess` is harmlessly ignored by `toggleCreateOrganization`). Good work.

PR #1467, opened 2026-07-21
refactor(theme): migrate to standard MUI spacing scale

Paul-AUB:
# refactor(theme): migrate to standard MUI spacing scale

## 🤔 What

- Switch `theme.spacing` from a custom array (`[2, 4, 8, 16, 24, 32, 64]`) to the standard MUI factor-of-8 number (`spacing: 8`) in `grottoTheme.js`
- Remap all 747 existing integer spacing usages (`theme.spacing()` calls, `sx` shorthand props, `Stack`/`Grid` spacing props) to the equivalent new factor, so every one renders the exact same pixel value as before
- Remove 76 fractional/overflow spacing values (e.g. `gap: 1.5`, `py: 8`) that were previously silently dropped by the array scale, so current rendering stays unchanged rather than suddenly activating
- Update `AGENTS.md` spacing guidance to reflect the standard scale

## 🤷‍♂️ Why

MUI's array-type spacing only accepts integer indices — passing any fractional factor (`theme.spacing(0.5)`, `sx={{ p: 1.5 }}`) throws a console warning and silently resolves to nothing (the declaration is dropped). This was happening throughout the app any time a developer wrote a fractional spacing value, without ever being noticed visually.

MUI's own documentation recommends the factor-of-8 scale by default and explicitly warns that the array form doesn't support fractional signatures. MUI's own components (`Divider`, `DataGrid` internals) have shipped fractional `theme.spacing()` calls that broke under array-based themes, confirming this is not a supported pattern even for library maintainers.

## 🔍 How

- `grottoTheme.js`: `spacing: [...]` → `spacing: paddingUnit` (8)
- An AST codemod (recast + ast-types) walked every `.jsx`/`.js` file to:
  - Remap integer `theme.spacing(n)` calls and spacing-prop shorthands (`m`, `p`, `gap`, `Stack`/`Grid` `spacing`, etc.) from the old array index to the equivalent new standard factor (`0→0.25`, `1→0.5`, `2→1`, `3→2`, `4→3`, `5→4`, `6→8`), preserving identical rendered pixels
  - Detect values that were previously inert (non-integer factors, or integers beyond the old array's length) and remove them entirely, since they never applied any spacing before
  - Leave `styled()()` CSS-in-JS bodies alone for bare numeric CSS properties (those are raw emotion CSS, not MUI spacing) — only explicit `theme.spacing()` calls inside them were touched
- Verified via `git show` against the pre-change commit that no multi-argument `theme.spacing(a, b)` call in the codebase mixes a valid and an inert argument, so no case needed manual disambiguation

## 🧪 Testing

- `yarn test`: 581 tests pass; the only 10 failures (`MapMassif.test.jsx`, `MapMassif.property.test.jsx`) are a pre-existing Leaflet mock issue (`map.getPane is not a function`), confirmed unrelated by reproducing them against the pre-change commit
- `yarn lint` could not run in the dev environment due to a pre-existing missing `@babel/eslint-parser` install, unrelated to this change
- No functional/visual change is expected: every touched value either renders the same pixels as before, or was already a no-op and is now simply absent from the code

## 📸 Previews

No visual change expected — this PR intentionally preserves current rendering. A visual smoke test of a few pages that had inert spacing values (documents list, advanced search filters, entity tables) is still recommended before merge.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -394,12 +394,15 @@ cy.get('.MuiButton-root').click();
 // ✅
 cy.get('[data-testid="save-btn"]').click();
 
-// ❌ Non-integer values with theme.spacing (array-type spacing only accepts integer indices)
-sx={{ mt: theme.spacing(1.5), p: theme.spacing(0.5) }}
-// ✅ Use integer multiples only
-sx={{ mt: theme.spacing(1), p: theme.spacing(2) }}
-// ✅ Or use raw pixel values for non-standard spacing
+// ❌ Raw pixel values for ordinary spacing (bypasses the theme scale)
 sx={{ mt: '12px', p: '4px' }}
+// ✅ Use theme.spacing() factors — the theme uses the standard MUI scale
+// (spacing: 8), so theme.spacing(factor) = factor * 8px. Fractional factors
+// (0.25, 0.5, 1.5, ...) are valid and resolve correctly.
+sx={{ mt: theme.spacing(1.5), p: theme.spacing(0.5) }}
+// ✅ Shorthand props go through the same scale — prefer them over theme.spacing()
+// when there's no other computation involved
+sx={{ mt: 1.5, p: 0.5 }}
 ```
 
 ---
--- a/packages/web-app/src/components/appli/AdvancedSearch/DocumentSearch.jsx
+++ b/packages/web-app/src/components/appli/AdvancedSearch/DocumentSearch.jsx
@@ -159,7 +159,7 @@ const DocumentSearch = () => {
           e.name,
           <Box
             key={e.name}
-            sx={{ display: 'flex', alignItems: 'center', gap: 1 }}>
+            sx={{ display: 'flex', alignItems: 'center', gap: 0.5 }}>
             <Icon sx={{ fontSize: 18, color: 'text.secondary' }} />
             <Translate>{e.name}</Translate>
           </Box>
@@ -183,7 +183,7 @@ const DocumentSearch = () => {
         sx={{
           display: 'flex',
           flexDirection: { xs: 'column', sm: 'row' },
-          gap: 1,
+          gap: 0.5,
           alignItems: { sm: 'center' },
           width: '100%'
         }}>
@@ -201,11 +201,10 @@ const DocumentSearch = () => {
             options={docTypeOptions}
             onChange={e => updateFilter('type', e)}
             value={filterState.type}
-            sx={{ mx: { xs: 0, sm: '4px' } }}
+            sx={{ mx: { xs: 0.25, sm: '4px' } }}
           />
         </Box>
       </Box>
-
       <SearchFilterAccordion
         filterCount={advancedFilterCount}
         expanded={advancedExpanded}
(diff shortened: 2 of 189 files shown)


urien (approved):
Effectivement je ne vois aucune différence

ClemRz (approved):

- **[packages/web-app/src/pages/Messages/index.jsx:68-69]** The comment says _"theme.spacing is an array here (see grottoTheme), so spacing(2) is 8px and spacing(1) is 4px — not the MUI defaults"_ — this is now stale and describes the old behavior that this PR removes. Consider updating or removing it.

- **[Multiple files: ~12 occurrences]** The codemod left behind empty `sx={{}}` props where previously-inert fractional values were stripped. These are dead code and add noise. Consider removing them (can be a follow-up). Files include: `PageHeader.jsx`, `MobileEntityList.jsx`, `AuthorsSelect/index.jsx`, `StringInput.jsx`, `FormAutoComplete.jsx`, `OperationSummary.jsx`, `MultipleSelect.jsx`, `IdentifierEditor.jsx`, `MultipleSelectWithOptions.jsx`, `AddFileForm/index.jsx`, `PagesEditor.jsx`, `TimestampFormatInput.jsx`.

- **[packages/web-app/src/components/appli/ImportObservationsWizard/steps/TimestampFormatInput.jsx:127]** Minor content change: `{sampleValue} → {formatted}` became `{sampleValue}→ {formatted}` (space before the arrow was removed). This looks like a codemod artifact. The space was intentional for readability.

- **[packages/web-app/src/components/appli/EntitiesForm/Document/FormContent.jsx:129,161,211]** and **[packages/web-app/src/components/appli/Entry/Snapshots/AccordionSnapshot.jsx:163]**, **[packages/web-app/src/components/common/Properties/Property.jsx:77]**, **[packages/web-app/src/components/appli/NotificationMenu/index.jsx:40-42]** — The codemod introduced unnecessary double parentheses around JSX, e.g. `{condition && ((<Box>...</Box>))}`. While functionally harmless, they're unusual and may confuse contributors. Consider removing the extra wrapping in a follow-up.

- **[~70 locations across the PR]** Many blank lines between JSX elements were removed. While this doesn't affect functionality, it changes the code's visual structure beyond what the spacing migration requires. This makes the diff harder to review and pollutes `git blame`. If possible, limit whitespace removal to a separate commit or PR for clean git history.

### Nitpicks (Optional)

- **[AGENTS.md:397-408]** The updated guidance now recommends `sx={{ mt: 1.5, p: 0.5 }}` and raw shorthand props as the preferred patterns, which contradicts the steering files that say "integer values only for theme.spacing()". The steering files (`.kiro/steering/architecture.md`) should be updated to match the new reality, or this could confuse future contributors/agents.

- **[packages/web-app/src/components/appli/Entry/Snapshots/AccordionSnapshotListPage.jsx:157-158]** `p: 0.25, m: 0.25` — This preserves the old rendering (2px from `spacing(0)` under the array scale), but the developer's intent when writing `p: 0, m: 0` was almost certainly "no padding/margin." Now that the confusing array quirk is gone, this might be a good opportunity to set these to `0` as the developer likely intended. Same for other `0 → 0.25` conversions that look like "reset to zero" intent (e.g., `pb: 0 → pb: 0.25` in various places).

- **[packages/web-app/src/components/appli/Organization/Details.jsx:41]** `border-radius: ${({ theme }) => theme.spacing(0.5)}` — This gives 4px border-radius, which is fine, but it's now mixing spacing tokens and border-radius (which is usually a `shape` concern). Not blocking, just noting the coupling.

PR #1470, opened 2026-07-23
feat(pwa): add PWA support and Android TWA app

Paul-AUB:
# feat(pwa): add PWA support and Android TWA app

## 🤔 What

Turns the web app into an installable Progressive Web App and ships a native
Android wrapper (Trusted Web Activity) built from it.

- **PWA**: service worker + web manifest via `vite-plugin-pwa`, making the app
  installable and offline-capable.
- **Digital Asset Links**: `public/.well-known/assetlinks.json` so the Android
  app can open the site full-screen without a browser chrome.
- **Android TWA** (`twa/`): a Bubblewrap-generated Gradle project wrapping the
  PWA — custom launcher activity, adaptive/maskable launcher icons, splash
  screens, app shortcuts and notification icon.
- **CI** (`.github/workflows/twa-build.yml`): builds the Android app with
  Gradle and handles signing-key transfer.
- **Docs**: README instructions for building the app locally.

## 🤷‍♂️ Why

Users increasingly access GrottoCenter from mobile. A PWA gives an installable,
offline-friendly experience directly from the browser, and the TWA lets us
distribute the same experience as a real Android app (e.g. via the Play Store)
without maintaining a separate native codebase — the Android app simply renders
the deployed PWA.

## 🔍 How

- Configured `vite-plugin-pwa` in `vite.config.mjs` to generate the service
  worker and web manifest (replacing the static `public/manifest.json`), and
  registered it in `src/index.jsx`.
- Added `assetlinks.json` under `.well-known/` to establish the verified link
  between the web origin and the Android package (removes the URL bar in the
  TWA).
- Generated the Android project under `twa/` with Bubblewrap: `build.gradle`,
  Gradle wrapper, `twa-manifest.json`, launcher/maskable icons, splash assets,
  shortcuts, and a custom `LauncherActivity`.
- Added `twa-build.yml` GitHub Actions workflow to build the app via Gradle,
  transferring the signing key through Gradle properties.

## 🔗 Related API PR

_None._

## 🧪 Testing

- Run `yarn build && yarn preview`, open in Chrome, and verify the install
  prompt appears and the app works offline after first load.
- Verify `/.well-known/assetlinks.json` is served on the deployed site.
- Build the Android app following the README instructions (or via the
  `twa-build` workflow) and confirm it launches the site full-screen with no
  browser URL bar.

## 📸 Previews

_N/A — infrastructure/build change._


ClemRz (changes requested):
Solid PWA + TWA implementation overall — the offline caching strategy is well thought out, the vite-plugin-pwa configuration is thorough, and the new components follow project conventions. A few things worth addressing before merging.

### Issues (Must Fix)

- **[.github/workflows/twa-build.yml:14]** The tag glob pattern `[0-9]+.[0-9]+.[0-9]+` uses an unescaped `.`, which matches _any_ character in a glob (including `-`). Tags like `1-0-0` or `1.0.0-rc1` would trigger a build, only to fail at the semver validation step. Consider escaping: `[0-9]+\.[0-9]+\.[0-9]+`. (The Resolve step still provides a correct defense-in-depth validation, so this is a minor correctness issue, but it's cheap to fix.)

- **[packages/web-app/vite.config.mjs:219]** The precache glob `**/*.{…,json}` will precache `/.well-known/assetlinks.json`. This is a `no-cache` file (by design in `staticwebapp.config.json`) — precaching it would serve a stale version offline and break the Digital Asset Links verification silently after a certificate rotation. Add it to `globIgnores`:
  ```suggestion
  globIgnores: ['**/stats.html', '**/*.{gz,br}', '**/.well-known/**'],
  ```

- **[packages/web-app/public/.well-known/assetlinks.json]** The file contains what appear to be real SHA-256 fingerprints, but `twa/twa-manifest.json` still has `"fingerprints": []`. Either both files should use placeholder strings (documented as needing replacement before first Play Store upload) or — more importantly — if these fingerprints are real, the Play App Signing fingerprint may be missing from `assetlinks.json`. The README clearly documents that **two** fingerprints are required (upload key + Play App Signing key), and only having one here means apps installed from the Play Store will show the Chrome URL bar. This needs to be verified before the first Play Console upload.

- **[packages/web-app/vite.config.mjs]** When `VITE_API_URL` is not set at build time (`apiPattern` is `null`), no runtime caching rule is registered for the API. This is silently correct for local dev, but a misconfigured production build would ship with zero API caching and no indication of the problem. Consider logging a warning at build time:
  ```suggestion
  if (!apiOrigin) {
    console.warn('[vite-plugin-pwa] VITE_API_URL is not set — API runtime caching disabled.');
  }
  ```

- **[packages/web-app/vite.config.mjs — manifest screenshots]** Four tablet screenshots (`home-tablet10.png`, `home-tablet7.png`, `map-tablet10.png`, `map-tablet7.png`) are added to `public/screenshots/` but are **not listed** in the PWA manifest's `screenshots` array. Since the manifest only has `form_factor: "narrow"` entries, the tablet shots are dead weight in the repo. Either add them to the manifest with `"form_factor": "wide"` (which is what Play Store uses for tablet listings and Chrome's install dialog on large screens), or remove the files.

- **[packages/web-app/vite.config.mjs:194]** Screenshot filename typo: `activitites-mobile.jpg` (double `i`) — the file is spelled the same way in `public/screenshots/`, so this is consistent, but both the filename and the manifest `label` (`'Recent activity feed'`) disagree. Consider renaming the file to `activities-mobile.jpg` for consistency.

- **[packages/web-app/src/components/common/AppBar/OfflineIndicator.jsx]** The `label` and `message` variables are computed after the early `if (isOnline) return null` guard. This is safe today (hooks are called unconditionally before the guard), but consider moving the `formatMessage` calls before the early return for consistency with the project's hook ordering convention and to avoid any future confusion.

- **[twa/manifest-checksum.txt]** This file is a Bubblewrap artefact (SHA-1 hash of `twa-manifest.json` used to detect drift). It has no runtime value and goes stale the moment `twa-manifest.json` is edited without running `bubblewrap`. Consider adding it to `.gitignore` (alongside the other Bubblewrap cache files) to avoid confusing future contributors.

- **[packages/web-app/src/pages/Account/index.jsx — OfflineDataSection]** The `OfflineDataSection` is rendered unconditionally, even for users who are not on a PWA-capable browser or whose service worker is not active (e.g., the app accessed via `localhost` in development with `devOptions.enabled: false`). `caches` being `undefined` is handled gracefully in `offlineCache.js`, but the section will always show a storage usage of `null` with no explanation in those cases. Consider gating the section's render on `'serviceWorker' in navigator` to avoid showing a confusing empty UI to desktop developers.

- **[.github/workflows/twa-build.yml]** GitHub Action `uses` references are not pinned to a commit SHA, only to a major version tag (`@v5`, `@v4`, `@v2`). Tags can be moved, which is a supply-chain risk. Best practice for security-sensitive CI is to pin to a SHA:
  ```
  actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v5
  ```

- **[twa/README.md]** The `robots.txt` concern mentioned in the original issue plan (`public/robots.txt` has `Disallow: /`) is flagged as "out of scope" but it directly undermines the discoverability goal of being on the Play Store. Worth adding a tracking note or a separate issue since the PWA must be crawlable for Play Store policies.

### Nitpicks (Optional)

- **[packages/web-app/vite.config.mjs]** The `escapeRegex` helper is defined inline and used only three times. It's correct, but consider extracting it as a named constant at the top of the module for clarity.

- **[commit history]** Several commits use bare `fixup`, `chore:`, `fix:` without a scope (e.g. `fixup sort translations`, `chore: add assets for playstore`, `fix: trim whitespace`). The project's commitlint config requires a scope (`type(scope): description`). While this doesn't block merging for non-squashed histories, a clean-up squash or rebase before merge would align with the team's commit convention.

- **[packages/web-app/public/logo512-playstore.png]** This file is added to `public/` but is neither referenced in the manifest (which uses `logo512.png` and `logo512-maskable.png`) nor in any source file. If it's only for the Play Store listing upload (which is done manually through the console), it could live outside `public/` to avoid being served and precached as a dead asset. Consider moving it to `twa/store_icon.png` (which already exists) or just removing the public copy.

ClemRz (changes requested):
Solid PWA + TWA implementation. The previous round of review items have all been addressed — the escaping fix in the tag glob, the `globIgnores` for `.well-known` and `screenshots`, the `VITE_API_URL` warning, the tablet screenshot manifest entries, the `OfflineDataSection` service-worker gate, and the `OfflineIndicator` hook ordering are all correct. A few remaining things below.

### Issues (Must Fix)

- **[packages/web-app/src/pages/Account/index.jsx — OfflineDataSection]** The error message key used in the `catch` block is the full sentence as a key: `'An error occurred. Please try again.'`. While that key does exist in `en.json`, it's an unusual pattern — most i18n keys in this codebase are either camelCase identifiers or very short phrases. More importantly, this key is passed directly to `formatMessage({ id: ... })` and there is **no `defaultMessage` fallback**. If the key is ever removed from a language file (or the component moves to a context where it isn't loaded), react-intl will render the raw key string to the user. Use a camelCase key consistent with the other new keys added in this PR:
  ```suggestion
  onError(formatMessage({ id: 'offlineDataClearError', defaultMessage: 'An error occurred. Please try again.' }));
  ```
  And add `"offlineDataClearError": "..."` to all 15 language files.

- **[packages/web-app/src/components/common/AppBar/OfflineIndicator.jsx:35]** The `slotProps` on the `Popover` uses a raw pixel value `mt: '4px'`. Per the project's convention (AGENTS.md), prefer the MUI shorthand factor over raw pixels:
  ```suggestion
        slotProps={{ paper: { sx: { mt: 0.5 } } }}>
  ```
  (`0.5 × 8px = 4px`, same visual result, consistent with the theme scale.)

- **[packages/web-app/src/components/common/AppBar/OfflineIndicator.jsx:5]** The import path `'../../../hooks'` traverses 3 `../` segments. Per the project's AGENTS.md rule, paths with 2+ `../` segments should use the `@/` alias:
  ```suggestion
  import { useOnlineStatus } from '@/hooks';
  ```

- **[packages/web-app/src/pages/ApplicationShell.jsx — AdminSessionExpiryBanner reorder]** The banner is moved from before `<AppBar />` to after it, with a comment explaining it was being hidden behind the fixed AppBar. This is a correct fix, not a PWA concern per se, but it is a **behaviour change** bundled into this PR (users who were seeing the session expiry banner in the wrong position will now see it correctly). Worth calling out explicitly in the PR description as a secondary fix so reviewers know this is intentional.

- **[packages/web-app/public/robots.txt + packages/web-app/index.html]** Two separate crawler-blocking mechanisms are in place: `robots.txt` has `Disallow: /` and `index.html` still has `<meta name="robots" content="noindex, nofollow">`. The README correctly notes the `robots.txt` issue as a known tracking item, but the `<meta>` tag in `index.html` was not mentioned and isn't addressed here. Both will block Play Store discoverability checks and Google's PWA quality criteria. These should be tracked in the same issue so they're fixed together before the app goes to production.

- **[twa/app/build.gradle — generateShorcutsFile task]** Minor inconsistency in the Gradle task name: `generateShorcutsFile` (missing `t` in `Shortcuts`). This is Bubblewrap-generated boilerplate so it may be intentional to avoid breaking existing CI that references the task by name, but it's worth a note for anyone who regenerates the project.

- **[twa/app/src/main/java/org/grottocenter/twa/LauncherActivity.java]** The `if/else` for `setRequestedOrientation` sets the same value (`SCREEN_ORIENTATION_UNSPECIFIED`) in both branches. This is Bubblewrap-generated code and the comment explains the reason (Android 8.0 Oreo crash), but the dead else branch is mildly confusing to anyone reading it without the comment. Low priority, but noted in case a future regeneration doesn't preserve the comment.

### Nitpicks (Optional)

- **[packages/web-app/src/hooks/useOnlineStatus.js]** The hook is exported as a named function declaration (`export function useOnlineStatus()`) while the project uses named function exports for hooks uniformly — this is consistent. No issue here, just confirming it's correct.

- **[twa/app/src/main/res/raw/web_app_manifest.json]** This file is a snapshot of the old `public/manifest.json` (it notably lacks the new `shortcuts` and `screenshots` fields added by `vite-plugin-pwa`). The README explains Chrome OS and Meta Quest use this to fall back to the web version, but it references `https://grottocenter.org/manifest.json` at runtime (see `webManifestUrl` in `build.gradle`), so the stale embedded copy is largely irrelevant. Still, it will diverge from the deployed manifest over time. Worth a comment or a tracking note that this file needs updating when the manifest changes significantly (icon set changes, scope changes, etc.).

ClemRz (approved):
All items from the previous two rounds have been addressed — the `offlineDataClearError` key is now camelCase with a `defaultMessage` fallback, the `mt: 0.5` theme-scale fix is in place, the `@/hooks` alias is used in `OfflineIndicator.jsx`, the tag glob is correctly escaped, the `.well-known` and `screenshots` entries are in `globIgnores`, the `VITE_API_URL` warning is wired up, and the tablet screenshots are listed in the manifest with `form_factor: "wide"`. The Android TWA project and CI workflow look correct.

A few remaining things worth calling out:

- **[packages/web-app/index.html:35]** The OG/Twitter image URLs are now hardcoded to `https://grottocenter.org/og-image.png`. This is fine for production, but any staging or preview deployment (e.g., the Azure preview URLs that are posted on this PR) will serve social cards with production URLs. Consider using the Vite env approach (`%VITE_PUBLIC_URL%/og-image.png`) or accepting the current behaviour explicitly in a comment so future contributors understand the tradeoff.

- **[packages/web-app/src/pages/Account/index.jsx — PersonalInfoSection / EmailSecuritySection]** Both sections still use sentence-as-key IDs for their error messages (e.g. `formatMessage({ id: 'This nickname is already taken.' })`, `formatMessage({ id: 'An error occurred. Please try again.' })`). The prior review flagged this pattern only for the `OfflineDataSection` catch block because that was the added code; these pre-existing uses are out of scope for this PR. Worth creating a follow-up ticket to align them with the camelCase convention that `offlineDataClearError` introduces.

- **[packages/web-app/src/index.jsx]** `registerSW({ immediate: true })` is called unconditionally, even when `devOptions: { enabled: false }` suppresses the SW in development. With `devOptions.enabled: false` (the current config), `registerSW` is a no-op in dev, so this is harmless. A brief comment to that effect would help the next developer who wonders why the call isn't guarded by `import.meta.env.PROD`.

- **[twa/twa-manifest.json]** `appVersion: "4"` and `appVersionCode: 4` are hardcoded legacy fields from the initial Bubblewrap generation. The CI workflow overwrites `versionCode`/`versionName` in `app/build.gradle` before building, so these values have no runtime impact on what ships. However, they could mislead someone who opens `twa-manifest.json` expecting to find the current version. A comment (or a `"_note"` field) clarifying that these are replaced by the CI at build time would prevent confusion.

### Nitpicks (Optional)

- **[twa/app/src/main/java/org/grottocenter/twa/LauncherActivity.java]** `generateShorcutsFile` typo (missing `t` in `Shortcuts`) is Bubblewrap-generated and intentionally left as-is per the author's comment. If the class is ever manually edited, it's worth fixing the task name too — just adding a `// typo: Bubblewrap-generated, keep as-is` comment on the method would suppress any future linter warnings without changing the task name.

- **[packages/web-app/public/lang/ar.json / he.json]** RTL language files are present and the offline keys are correctly added. Worth a quick manual sanity-check that `offlineIndicatorTitle: "غير متصل"` and `offlineIndicatorTitle: "לא מחובר"` render correctly in the actual `OfflineIndicator` tooltip before merging, since these are the only RTL cases and the Popover is right-anchored.

PR #1473, opened 2026-07-24
refactor(map): improve map performances

Paul-AUB:
refactor(map): improve map performances

## 🤔 What

- Introduce a tile-based caching strategy for map data requests (`mapTileCache.js`) with a dedicated cache-invalidation middleware
- Optimize the viewbox drawing of entrances
- Add icons on map tooltips
- Limit requests based on map zoom level and gracefully manage cache failures
- Replace the manual heatmap rendering with `supercluster` for caves, networks and massifs
- Organizations are also rendered with supercluster
- Extend `supercluster` clustering to also manage organization markers
- Simplify related map actions and components (`Map.js`, `MapClusters/*`) following the supercluster migration

## 🤷‍♂️ Why

The map was issuing too many redundant requests and re-rendering too much data client-side (manual heatmap), causing sluggish performance on zoom/pan, especially with a large number of entrances, caves, networks, massifs and organizations displayed simultaneously.

## 🔍 How

- `mapTileCache.js` stores fetched map data per tile/zoom bucket and a new `mapCacheInvalidationMiddleware` clears stale entries when filters change
- `MapClusters` components now delegate spatial clustering to `supercluster` instead of computing a manual heatmap, unifying the clustering logic across caves, networks, massifs and organizations
- Map actions (`actions/Map.js`) were simplified now that request/caching concerns are handled by the tile cache layer

## 🧪 Testing

- Browse the map at various zoom levels and pan around; verify caves, networks, massifs and organizations markers cluster/uncluster correctly
- Verify tooltip icons display correctly for each marker type
- Verify map requests are reduced/cached when panning within already-loaded tiles (check network tab)

image
image
image


ClemRz (changes requested):
Solid, well-thought-out refactor. The architecture is sound — replacing the d3-hexbin layer with supercluster is a meaningful improvement in both maintainability and runtime behaviour. The tile cache design is clever: the version counter + microtask coalescing pattern is a nice touch that avoids Redux noise on hot pan paths. Comments are unusually good throughout. A few things worth fixing or considering before merge:

---

### Issues (Must Fix)

- **[`utils/mapTileCache.js:117`]** `inFlight` is never cleared on success. The success branch writes `{ data, fetchedAt: now() }`, which drops `inFlight` — but only because the record is *replaced*, not mutated. This is fine today. However on the failure path the record is also replaced (`{ data?, fetchedAt, failed: true }`), so `inFlight` is dropped there too. The problem is subtler: the record set just *before* the fetch starts at line 145 is `{ ...(existing || {}), inFlight: promise }`. If `evictIfNeeded` runs between when the in-flight record is written and when the promise resolves (e.g. another tile's success triggers eviction of this exact key), the promise will still resolve and write a fresh record — but any *second* call to `fetchForBounds` for the same tile during that window will not find `inFlight` and will kick off a duplicate request. The LRU eviction is oldest-first, so in practice this is unlikely for recently-fetched tiles, but the guard `if (existing?.inFlight) return existing.inFlight` only works if the record hasn't been evicted. Consider explicitly clearing `inFlight` on completion (both success and failure paths) rather than relying on record replacement:

  ```js
  // In the .then(data => { ... }) block:
  const rec = s.tiles.get(key);
  if (rec) delete rec.inFlight;
  s.tiles.set(key, { data, fetchedAt: now() });
  
  // Similarly in the .catch:
  const prev = s.tiles.get(key);
  if (prev) delete prev.inFlight;
  s.tiles.set(key, { ... });
  ```

- **[`utils/mapTileCache.js:57–68`]** `computeUnion` deduplicates by `item.id`, but the tile-fetched marker objects for entrances and networks use `id` as a numeric identifier. There is no guarantee that organizations fetched from the tile cache also have an `id` field — the comments note organizations are fetched via `getMapGrottosUrl`, and the inline `fetchAllOrganizationsCoordinates` strips them down to `[longitude, latitude]` arrays (no `id`). However the *tile cache path* (`fetchOrganizations` → `fetchForBounds`) dispatches the raw API response as `FETCH_MAP_ORGANIZATIONS_SUCCESS`, which may include full objects with `id`. These are two different redux keys (`state.map.organizations` vs `state.map.organizationsCoordinates`), so `computeUnion` will never see the stripped arrays. That said, if any API response returns items without an `id` field, `seen.has(undefined)` will be true after the first such item and all subsequent id-less items will be silently deduplicated/dropped. Add a guard:

  ```js
  if (item == null) continue;
  if (item.id != null && seen.has(item.id)) continue;
  if (item.id != null) seen.add(item.id);
  out.push(item);
  ```

- **[`middlewares/mapCacheInvalidationMiddleware.js:35–65`]** Action type strings are hardcoded literals instead of imported constants. If any of the upstream action creators rename their constant (e.g. `POST_ENTRANCE_SUCCESS` → `CREATE_ENTRANCE_SUCCESS`), the middleware silently stops invalidating the right tiles with no compile-time warning. Import from the relevant action files:

  ```js
  import { POST_ENTRANCE_SUCCESS, UPDATE_ENTRANCE_SUCCESS, DELETE_ENTRANCE_SUCCESS } from '../actions/Entrance';
  ```

---

- **[`components/common/Maps/MapClusters/ClusterLayer.jsx:95–98`]** The `refresh` callback includes `type`, `pane`, and `onLeafClick` in its dependency array, which is correct. But `type` and `pane` are strings (cheap to compare) while `onLeafClick` will cause `refresh` to be recreated every time the parent renders if it isn't memoized. In the current callers none of them pass `onLeafClick`, so this is latent. Worth noting in the JSDoc that callers must memoize `onLeafClick` to avoid unnecessary cluster re-renders.

- **[`utils/mapTileCache.js` — module-level `state` object]** The module-level `state` singleton means that if the Redux store is recreated (e.g. in tests, or if `ApplicationShell` ever unmounts and remounts), `s.dispatch` could point to a stale dispatch. The `_resetForTests` export helps in tests, but there's no lifecycle hook tied to store creation. Consider accepting `dispatch` at `registerEntity` time (or passing it through a context) rather than capturing it per-call in `fetchForBounds`. At minimum, a comment explaining the lifecycle assumption would help future contributors.

- **[`components/common/Maps/MapClusters/ClusterLayer.jsx:205–212`]** `ClusterLayer.propTypes` marks `data` as `PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.number))` but doesn't set it as required. An undefined `data` will return `null` from `useCluster` (as designed), which means a misconfigured caller silently renders nothing. Given that all four callers in `index.jsx` always pass a (potentially empty) array from Redux state, this is fine in practice — but marking `data` as `isRequired` (defaulting to `[]` via JS default parameters) would make misconfiguration visible earlier.

- **[`utils/mapTileCache.js:153–183`]** `fetchForBounds` ignores tiles that already have a valid in-flight request (no explicit check — it relies on `fetchTile`'s early return). This is correct. However there's a subtle issue with `scheduleEmit` being called at the top of `fetchForBounds` (line 161) *before* checking which tiles need fetching. If all tiles are fresh (cache hit), `scheduleEmit` will schedule a dispatch only to be short-circuited by the `lastEmittedVersion === tilesVersion` guard — that's fine, it's the intended optimization. But if there are *no* tiles in the cache yet (first load), `scheduleEmit` dispatches an empty union immediately before any tiles arrive. The reducer sets `organizations: []` which is the same as its initial state, so there's no visible regression, but it does add an unnecessary Redux dispatch on every initial `moveend`. Worth a comment explaining this is intentional (sending the initial empty state to trigger any loading UI).

- **[`components/common/Maps/MapClusters/index.jsx:348–354`]** `clusterConfigs` is defined inline in the render function body, which means it creates a new array every render. The array itself is small and cheap, but each `ClusterLayer` receives a new `data` reference from Redux state on every re-render of `HydratedMap`. Since `useCluster` memoizes on `data` reference equality, and Redux state is stable (immutable updates), this is fine — the clusters only rebuild when the data actually changes. Just worth confirming that `entrances`, `networks`, etc. props are already stable references coming from `useSelector`.

- **[`utils/tileMath.js`]** No unit tests for `latLngToTile`, `tileToBounds`, or `tilesForBounds`. These are pure math functions that are easy to test and critical to correct cache behaviour (wrong tile boundaries → missed or double-fetched tiles). Edge cases worth covering: lat/lng at the exact anti-meridian (±180), poles (clamped to ±85), zoom=0, and a viewport crossing the anti-meridian. The `mapTileCache.js` module similarly has no tests. These would be a great candidate for fast-check property tests (e.g. `tileToBounds(latLngToTile(lat, lng, z), z)` should contain `(lat, lng)` for all valid inputs).

---

### Nitpicks (Optional)

- **[`components/common/Maps/MapClusters/ClusterLayer.jsx:12`]** `import * as L from 'leaflet'` uses the namespace import, while the rest of the codebase mixes `import L from 'leaflet'` (default) and `import * as L from 'leaflet'` (namespace). The deleted `useHeatLayer.jsx` specifically called out why default import is needed for `@asymmetrik/leaflet-d3` augmentation. That concern is gone now, but for consistency with the majority of files, consider using `import L from 'leaflet'`.

- **[`components/common/Maps/common/MapContainer.jsx:210`]** `renderer ?? undefined` is redundant: `null ?? undefined` evaluates to `undefined`, and passing `undefined` to a React prop is the same as not passing it at all. `renderer={renderer || undefined}` would be slightly clearer, or just document that `null` is passed through intentionally (it isn't — `null` would override Leaflet's default renderer with null, which would likely crash). The current `?? undefined` pattern is safe but confusing.

- **[`components/common/Maps/common/Markers/tooltipHelpers.js`]** The `escapeHtml` function doesn't escape `"` (double quotes) or `'` (single quotes). This is fine for tooltip HTML content rendered as text nodes, but if `iconSrc` ever comes from user input (currently it doesn't — it's always a static import), an unescaped `"` in the `src` attribute would break the HTML. Low risk given current callers, but worth a note.

- **[`components/common/Maps/MapClusters/ClusterLayer.jsx:51–57`]** The CSS has a duplicate `.cluster-bubble:hover` rule — one setting `transform: scale(1.08)` and another (at the end of the block) setting `filter: brightness(1.15)`. Both apply on hover, which is the intended behaviour, but combining them into a single rule would be cleaner.

- **[`components/common/Maps/MapClusters/index.jsx` — localStorage key migration]** Existing users have `grottocenter_selectedHeats` and `grottocenter_selectedMarkers` in their localStorage. These are now orphaned (never read, never written). The new `grottocenter_selectedLayers` key starts from `DEFAULT_SELECTED_LAYERS` (only entrances on). Users who had previously enabled networks or organizations will silently lose that preference on first load. Not a bug, but a UX degradation for existing users. A one-time migration (read old keys, merge into new key, delete old keys) would be considerate.

urien (approved):
Parfait pour moi

ClemRz (changes requested):
Excellent work — this is the cleanest performance refactor on the map in a long time. The supercluster migration eliminates the entire d3-hexbin dependency, the tile cache design is genuinely novel (version counter + microtask coalescing is a sharp pattern), and all three must-fix items from the previous review are correctly addressed. The comments are unusually detailed and the test coverage for `tileMath.js` is thorough. A few new things to address before merge:

---

### Issues (Must Fix)

- **[`components/common/Maps/MapClusters/index.jsx:13`]** `LAYER_TYPES_LIST` is imported from `DataControl` but never used in this file. ESLint already flags it as `no-unused-vars`. The import should be removed.

  ```suggestion
  import DataControl, { layerTypes } from './DataControl';
  ```

- **[`middlewares/mapCacheInvalidationMiddleware.js:44–46`]** `POST_ENTRANCE_SUCCESS` is handled with `invalidateAtOrFallback('entrances', action.type, action.data)`, but `CreateEntrance.js` dispatches `{ data, type: POST_ENTRANCE_SUCCESS }` where `data` is the raw API response — and the API returns the full entrance object which includes `latitude` and `longitude`. This part is fine. However `POST_ORGANIZATION_SUCCESS` is handled with `action.organization` (the field name used in `postOrganizationSuccess`), while the organization API response (`action.organization`) likely does not carry `latitude`/`longitude` — it's a Grotto/organization record, not a geo-point. Looking at `CreateOrganization.js`, the `organization` object comes directly from `response.json()` with no coordinate extraction. If the organization API response does not include `latitude`/`longitude`, `invalidateAtOrFallback` will always fall back to `invalidateAll('organizations')` with a dev warning on every creation. Either document that the org API response carries these fields (link to the Swagger spec), or remove the targeted invalidation attempt and call `invalidateAll('organizations')` directly to avoid the misleading `console.warn`.

- **[`middlewares/mapCacheInvalidationMiddleware.js`]** Massif mutations (`POST_MASSIF_SUCCESS`, `UPDATE_MASSIF_SUCCESS`, `DELETE_MASSIF_SUCCESS`, `DELETE_MASSIF_PERMANENT_SUCCESS`) are not handled. The massif cluster layer feeds from `massifsCoordinates`, which is populated by `fetchAllMassifsCoordinates()` — a one-shot bulk fetch, not the tile cache — so `invalidateAll('massifs')` wouldn't apply here. But a user creating or deleting a massif would see the cluster layer stay stale until the next page reload. At minimum, a comment in the middleware explaining that massifs use a one-shot fetch (not the tile cache) and therefore don't need invalidation here would prevent future contributors from wondering if this is a gap.

---

- **[`components/common/Maps/MapClusters/ClusterLayer.jsx:206–211`]** The cleanup `useEffect` captures `markersRef.current` in a local variable at effect-registration time and uses it in the cleanup — which is the correct pattern. However the `refresh` `useEffect` (lines 197–199) does not do the same: it calls `refresh()` directly on every `refresh` recreation. Since `refresh` already performs its own `markersRef.current` reads synchronously inside its body, this is safe in practice — but it means every change to `enabled`, `supercluster`, `pane`, or `onLeafClick` triggers a full marker rebuild even for cases like `enabled=false` where the early-return short-circuits quickly. This is intentional and correct; worth a comment confirming it.

- **[`utils/mapTileCache.js:215–240`]** `invalidateAll` and `invalidateTileAt` both call `fetchForBounds(entity, s.lastBounds, s.lastApiZoom, s.dispatch)` if a prior `lastBounds` is recorded. Since `fetchForBounds` calls `scheduleEmit` up-front, this will trigger a Redux dispatch on every tile invalidation even when nothing is in the cache yet (e.g. if the map was never in the viewport that covers the mutated entity). This is harmless — the union would be empty — but it adds a needless re-render. The `s.tilesVersion` guard in `scheduleEmit` normally short-circuits this, but `invalidateAll` sets `fetchedAt = 0` on all tiles *without* bumping `tilesVersion`, so `scheduleEmit` inside `fetchForBounds` fires with `lastEmittedVersion === tilesVersion`, correctly short-circuiting. That's the intended behavior. A brief comment on why `invalidateAll` doesn't bump `tilesVersion` would make this easier to follow.

- **[`components/common/Maps/MapClusters/index.jsx:375–386`]** `clusterConfigs` is constructed inline in the render body. Each render creates a new array and four new object literals — all trivially cheap — but each `ClusterLayer`'s `data` prop comes from `entrances`, `networks`, etc. which are stable Redux references (they only change when new tile data arrives). The `enabled` computation `!!selectedLayers[layer] && !off` is also cheap. No memoization is needed here, but worth a quick comment that `data` references are stable so `useCluster` won't rebuild the kD-tree on every render.

---

### Nitpicks (Optional)

- **[`components/common/Maps/MapClusters/index.jsx:10`]** `* as L from 'leaflet'` is used only at line 486 for `L.canvas(...)`. The existing codebase convention (and `ClusterLayer.jsx` itself) uses the default import `import L from 'leaflet'` — the namespace import is a legacy of the `@asymmetrik/leaflet-d3` augmentation concern that was just removed. Consider switching to the default import for consistency.

  ```suggestion
  import L from 'leaflet';
  ```

- **[`components/common/Maps/common/MapContainer.jsx:261`]** `renderer: PropTypes.object` is very broad. `PropTypes.instanceOf(Object)` or even `PropTypes.shape({})` would be more precise, but a comment explaining it's a Leaflet renderer instance (`L.canvas()` / `L.svg()`) would be enough to guide future callers.

- **[`utils/mapTileCache.js:166–167`]** The `fetchForBounds` function stores `bounds` (a Leaflet `LatLngBounds` object) in `s.lastBounds` and later calls `tilesForBounds(bounds, CACHE_ZOOM)`. But `tilesForBounds` expects `{ sw_lat, sw_lng, ne_lat, ne_lng }` (a plain object), not a Leaflet `LatLngBounds`. Looking at the callers in `Map.jsx`, `handleUpdate` passes `map.getBounds()` wrapped into a criteria object with those flat keys — so this is actually fine because `criteria` (not the Leaflet bounds) is what's stored. Just worth confirming the type contract in a comment on `lastBounds`.

ClemRz (changes requested):
This is a high-quality follow-up round. All must-fix and should-consider items from the previous review have been addressed cleanly, including the `clearInFlight` guard, the `computeUnion` id-less-item fix, the imported action constants, and the `POST_ORGANIZATION_SUCCESS` direct `invalidateAll` call. Tests pass (639/639), lint is clean, and the staged preview confirms the visual output.

A few small things left worth calling out, but nothing blocking.

---

### Issues (Must Fix)

- **[`components/appli/Massif/MapMassif.jsx:84`]** The low-zoom `fetchMarkers` path calls `markersRef.current(null)` to clear real markers, but this runs **before** `setShowClusters(true)` and the coordinate fetch. The comment says "at low zoom the cluster layer takes over via the JSX", but if the user zooms back out while an in-flight coordinate fetch is running, the fetch resolves and calls `setEntranceCoordinates(data)` followed by `fetchMarkersRef.current()`. At that point the zoom is below `MARKERS_LIMIT` again and `fetchMarkersRef.current()` re-enters the low-zoom branch — which is fine. However the `abortRef` inside `fetchMarkers` is only assigned in the high-zoom branch (line 98: `abortRef.current = new AbortController()`), so the low-zoom path never calls `abortRef.current?.abort()` before clearing markers. If the component unmounts mid-flight the only cleanup is the `useEffect` abort at the top (line 55) — that one does fire correctly. This is actually safe today but the two `AbortController` sites (one at mount cleanup, one inside `fetchMarkers`) create a subtle conceptual mismatch: the one inside `fetchMarkers` only covers the high-zoom real-marker fetch, yet the outer `useEffect` cleanup aborts it on unmount regardless. Not a bug in the current code, but worth a short comment in `fetchMarkers` explaining that `abortRef` is for the high-zoom marker fetch only, and the coordinates fetch (lines 118–136) manages its own `AbortController` separately.

---

- **[`components/common/Maps/MapClusters/index.jsx:99-101`]** `HydratedMap` receives `onUpdate` as a required callback (it's called unconditionally at `onUpdateRef.current(...)`) but the PropType is `PropTypes.func` with no `.isRequired`. If a parent forgets to pass it, the app crashes silently on the first `moveend`. Marking it `isRequired` would surface the misconfiguration earlier.

- **[`utils/mapTileCache.js:184–196`]** Inside `fetchForBounds`, the logic for deciding whether to call `fetchTile` only checks `!rec || rec.fetchedAt === 0` (hard-miss) then TTL/failure-cooldown. There's no explicit check for `rec.inFlight` in the outer loop — that guard lives inside `fetchTile` itself (line 118: `if (existing?.inFlight) return existing.inFlight`). This is correct, but the inner-loop comment says "MISSING (no record OR fetchedAt=0 from invalidation)" without mentioning that the `inFlight` dedup is delegated to `fetchTile`. A one-line comment noting the delegation would help readers who are tracing the "why doesn't this loop issue duplicate requests?" question.

- **[`components/common/Maps/MapClusters/ClusterLayer.jsx:159–167`]** When clicking a cluster, the target zoom is computed as `Math.max(expansion + 2, map.getZoom() + 3)`. At `MARKERS_LIMIT - 1` (zoom 12), this always jumps to at least zoom 15, which skips the `MARKERS_LIMIT` threshold and lands directly in real-marker territory. That's likely intentional — clicking a cluster should always resolve it. Worth a brief comment confirming this is deliberate so it doesn't get "corrected" later.

- **[`utils/mapTileCache.js` — no unit tests]** `mapTileCache.js` is the most complex new module and the only one without a test file. The `computeUnion` dedup logic (the `item.id != null` guard), `evictIfNeeded` LRU behaviour, and the `scheduleEmit` version-gate short-circuit are all testable in isolation and were exactly where bugs were found in the previous review round. Worth adding a `mapTileCache.test.js` before this merges — even 8–10 unit tests would cover the critical paths and act as a regression net for the next person who touches this file.

---

### Nitpicks (Optional)

- **[`components/common/Maps/MapClusters/index.jsx:520–524`]** `HydratedMap.propTypes` lists `entrances`, `networks`, `organizations`, `massifs` (all `PropTypes.arrayOf(...)`) without `.isRequired`. The default values in the destructuring cover the `undefined` case, so this won't crash — but a stale-reducer bug that returns `undefined` instead of `[]` would silently pass through `useCluster`, which guards on `!Array.isArray(data)`. Since the defaults are set in the function signature, the PropTypes can stay optional; just noting it's there.

- **[`utils/mapTileCache.js:45`]** `ensureEntity` uses `if (!state[entity])` which treats a falsy but defined value as missing. In practice `state[entity]` is always either `undefined` (not yet set) or a proper object, so this is fine. But `Object.prototype.hasOwnProperty.call(state, entity)` or `state[entity] === undefined` would be marginally more explicit about intent.

- **[`components/common/Maps/MapClusters/index.jsx:483`]** `L.canvas({ padding: 0.5 })` is created inside `useMemo([], [])` — effectively once per mount. The comment correctly explains the rationale. Worth adding `// eslint-disable-next-line react-hooks/exhaustive-deps` here is not needed since the dep array is truly empty by design (fresh instance per mount), but the comment could mention that the empty dep array is intentional and not a missing-dependency oversight.

ClemRz (changes requested):
Excellent final state. All three rounds of must-fix items have landed cleanly — the `clearInFlight` guard, the `computeUnion` id-less fix, the imported action constants, and the `POST_ORGANIZATION_SUCCESS` direct `invalidateAll` call are all in place. Tests pass (658/658), lint is clean, and the staged preview confirms the visual output.

A few remaining things worth calling out before merge.

---

### Issues (Must Fix)

- **[`components/common/Maps/common/Markers/tooltipHelpers.js:9`]** `makeIconTooltip` interpolates `iconSrc` directly into the `src` attribute of an `` tag inside an HTML string that is set as `innerHTML` by Leaflet's `bindTooltip`. Today `iconSrc` always comes from `assets/icons/*.svg` imports (bundled as static paths by Vite), so there is no immediate XSS risk. However the function signature accepts any `iconSrc` value and has no guard. If a future caller passes a user-controlled URL (e.g. a caver's avatar from the API), the `src` attribute itself is fine (images can't execute scripts), **but** there is no escape applied, meaning an attacker could break out of the attribute with a well-placed `"` and inject arbitrary HTML. The fix is a one-liner — apply the same `escapeHtml` to `iconSrc` as is already applied to `name`:

  ```js
  export const makeIconTooltip = (iconSrc, name) =>
    `<span style="display:flex;align-items:center;gap:6px"><img src="${escapeHtml(iconSrc ?? '')}" width="16" height="16">${escapeHtml(name ?? '—')}</span>`;
  ```

---

- **[`components/common/Maps/MapClusters/MassifPolygons.jsx:3`]** Uses `import * as L from 'leaflet'` while all other new files added in this PR (`ClusterLayer.jsx`, `index.jsx`) use the default `import L from 'leaflet'`. The existing codebase is split (most files use namespace import), but since this PR introduced `import L from 'leaflet'` in two new files as the new convention, `MassifPolygons.jsx` is a minor inconsistency — it was modified in this PR, so it was touched anyway. Not a bug, but worth standardising to `import L from 'leaflet'` for the files this PR modifies.

- **[`actions/Map.js:248-285`]** `fetchMassifs` is still debounced via `redux-debounced` (`.meta.debounce`). This was intentional since massifs aren't tile-cached and still fire on every `moveend` at polygon zoom. The comment at line 121 says "so `redux-debounced` is no longer needed here — the request rate is bounded by the tile grid" — that's true for entrances/networks/organizations, but `fetchMassifs` was not migrated to the tile cache and still needs the debounce. Consider adding a clarifying comment on `fetchMassifs` noting that the debounce is intentionally kept (unlike the tile-cached thunks) so the rationale doesn't get misread as a missed migration.

- **[`utils/mapTileCache.test.js` — missing `clearInFlight` race test]** The `clearInFlight` guard was a must-fix in a previous review round and the comment in the code is thorough. However there's no test that exercises the eviction-during-in-flight scenario it protects against: a tile is fetched, `evictIfNeeded` runs and deletes the record mid-flight, then the fetch resolves. Adding a test here would lock in the guard's behaviour as a regression net. It's a tricky scenario to simulate (requires mocking `MAX_TILES` or filling the cache first), so leaving it as a suggestion rather than a blocker.

---

### Nitpicks (Optional)

- **[`components/common/Maps/MapClusters/ClusterLayer.jsx:87`]** The `refresh` callback's `useCallback` dep array includes `onLeafClick`. The JSDoc correctly notes callers should memoize it. Consider also adding it to the `ClusterLayer.propTypes` JSDoc comment or a `// callers must memoize onLeafClick` inline note directly above the prop definition so the warning is co-located with the type declaration and easier to find during code search.

- **[`utils/mapTileCache.js:33`]** `ensureEntity` uses `if (!state[entity])` which treats a falsy-but-defined value as absent. In practice `state[entity]` is always `undefined` or a proper object, so this is fine today. `state[entity] === undefined` would be more explicit about intent (already noted as a nitpick in the prior review — flagging again for completeness since it was described as "informational only" but not acted on).

- **[`components/common/Maps/MapClusters/Markers.jsx:3`]** `import * as L from 'leaflet'` is not present here (just a note: `Markers.jsx` does *not* import leaflet, so this file is clean). Keeping the nitpick scope to the files that do have the inconsistency mentioned in Suggestions.

ClemRz (changes requested):
This is a clean final state. All six items from round 4 are correctly resolved: the XSS fix (`iconSrc` is now escaped), the debounce comment on `fetchMassifs`, the `clearInFlight` race test, the `onLeafClick` memoization note, and the `ensureEntity` explicitness. Tests pass (658/658 at last run), lint is clean, and the staged preview confirms the visual output.

Two small things left before merge.

---

### Issues (Must Fix)

- **[`components/appli/Massif/MapMassif.jsx:125`]** The comment on the one-shot coordinate fetch still reads _"The hexbin layer filters client-side"_. The hexbin layer was removed in this PR — it should say something like "The ClusterLayer queries the visible bbox on each moveend from the kD-tree". Leaving stale references to the deleted hexbin layer will mislead the next person who reads this code and wonders where that layer went.

  ```suggestion
    // One-shot coordinate fetch for the entire massif bbox.
    // The ClusterLayer builds a kD-tree from these coordinates and queries
    // only the visible bbox on each moveend — no re-fetch needed on pan/zoom.
  ```

---

- **[`components/common/Maps/MapClusters/ExploredOverlay.jsx:5`]** `import * as L from 'leaflet'` — this file was modified in this PR (the `tooltipHelpers` import was added). The rest of the files touched by this PR (`ClusterLayer.jsx`, `index.jsx`, `MassifPolygons.jsx`) all use the default `import L from 'leaflet'`. Switching `ExploredOverlay.jsx` here would make every file modified in this PR consistent with the new convention.

  ```suggestion
  import L from 'leaflet';
  ```

---

### Nitpicks (Optional)

- **[`components/appli/Massif/MapMassif.jsx:128`]** Two lines later in the same comment block: _"heatmap is displayed immediately if still at low zoom"_ — another stale hexbin/heatmap reference. Consider updating to "cluster layer is shown immediately…" while touching the block for the must-fix above.

ClemRz (approved):
This is a clean resolution of all the items from round 5. Both the must-fix (stale hexbin comment in `MapMassif.jsx`) and the suggestion (`ExploredOverlay.jsx` import convention) have been applied correctly in commit `ecf17077`. The README, Storybook story name, and the `index.jsx` comment are all updated. No remaining stale references to the removed hexbin layer.

The full PR is in excellent shape: the `supercluster` migration, tile cache architecture, `clearInFlight` guard, `computeUnion` id-less fix, imported action constants, and test coverage for both `tileMath.js` and `mapTileCache.js` are all solid. Ready to merge.

---

### Nitpicks (Optional)

- **[commit messages]** The four `fix: PR comments` commits are missing a scope — commitlint's `scope-empty: [error, never]` rule requires one (e.g. `fix(map): address PR review comments`). Since the PR will be squash-merged into `develop`, this only affects the branch history, not the final commit. Worth a note for future fix-up rounds.

PR #1476, opened 2026-07-30
feat(Document): rework collection children display and separate destructive actions

Paul-AUB:
## 🤔 What

Rework of the document page, mostly around how a collection presents the documents it contains.

**Collection / article children**

- New `DocumentChildrenList` component rendering a document's children in three flavours: issues of a collection (tile grid), articles (list), and anything else.
- A collection's issues are shown as a shelf of tiles: the issue designation leads, the publication date sits underneath.
- File-availability badge on children that carry an attached file, with a legend naming the marker once per section.
- Per-section sort control — _Newest first_ / _Oldest first_ / title — shown only when reordering can actually change the list.
- The collection header now states the span its issues cover (_Published from {start} to {end}_, or _Published in {year}_).

**Shared layout**

- `ResponsiveActions`: the `divider` item prop is replaced by `destructive`. Destructive actions get their own `ButtonGroup` on desktop and a divider in the mobile menu, so Delete no longer sits flush against Print or Page history. Applied to Delete on Entrance, Network, Massif, Organization and Person.
- `ScrollableContent` exports its `CountBadge` so a heading rendered outside a card can carry the same badge.

**Redux / types**

- `fetchDocumentChildren` no longer sorts its payload and no longer takes a locale.
- New `DocumentChildPropTypes` and `FilePropTypes`; `authorizationDocument` retyped from `string` to the document shape it actually holds.

**Fixes and cleanup**

- `SummaryText` no longer renders an empty block for an empty description (it used to leave a stray gap above the next section).
- Removed the unused `LicenseLink` styled wrapper in `AddFileForm`.
- 7 new i18n keys, translated across all 15 language files.

## 🤷‍♂️ Why

A collection page was a flat alphabetical list of child titles, and those titles are highly redundant by construction: every issue of _Scialet_ is titled `Scialet No 47 (2018)`, so each row restated the collection name the reader already had in front of them, and very often repeated it a third time as a description. Chronology — the central descriptive element of a serial — was nowhere: the list was ordered by title, which puts `No 10` next to `No 1`.

Separately, there was no way to tell which issues actually have a file attached without opening them one by one.

On the shared-layout side, the previous `divider` prop asked every caller to insert the separator itself, in the right position. Six pages build their own action list, and a rule that has to be repeated six times is a rule that gets forgotten — hence the switch to marking the item instead.

## 🔍 How

**Labels — `utils/documentChildrenLabel.js`**

`getChildLabel` is strictly subtractive: it removes the run of leading words a child's title shares with the collection title, plus the punctuation that run leaves behind. A title departing from the collection's pattern shares no leading word and comes back untouched, and the last word is never removed, so a label can never end up empty.

It deliberately does **not** strip a trailing year. That was tried and measured against 6148 real issues: it cut `(Jan.-Mar. 2005)` down to `(Jan.-Mar.`, left orphan separators behind, and collapsed `Monthly Newsletter - February 1996` and `… February 1993` onto the same label — two different issues rendered indistinguishable. In a date, the year is what identifies the issue. It is dropped from the _display_ only when the designation already states it.

`hasOwnDescription` hides a description that is already contained in the title, which is the common case for imported children.

**Sorting — `utils/documentChildrenSort.js`**

`datePublication` is a truncated ISO string (`2011`, `2011-06`, `2011-06-15`), so plain string comparison orders it chronologically — no parsing, and no risk of reading a year-only value as a timestamp. Titles are the tie-breaker, through a single `Intl.Collator` built once per sort: passing a locale and options to `localeCompare` bypasses the engine's cached collator and rebuilds one on every comparison, which is n log n of them on a collection holding thousands of issues. Undated documents always close the list, in both directions.

`canReorderDocumentChildren` is what gates the sort control: since the comparators fall back to the title on equal dates, a collection whose documents all share one date produces the same list under all three orders. It lives next to the comparators on purpose — the answer holds only for as long as they behave that way.

Sorting moved out of the action and into the view. Doing it in both places meant two owners for one decision, and froze the collation to whichever locale happened to be active at fetch time.

**Availability marker**

Derived from attached files only. Detecting a URL inside the free-text description was tried and dropped: `.la`, `.at`, `.be`, `.it` and `.co` are real TLDs, so French prose with a missing space after a full stop (`l'aval.On`) was flagged about as often as a genuine link. Only presence is marked — badging absence would put an identical icon on the great majority of rows, which is the noise the indicator exists to replace; the legend is what makes the absence readable. It is a chip rather than a bare icon, and has no tooltip, so it reads as a statement about the item rather than something to click.

**Tile sizing**

The type tier is chosen once per grid, from the longest label in it. Sizing each tile on its own label put `N° 9, Mars` at display size next to `N° 17, Février` at two thirds of it — the same kind of item rendered two different ways inside one row.

**Tests**

446 lines of unit tests covering the label derivation and the comparators, including the redundancy and empty-label edge cases described above.

## 🔗 Related API PR

None — this PR consumes `GET /documents/{id}/children` as it already exists.

## 🧪 Testing

`yarn lint` and `yarn test` pass (699 tests).

Manual checks worth doing:

1. Open a collection with many issues (e.g. a bulletin series) and confirm the tiles show the issue designation rather than the full repeated title.
2. Switch the sort between _Newest first_, _Oldest first_ and title, and confirm undated issues stay at the end in both date orders.
3. Open a collection where every issue carries the same date — the sort control should not appear at all.
4. Confirm the file badge only appears on issues that really have a file attached, and that the legend matches.
5. On a document page with delete rights, check that Delete is visually separated from the other actions — as a distinct button group on desktop, and after a divider in the overflow menu on mobile.
6. Check the same separation on Entrance, Network, Massif, Organization and Person pages.

## 📸 Previews
image
image
image
image


ClemRz (changes requested):
### Issues (Must Fix)

- **[src/pages/DocumentDetails/index.jsx:~280]** `sortedChildOther` is computed with `DEFAULT_CHILDREN_SORT_ORDER` (newest-first) but `childOther` is passed to `<ChildrenControls>` with no `onSortOrderChange` handler — so the control correctly hides the sort select, but the "child documents" section's order will silently differ from the "articles" section when both exist on the same document. More importantly, `childOther` is unsorted from the store; without a sort handler a user has no way to reorder and may not realise they are seeing fetch order. If intentional (i.e. alphabetical default is wanted here), a comment should explain it; if not, a sort state + handler should be added like the articles section.

- **[src/pages/DocumentDetails/index.jsx:~811]** Sort state (`issuesSortOrder`, `articlesSortOrder`) is held in the `Document` component and never reset when `documentId` changes. `DocumentDetails` fetches new data whenever `documentId` changes (line ~810), but `Document` receives the new `documentChildren` while keeping the old sort orders. On a fast in-page navigation (breadcrumb to parent → back → sibling collection) the sort select could show stale "Oldest first" while the tiles render the new collection. Consider resetting sort state in `DocumentDetails` or passing a `key={documentId}` to `Document` to force a remount.

- **[src/utils/documentChildrenLabel.js:~48-66]** `getChildLabel` mutates the `words` array derived from `title.split()` in-place with `words.shift()`. While the array is a fresh copy created by `slice`, the mutation is unexpected in a pure utility. More critically: the label re-computes on every render call (called inside `useMemo` in `DocumentChildrenTiles`) and the mutation-based loop does not affect correctness, but the pattern is worth fixing for clarity. Low-risk correctness-wise, but worth noting.

- **[src/pages/DocumentDetails/DocumentChildrenList.jsx:~93-100]** `ChildrenControls` receives `documents` only to drive two `useMemo` calls (`hasFiles`, `showSort`). Both scan the full array on every render of the parent. Since `DocumentChildrenTiles` and `DocumentChildrenList` already receive the pre-sorted list, consider memoizing `hasFiles` and `canReorder` in `index.jsx` alongside the `sortedChild*` memos and passing the derived booleans down instead of the raw documents array. This avoids redundant array scans.

- **[src/pages/DocumentDetails/DocumentChildrenList.jsx:~268-285]** `DocumentTile` uses `Tooltip` with `title={tooltip}` where `tooltip` is built from `doc.title` + optional description. When `doc.title` is undefined (which `DocumentChildPropTypes` allows), `tooltip` falls back to an empty string and `aria-label` is set to `undefined`. The tile then has no accessible name at all. Consider a fallback like `aria-label={tooltip || formatMessage({ id: 'Document' })}`.

- **[src/pages/DocumentDetails/DocumentChildrenList.jsx:~312-330]** The `ChildrenSortSelect` `renderValue` produces `<><SwapVertIcon />{label}</>`, which wraps the icon and label in a React Fragment as the trigger value — this is not natively `aria`-labelled. The `inputProps['aria-label']` covers screen readers, but consider also adding a `SelectDisplayProps={{ 'aria-label': formatMessage({ id: 'Sort by' }) }}` to ensure the native role is picked up by VoiceOver/NVDA on the visible `<div role="button">` element.

- **[src/pages/DocumentDetails/index.jsx:~387-394]** The `coverage` node (holdings statement) uses `breadcrumb ?? coverage` in `subheader`. A collection that _has_ a parent document would silently omit the date span. The comment says "navigation wins," which is a valid design choice, but a collection with a parent is an uncommon-but-valid database state. Consider at minimum asserting the assumption in a comment (e.g., "collections are expected to be root documents") or combining them: `breadcrumb` on one line and `coverage` as a second subtitle line.

- **[src/components/common/Layouts/ResponsiveActions.jsx:~48]** The new desktop grouping uses `group[0].key` as the `ButtonGroup`'s React `key`. If the first item in a group is hidden (its `hidden` flag was used before `visibleItems` was computed, so this can't happen here — but it's subtle). The logic is correct since `visibleItems` is already filtered, but the comment at the filter site (`items?.filter(item => !item.hidden)`) is far away from the grouping logic. A small inline comment there would prevent a future reader from adding a hidden item without realising it could shift group keys.

- **[src/utils/documentChildrenSort.js:~20-28]** `compareDates` returns `dateA < dateB ? -direction : direction`. When `direction = -1` (DESC), the result is `-1 * -1 = 1` (a > b), meaning newer dates sort last. That appears reversed — verify this is intentional. Looking at the test: `[doc('b','2001'), doc('c','2003'), doc('a','2002')]` sorted DESC gives `['c','a','b']`, confirming newest first, so `direction = -1` does produce DESC. The math works but the sign convention is non-obvious; a brief comment on the sign would help.

- **[src/pages/DocumentDetails/Section.jsx:~36-49]** `TextLink` returns a React Fragment (`<>...</>`) when an `icon` is present. Since `TextLink` is used inline inside `Typography` and `ChildTitle` (which is itself a `Typography`), a Fragment as inline content is fine — but it means the icon's wrapping `Box` (with `display: 'inline-flex'`) competes with the parent's `inline-flex` / `flex` layout. This causes a minor vertical-alignment inconsistency for the icon on the `Authorization` detail row (an icon-wrapped link inside a `Property`). Consider using `component="span"` with `sx={{ display: 'inline-flex', alignItems: 'center', gap: 0.25 }}` as a single wrapper instead of the extra Fragment.

### Nitpicks (Optional)

- **[src/pages/DocumentDetails/DocumentChildrenList.jsx:~174]** The `ChildrenControls` JSDoc says "With neither, the component returns null rather than an empty row" — but the component also returns null when `documents` is undefined/null. Worth documenting that null-document input is safe.

- **[src/utils/documentChildrenLabel.js:~28]** `trimSeparators` does not escape the `-` in the character class — it happens to be at the start (`[\s\-–—...]`) which is unambiguous, but a trailing or mid-position `-` would need escaping. It's fine as written; just a minor regex style note.

- **[src/pages/DocumentDetails/index.jsx:~440-441]** The comment `{/* One section each, so no SectionStack... */}` is accurate and useful. However, the deleted/loading/error sections each render without any vertical spacing between them and the next sibling. If both `isLoading` and `error` somehow resolved together (edge case), the two `ScrollableContent` cards would be flush against each other with no gap. `SectionStack` exists precisely for this. Consider wrapping the three in a conditional `SectionStack` or adding a `sx={{ mb: 1 }}` to each.

- **[packages/web-app/public/lang/*.json]** All 15 language files received the same 7 new keys. The `"Geographic coverage"` key is a new label for the previously-named `"Regions"` field. The old `"Regions"` key is still present in the files (it's used elsewhere), so there is no regression — but it may be worth a follow-up search for any remaining `id: 'Regions'` usage that conceptually means iso3166 coverage rather than an actual region list.

ClemRz (changes requested):
All previous issues addressed cleanly. The fixes are solid and the explanations on record for the unchanged items are convincing. A few new observations below.

### Issues (Must Fix)

- **[packages/web-app/public/lang/*.json]** `"No file attached"` is added to all 15 language files but is never referenced in any source file (confirmed by grep). The existing longer key `"No file attached — please check that it is not an oversight."` is the one actually used in `EntitiesForm/Document/index.jsx`. Dead i18n keys accumulate translation debt for every future language update. Either wire the short key up where it was intended, or remove it from all 15 files.

- **[src/utils/documentChildrenLabel.test.js]** `hasOwnDescription` and `getIssuesYearRange` are exported and used by the UI but have no dedicated tests. `hasOwnDescription` has a subtle case — a description that is a *substring* of the title is hidden, while one that merely *overlaps* should show — and `getIssuesYearRange` has the min/max-on-strings invariant that is easy to break if the format ever changes. Both are worth a few test cases.

- **[src/pages/DocumentDetails/DocumentChildrenList.jsx:200-211]** The legend `MarkerChip` in `ChildrenControls` renders the "File available" label as visible text alongside an icon, which is fine for sighted users. However, the chip itself has no `aria-label` or `role="status"` — it is not interactive, so the visual text is read, but the `InsertDriveFile` icon before it has no alt text. Since the icon is decorative here (the label already says "File available"), adding `aria-hidden="true"` to the icon keeps screen readers from announcing "insert drive file icon File available". The per-tile `AvailabilityMarker` already hides its label text behind the chip and uses `aria-label` correctly; the legend is inconsistent with that.

- **[src/utils/documentChildrenSort.js:47-64]** `sortDocumentChildren` accepts a `locale` parameter but `Intl.Collator` silently falls back to the browser default if the locale string is malformed or unsupported. The `locale` comes from Redux state (`state.intl.locale`) which is user-controlled and could in principle hold a value the runtime rejects. Wrapping the `new Intl.Collator(...)` call in a try/catch with a fallback to `'en'` would prevent a silent sort degradation that is hard to reproduce in testing.

### Nitpicks (Optional)

- **[src/pages/DocumentDetails/index.jsx:131-142]** The three sort-order state initialisers (`issuesSortOrder`, `articlesSortOrder`, `otherSortOrder`) are now reset correctly via `key={documentId}` on `<Document>` — great fix. Worth a brief inline comment next to the `key` prop explaining *why* it is there (reset sort state + wantedDeletedState on document navigation), since it looks like an unusual `key` usage from the outside and may be removed as "redundant" during a future refactor.

- **[src/pages/DocumentDetails/DocumentChildrenList.jsx:351-375]** `DocumentTile` is memoised with `React.memo`, which means it bails out when `doc`, `display`, and `labelTier` are all reference-equal. `doc` comes from the store and `labelTier` is shared per grid, so this is effective. Worth double-checking: `display` is computed inside the `useMemo` in `DocumentChildrenTiles` and is a new object on every recalculation of `items`. Since `items` is memoised on `[documents, collectionTitle]`, `display` objects are stable across unrelated parent re-renders — the memoisation is sound. No change needed, just flagging the reasoning is non-obvious.

ClemRz (approved):
This is a well-crafted PR. The two previous review rounds addressed every blocking issue cleanly, and the follow-up work (dead key removal, `hasOwnDescription` / `getIssuesYearRange` tests, `aria-hidden` on the legend icon, `Intl.Collator` try/catch, `key={documentId}` comment) is all in good shape. The code quality, comment density, and test coverage are notably above average.

A few new observations below.

- **[src/utils/documentChildrenSort.js:63]** `Intl.Collator.prototype.compare` is specified as a getter that returns a bound function (ECMA-402 §10.3.4), so the destructuring `({ compare } = new Intl.Collator(...))` is safe and portable — no `this`-binding issue. Worth a one-line comment next to the destructuring, though, since "destructuring a method off an object" reads as a bug to most readers at first glance. Something like: `// .compare is a bound getter per ECMA-402, safe to destructure.`

- **[src/pages/DocumentDetails/DocumentChildrenList.jsx:178–227]** `ChildrenControls` is passed as the `icon` prop of `ScrollableContent` for the articles/issues/other sections (lines ~760, ~778, ~794 of `index.jsx`). `ScrollableContent` wraps that slot in a `div` with an `onClick` that calls `e.stopPropagation()` and expands the card when collapsed. That handler fires before the sort `Select`'s own click, so clicking the sort select on a collapsed card expands it instead of opening the dropdown. The card is `defaultExpanded`, so this only bites the user if they collapse the section and then try to re-sort. Not a crash, but it is confusing. Consider passing `ChildrenControls` as part of a custom `title` node or rendering it inside the `content` instead, or adding `e.stopPropagation()` on the `ControlsRow`'s root element.

- **[src/pages/DocumentDetails/index.jsx:810–870]** `DocumentDetails` now passes `error={error ?? childrenError}` to `Document`. If the detail fetch succeeds but the children fetch fails, `documentData` is set but `error` is truthy — the component renders both the document content **and** the error `ScrollableContent`. That is probably fine as a degraded state, but it is not the same behaviour as before (previously only a detail-fetch error showed the error card). A comment noting this intentional dual-display would prevent a future reader from collapsing the two conditions.

- **[src/pages/DocumentDetails/Section.jsx:34–55]** `TextLink` now returns a React Fragment when `icon` is present. `TextLink` is used inline inside `Typography` (as `DetailItem`'s value), and a Fragment as an inline child of `Typography` is fine — but the `Box` wrapping the icon uses `display: 'inline-flex'` while the sibling `label` is either an `AppLink` or a `Typography component="span"`, both inline. On Firefox the `inline-flex` box and the inline text can produce a 1–2 px vertical misalignment for certain icon sizes. This was raised in the first review (Suggestion 8) and accepted as-is, but worth a quick visual check across browsers now that it is live in several more `DetailItem` slots (Editor, Library, Parent document).

### Nitpicks (Optional)

- **[src/pages/DocumentDetails/index.jsx:124]** `const { locale } = useSelector(state => state.intl)` moved from `DocumentDetails` into `Document`. This is correct — `locale` is needed for sorting inside `Document`. Just noting that the eslint-disable comment on the `useEffect` dependency array at line ~835 (`// eslint-disable-next-line react-hooks/exhaustive-deps`) now suppresses the `locale` missing-dep warning for the `fetchDocumentChildren` call. Since `locale` was removed from that call intentionally (sorting moved to the view), the suppression is still valid — but it now silently covers a wider set of deps. If the effect is ever extended, the disable comment should be re-evaluated.

- **[src/pages/DocumentDetails/DocumentChildrenList.jsx:102–148]** `ChildrenSortSelect` uses `Select` with `variant="standard"` and `disableUnderline`. The `renderValue` returns a Fragment `<><SwapVertIcon />{labels[selected]}</>`. On the initial mount the icon renders even before the user has ever interacted with the select, which is fine. However, the icon is not wrapped in a `Box` or `span` with explicit sizing, so its vertical alignment within the `MuiSelect-select` div depends on the flex layout set in `sx`. This is correct as written — just confirming the `display: 'flex', alignItems: 'center'` in the `sx` handles it. No change needed.

- **[packages/web-app/public/lang/\*.json]** The 15-language additions look clean. Spot-checked `fr.json` ("Paru de {start} à {end}") and `de.json` ("Erschienen von {start} bis {end}") — both use the correct `{start}` and `{end}` placeholders. One minor language note for the French: "Plus récents d'abord" is grammatically fine but the more idiomatic colloquial form would be "Les plus récents d'abord" — leaving this as an editorial note, not a blocker.

urien (approved):
Je te remercie pour ce développement,
Je vais essayer de repérer des collections pour lesquelles le dispositif de tri ne fonctionne pas

PR #1480, opened 2026-07-31
refactor(Typography): normalize heading hierarchy and spacing scale

Paul-AUB:
# refactor(Typography): normalize heading hierarchy and spacing scale

## 🤔 What

- Reduce headings to three consistent roles (page title `h1`, section title `h2`, subsection title `h3`), documented in `AGENTS.md`
- Rework `grottoTheme.js` typography variants (`h1`–`h6`) to a flat, weight-600 scale with fluid sizing only on `h1`
- Update `InfoSection`, `ScrollableContent`, `PageTitle`, `SectionTitle` and the many call sites across `appli`/`common`/`pages` to use theme variants instead of ad hoc font sizes
- Remove the legacy Skeleton CSS 62.5% root font-size hack and the now-unused `GridSystem.jsx` helper, trimming most of `App.css`
- Switch spacing values to the standard MUI 8px measure system (`theme.spacing()` / shorthand `sx` factors) instead of raw pixel values
- Fix an `h6` level regression found while reviewing the hierarchy
- Localisation title has been renamed Access to deduplicate the Localisation section where lives the coordinates.
- Link anchor size slightly reduced

## 🤷‍♂️ Why

Headings and spacing were previously driven by hardcoded font sizes/pixel values and an inherited `htmlFontSize: 10` hack, which silently broke third-party stylesheets (Swagger UI, Leaflet plugins) and made the document outline inconsistent across pages.

## 🔍 How

Introduced a documented 3-level heading model (see the Typography section of `AGENTS.md`) separating visual scale (`variant`) from document outline (`component`), and migrated components to rely on theme variants and `theme.spacing()` factors instead of hardcoded values.

## 🧪 Testing

- `yarn lint`
- Manual visual check of heading hierarchy across cave/entrance/network/organization pages

## 📸 Previews

image

image
image
image


ClemRz (changes requested):
This is a well-scoped and well-motivated refactor. The removal of the Skeleton CSS 62.5% hack and the `htmlFontSize: 10` coupling, the 3-level heading model with `variant`/`component` decoupling, the centralised `MuiBreadcrumbs` override, and the `responsiveFontSizes()` removal are all the right calls. Documentation in `AGENTS.md` is thorough. A few correctness issues are worth addressing before merge.

### Issues (Must Fix)

- **[`src/components/common/Maps/common/Markers/useMarkers.jsx:11`]** The `GlobalStyles` rule sets `.leaflet-container { font-size: 0.625rem }` (10px), which overrides Leaflet's own `0.75rem` default and cascades into every popup and tooltip rendered inside the map. The comment in `App.css` explicitly says the scoped overrides are for `.leaflet-control-attribution` and `.leaflet-control-scale-line` *because* popups and tooltips should not be affected. This `useMarkers.jsx` rule contradicts that decision and will make map popup text (which relies on the Leaflet container font-size for its own `em`-relative sizing) render at 10px.

  Consider either removing this rule entirely (Leaflet's 12px default is now correct at a 16px root) or scoping it to the same two control elements as `App.css`:
  ```suggestion
  export const MarkerGlobalCss = null;
  ```
  Or if there is a reason to keep a container-level override, align it with the intended value and update the `App.css` comment accordingly.

- **[`src/conf/grottoTheme.js:132`]** `h2.fontSize` is `'1.4rem'` but the inline comment reads `// 23`. `1.4 × 16 = 22.4px`, not 23px. If the intended size is 24px (a round step above h3's 20px), the value should be `'1.5rem'`. If 22.4px is intentional, the comment should be `// 22` (or dropped). Either way the comment contradicts the value:
  ```suggestion
  h2: {
    fontSize: '1.5rem', // 24
    fontWeight: 600,
    lineHeight: 1.2
  },
  ```

- **[`src/components/appli/AdvancedSearch/SearchElements.jsx:779`]** The filter-count `<Chip>` label font-size is `0.4375rem` (7px). That is the result of dividing `0.7rem` by 1.6 (the old 62.5% ratio), but `0.7rem` at a 16px root is already 11.2px — a reasonable small label. The division was unnecessary and produces unreadably tiny text:
  ```suggestion
  sx={{ height: 20, fontSize: '0.7rem' }}
  ```

- **[`src/components/common/Layouts/PageTabs.jsx:111`]** The tab badge font-size is set to `0.625rem` (10px). The original `1rem` at the old root was 10px, so this is a faithful conversion. However, `1rem` at a standard 16px root would be 16px — already too large for a small badge counter. Consider a value like `0.75rem` (12px) which is legible without being oversized, rather than a direct numerical copy of the old pixel value.

- **[`src/components/appli/AdvancedSearch/SearchElements.jsx:122,291`]** The `minWidth`/`maxWidth` values `9.375rem` and `18.75rem` are exact 62.5%-root translations of `15rem` and `30rem`. These are layout widths, not font sizes — they were `15rem = 150px` and `30rem = 300px` before (at 10px root). At the 16px root, `9.375rem = 150px` and `18.75rem = 300px`, so the pixel values are preserved, which is correct. However, the fractional `rem` values are harder to reason about than their pixel equivalents. Consider whether `150px`/`300px` as literal values or `theme.spacing(18.75)`/`theme.spacing(37.5)` would be clearer to future maintainers.

- **[`src/components/appli/EntitiesForm/utils/FormContainers.jsx:64`]** `FormSection` now uses `component="h2"`, and the comment says it "sits under the page title". This is correct for entity creation pages that have a single form. However, some forms (e.g. `EntranceDetail + CaveDetail + EntranceAttributes`) stack multiple `FormSection`s inside a single form that is itself already inside a section. If the containing page does not render an `h2` above the form, the first `FormSection` heading is the only `h2` and the rest are siblings, which is correct. But if a section-level `<ScrollableContent>` wraps the form (which renders its card title as `h2`), the `FormSection` headings would generate a second `h2` tier. Consider documenting the assumption or aligning the component to `"h3"` with a note that callers on form pages without an enclosing `h2` should override it.

- **[`src/pages/Messages/index.jsx:288`]** The `Conversations` heading uses `variant="h3" component="h1"`. The messages page appears to be a full page (inside `PageContainer`) with no `PageTitle` rendering an `h1` above it. Using `variant="h3"` (20px) for the only `h1` on the page gives it the same visual weight as a subsection title. Consider `variant="h2"` here to align with the page-level visual scale, or confirm the `h1` is intentionally understated.

- **[`src/components/appli/Entry/Locations/index.jsx:46` and `src/components/appli/Entry/Snapshots/index.jsx:132`]** The section title `'Location'` was renamed to `'Access'`. This is a semantic improvement (matching the PR description), but it is a visible user-facing string. The existing translation key `'Location'` (used in `Properties.jsx:86`) and the new `'Access'` key (used in `Locations/index.jsx`) now coexist for what was previously the same label. Confirm the translation files have been updated for all 15 languages, and that `'Access'` isn't already used elsewhere with a different meaning.

- **[`src/components/appli/DataHomepage/index.jsx:15`]** The `SectionTitle` styled component uses hardcoded `font-size: 35px` with `font-weight: 300` and `letter-spacing: -1px`. These values restore the Skeleton-CSS styling that was globally applied before. This component is in the `helpers/` folder (deprecated by convention) and uses a fully custom scale that bypasses the theme. Consider migrating it to a theme variant or at least noting it as a known deviation.

### Nitpicks (Optional)

- **[`src/conf/grottoTheme.js`]** The `h2` comment ladder reads `// 23` — see the Issue above, but even if intentional, `22.4` or `22` would be more accurate than `23`.
- **[`src/components/appli/EntitiesForm/Riggings/AnchorToolbar.jsx:72`]** `fontSize: '0.8125rem'` (13px) is a direct conversion of `1.3rem` at the old root. That is a reasonable text size, but `body2` is already `0.8125rem` — using `fontSize: 'inherit'` or the `body2` variant would tie this value to the theme scale.
- **[`src/helpers/Entity.jsx:27`]** `EntityId` is now `0.625rem` (10px). That is quite small — 10px is below common accessibility minimums for body text (typically 12–13px). Consider `0.75rem` (12px).

urien (approved):
Pour moi la mise en forme est visuellement cohérente sur l'ensemble des pages

ClemRz (changes requested):
All previous review issues are addressed. The `MarkerGlobalCss` export is gone entirely (correct call — Leaflet's own default is right at 16px), the chip font-size is fixed to `0.7rem`, the `h2` comment is now `// 22`, and the `FormContainers`/`Messages`/`Access` items are well-documented or intentional. The bonus `LegendHeader` refactor is a clean improvement. Below are the remaining items I noticed on this pass.

### Issues (Must Fix)

- **[`src/conf/grottoTheme.js:131`]** The inline comment on `h1` still reads `// The floor stays above h2 (24)`. After the `// 23` → `// 22` correction, h2 sits at 22.4px, not 24px. The invariant is still true (26px > 22.4px) but the parenthetical `(24)` is now stale and will mislead the next person who checks the scale. The same number appears in the `AGENTS.md` typography section under "fluid sizing" (`h1`'s floor is 26px precisely because `h2` sits at 24px`). Both spots should read `22` (or `22.4`) to match the `// 22` comment on the `h2` entry itself.
  ```suggestion
      // 26 at 360px, 32 from 1200px up. The floor stays above h2 (22).
      fontSize: 'clamp(1.525rem, 1.26rem + 0.72vw, 2rem)',
  ```

- **[`src/components/common/card/RandomEntryCard.jsx:51`]** `WhiteRating` now has `fontSize: '0.625rem'` (10px). MUI's `Rating` uses `fontSize` to control the star icon size via `em`, so the five stars will render at roughly 10px wide — too small to convey the rating meaningfully, and the `readOnly` stars are the only visual quality indicator on this card. The original value at the old 10px root was `1rem = 10px`, so this is again a direct numerical copy of the old pixel value rather than a deliberate choice. Consider `0.875rem` (14px) or the `caption` variant size (`0.75rem`, 12px) for a legible but compact display:
  ```suggestion
  const WhiteRating = styled(MuiRating)({
    color: 'white',
    fontSize: '0.875rem'
  });
  ```

- **[`src/pages/Messages/index.jsx:239`]** The conversation-list timestamp uses `fontSize: { xs: '0.625rem', sm: '0.6875rem' }` (10px / 11px). These are below the 12px floor that most accessibility guidelines cite for secondary text; 10px is the same size as the `EntityId` that was flagged in the previous review. A `caption` token (`0.75rem`, 12px) would stay compact while remaining legible:
  ```suggestion
                      sx={{
                        fontSize: { xs: '0.6875rem', sm: '0.75rem' },
                        opacity: 0.8
                      }}>
  ```

### Nitpicks (Optional)

- **[`AGENTS.md` — Typography section]** The fluid-sizing paragraph still contains `h1`'s floor is 26px precisely because `h2` sits at 24px`. Align with the `// 22` correction to `22` or `22.4` for internal consistency.
- **[`src/App.css:88`]** `.leaflet-control-attribution` is pinned to `0.46875rem` (7.5px) to preserve the pre-migration visual size. 7.5px is below the range where text can be considered legible, and attribution text carries a licence-compliance purpose. Worth noting that Leaflet's own default for this element is around 11px — if the visual goal is "tiny but readable", `0.625rem` (10px) would still be noticeably smaller than the rest of the UI while staying in a legible range.
- **[`src/pages/Messages/ConversationDetail.jsx:752`]** The character-counter box uses `fontSize: '0.625rem'` (10px). This only appears when the user has typed 4000+ characters, so it is a warning utility rather than content — 10px is defensible here, but worth a note that this matches the floor of the now-explicitly-discouraged sub-12px range.

ClemRz (changes requested):
All the must-fix items from the previous two reviews are resolved — the `MarkerGlobalCss` export is fully gone, the chip font-size is corrected, `h2` sits at 23px, and the `FormContainers`/`Messages`/`Access` items are either fixed or deliberately kept with good documentation. The `LegendHeader` bonus refactor is a clean improvement. Below are the remaining items I noticed on this pass.

### Issues (Must Fix)

- **[`src/components/common/SideMenu/index.jsx:65`]** `Content` has `font-size: 0.59375rem` (9.5px). That is the straight 62.5% conversion of `0.95rem`, but the content wrapper is not a text node — its font-size flows into every `ListItemText`, menu label, and button inside the drawer. 9.5px for body-level navigation text is below the legibility floor for any viewport. The original `0.95rem` at a 16px root is `15.2px`, which is fine; this looks like an unintended division. Consider removing the `font-size` override entirely and letting the theme's `body1` (15px/16px) apply, or using `0.9375rem` (15px) if a slight size reduction was intentional:
  ```suggestion
    font-size: 0.9375rem;
  ```

- **[`AGENTS.md:457`]** The example `clamp()` formula in the "Fluid sizing" section reads `clamp(1.625rem, 1.46rem + 0.72vw, 2rem)` (floor 26px), but the actual value in `grottoTheme.js` is `clamp(1.525rem, 1.26rem + 0.72vw, 2rem)` (floor ~24.4px). A reader following this documentation to reproduce the h1 rule will use a different formula and get a different result. Align the example with the actual value.

- **[`src/App.css:88`]** `.leaflet-control-attribution` and `.leaflet-control-scale-line` are pinned to `0.46875rem` (7.5px). The PR description acknowledges this was a deliberate won't-fix in the last update. Just leaving a note for the record: Leaflet's own default for the attribution control is ~11px, and the OpenStreetMap attribution carries a licence-compliance purpose. If the intent is to keep it visually suppressed, `0.5rem` (8px) is the common minimum that screen readers can still surface; `0.46875rem` may still be fine but is at the outer edge of what any browser will render legibly at default zoom.

- **[`src/components/common/SideMenu/MenuLinks.jsx:27`]** `SectionHeader` is now `0.71875rem` (11.5px). That is compact but reasonable for an ALL-CAPS nav label. The `lineHeight: '1.25rem'` (20px) alongside that font-size creates a 1.74 line-height, which is generous for a single-line label. Not a bug, but worth noting if visual density is a concern.

- **[`AGENTS.md` — Typography table`]** The "Subsection title" row maps to `h5` as both the visual scale and the component. The actual pattern used throughout the PR is `variant="h5" component="h3"` (or `component="h2"` in untitled cards). The table would be more accurate if the "Visual scale" column read `h5` and the "Component" column read `h3` (default) — matching the prose just below it.

### Nitpicks (Optional)

- **[`AGENTS.md:426`]** The h1 comment in `grottoTheme.js` now correctly reads `// The floor stays above h2 (23)`, but the prose in `AGENTS.md` still reads `h1's floor is 26px precisely because h2 sits at 23px`. The invariant is stated correctly (`h1`'s floor > `h2`'s size), but the 26px figure is stale — the actual clamp floor is `1.525rem × 16 = 24.4px`. A small correction would keep the documentation consistent with the code.
- **[`src/helpers/Entity.jsx:27`]** `EntityId` remains at `0.625rem` (10px), as documented. The previous review noted this; flagging it again only as a reminder if accessibility feedback arrives — the deliberate choice is well-reasoned for its use case (de-emphasised record ID in search results).

ClemRz (commented):
All must-fix items from the previous three review passes are resolved — the `SideMenu/index.jsx` font-size override is gone (theme `body1` now applies cleanly), `MenuLinks.jsx` `lineHeight` is unitless `1.5`, `RandomEntryCard` stars are at `0.875rem`, the Messages timestamp is `0.6875rem/0.75rem`, the character counter is `0.75rem`, and the `AGENTS.md` clamp example now correctly cites `1.525rem` and 24.4px. The `LegendHeader` bonus refactor is clean. One small documentation inconsistency remains.

### Issues (Must Fix)

- **[`src/conf/grottoTheme.js:126`]** The inline comment on `h1` reads `// 26 at 360px, 32 from 1200px up`. The minimum value of `clamp(1.525rem, 1.26rem + 0.72vw, 2rem)` at a 16px root is `1.525 × 16 = 24.4px`, not 26px (that figure was the previous value before the h2 bump iteration). `AGENTS.md` already correctly states 24.4px, so the two are now inconsistent:
  ```suggestion
      // 24.4 at 360px, 32 from 1200px up. The floor stays above h2 (23).
      fontSize: 'clamp(1.525rem, 1.26rem + 0.72vw, 2rem)',
  ```

### Nitpicks (Optional)

- **[`src/components/appli/ImportCSV/ImportContainer.jsx`]** `makeStyles` is a JSS API (MUI v4 legacy). The rest of the codebase uses `styled()` or `sx`. Not introduced by this PR, but worth noting for a follow-up migration.
- **[`src/pages/Messages/index.jsx:239`]** The conversation-list timestamp is `xs: '0.6875rem'` (11px). Still under the 12px floor, though the improvement from 10px is meaningful. A follow-up to `0.75rem` across both breakpoints would align with the `caption` token, but this is a pre-existing rough edge, not a regression introduced here.
- **[`src/helpers/Entity.jsx:27`]** `EntityId` remains at `0.625rem` (10px). Noted again only as a reminder — the deliberate de-emphasis is documented and reasonable for its use case.

ClemRz (approved):
The last must-fix from the previous pass — the stale `// 26` comment on `h1` in `grottoTheme.js` — is now resolved: the comment correctly reads `// 24.4 at 360px, 32 from 1200px up. The floor stays above h2 (23)`, and `AGENTS.md` is consistent with it. All previous issues are addressed.

- **[`src/components/common/Layouts/PageTabs.jsx:219`]** The mobile bottom-nav label uses `'0.5625rem !important'` (9px). This is the same pixel size as before (the old `0.9rem` at the 10px root), so it is not a regression introduced here, but the conversion makes the value explicit for the first time: 9px is below the floor that browsers render legibly at default zoom, and `!important` prevents any responsive override. Worth revisiting in a follow-up once the rest of the refactor settles — `0.625rem` (10px) would be a safer floor without changing the visual character of the compact bar.

- **[`src/components/appli/ImportCSV/ImportContainer.jsx:7`]** `makeStyles` is the JSS API (MUI v4 legacy). The rest of the codebase has moved to `styled()` / `sx`. This component is not part of the refactor's scope, but the single-class usage (`stepper: { margin: '...' }`) is easy to migrate; a follow-up `tech/` PR would be a clean win.

### Nitpicks (Optional)

- **[`src/helpers/Entity.jsx:27`]** `EntityId` remains at `0.625rem` (10px). Flagged again only as a reminder in case accessibility feedback arrives — the deliberate de-emphasis is documented and the tradeoff is clear.
- **[`src/components/appli/DataHomepage/index.jsx`]** `SectionTitle` is in the deprecated `components/appli/DataHomepage/` folder (not `helpers/` as noted in an earlier pass). The deviation from the theme scale is now clearly documented in a comment — good. A follow-up migration to move or reclassify this component would complete the cleanup.

PR #1481, opened 2026-07-31
feat(Map): unified location control + waypoint navigation

Paul-AUB:
# feat(Map): unified location control + waypoint navigation

## 🤔 What

Two things, tightly coupled because they share the same sensors:

**1. A single "location" control replacing `LocateControl` + `CompassControl`**

- Drops the `leaflet.locatecontrol` dependency (and its CSS override, its SVG asset,
  and the `GlobalStyles` CSS-variable bridge in `App.jsx` that only existed to feed it).
- One button driving a 4-state machine: `off → follow → compass`, with `located`
  as the detached state after the user acts on the map.
- Blue location dot with an accuracy circle and a direction cone.
- A floating round north badge (top-right) appears whenever the map ends up rotated.
- All map sensors (geolocation watch, orientation, wake lock) are now owned by a single
  map-scoped provider, `MapLocationProvider`.

**2. A navigation waypoint** — long-press the map to drop a target, then walk to it
following a live distance / bearing / arrow HUD. Available on the global map and on the
fullscreen entrance map, sharing one persisted target.

**3. Geolocation is no longer requested unless it will actually be used** (details below).

New i18n keys (6) added and translated across all 16 lang files.

## 🤷‍♂️ Why

The map's field-navigation story was built from three independent pieces that each owned
their own sensor lifecycle:

- `leaflet.locatecontrol` started its own `watchPosition`, `CompassControl` its own
  orientation subscription, and `pages/Map.jsx` fired a third `getCurrentPosition` on
  every visit — **including when the position was going to be thrown away**, because a URL
  target or a saved position took priority. The permission prompt appeared before the user
  had asked for anything.
- Two buttons for one intent ("where am I / which way am I facing") meant an unstable
  bottom-right stack and duplicated state that could disagree.
- Nothing helped a caver actually *reach* an entrance: you could see a pin, but not how far
  it was, in which direction, or whether you were walking the right way.

Battery and GPS behaviour matter here more than usual: the primary scenario is a phone,
outdoors, walking to a cave entrance, often with a weak fix under canopy.

### Shared sensor layer — `MapLocationContext.jsx`

`MapLocationProvider` is mounted once inside `CustomMapContainer`, so **every map** gets
exactly one geolocation watch and one device-orientation subscription, whatever the number
of consumers (control, dot, waypoint HUD).

Two ref-counted, declarative hooks let features express a *need* without owning a lifecycle:

```js
useRequestUserLocation(mode !== MODE.OFF); // keeps the watch alive while true
useRequestHeading(shown);                  // same for the compass heading
```

The ownership rule is deliberate: **the location control is the only caller of
`useRequestUserLocation`**. Features that merely *use* the position (the waypoint) read it
without requesting it. That is what keeps the button honest — `active` in the context and
the control's own mode can never disagree, and the permission prompt stays tied to a real
user gesture.

### Geolocation optimizations

| # | Optimization | Detail |
|---|---|---|
| 1 | **No prompt without intent** | `useGeolocation({ enabled })` keeps the hook fully dormant — no watch, no permission prompt — until a consumer flips it on. |
| 2 | **`pages/Map.jsx` fallback made conditional** | Geolocation is only enabled when there is *neither* a URL target *nor* a saved position to restore. Previously the prompt fired on every map visit and the answer was discarded. |
| 3 | **Waypoint restore never wakes the GPS** | A waypoint rehydrated from `localStorage` is a leftover, not a statement of intent. It used to turn the GPS on (prompt included) while the control's button still read "off". |
| 4 | **One watch, one sensor** | Three independent subscriptions → one shared, ref-counted pair. |
| 5 | **Accuracy tiered by use case** | `enableHighAccuracy` is **off by default** — a coarse network fix is fine and much cheaper for "center the map on me". Only the field-navigation provider opts into GPS. |
| 6 | **`timeout` / `maximumAge` tuned per mode** | A one-shot gets `10s / 10s`. A watch gets *no* timeout and `maximumAge: 0`: the watch contract is "tell me when you know", and the timeout applies to **every** acquisition — a finite one manufactured `TIMEOUT` errors out of an ordinary 30–60s cold fix under canopy and left tracking looking broken while it was simply still searching. |
| 7 | **`NO_TIMEOUT = 0xFFFFFFFF`, not `Infinity`** | `timeout` is a WebIDL `unsigned long`; an engine that doesn't apply the spec's `[Clamp]` coerces `Infinity` to `0` — a timeout firing on *every* acquisition. |
| 8 | **Transient errors are not failures** | Once a fix is on screen, `TIMEOUT` / `POSITION_UNAVAILABLE` mean "no fresher fix right now", not "tracking failed". Only a denied permission (code 1) is terminal. Before this, every GPS blink in the field turned the control red and re-toasted. |
| 9 | **Re-subscribe on return to foreground** | Screen off / tab backgrounded: browsers suspend an active watch and may drop it for good, yet the `watchId` stays valid so nothing ever re-arms it — tracking looks frozen forever. A `visibilitychange` listener bumps a `resumeTick` that re-subscribes. |
| 10 | **Screen wake lock while navigating** | `useWakeLock` is tied to the *request count*, not to the watch: the screen locking is both a usability problem and the main reason the watch stops delivering. Best-effort (missing API, refused on low battery) and re-acquired on every return to the foreground, since the browser releases it whenever the page hides. |
| 11 | **30s keep-warm grace period** | Losing the last consumer doesn't kill a *working* watch straight away: re-acquiring a GPS fix from cold costs tens of seconds, and toggling the control off/on is an ordinary gesture. A watch that never delivered is stopped at once — that's what makes the next tap a genuine retry (cleared error, prompt re-issued). |
| 12 | **Orientation sensor throttled** | Hard 120 ms rate cap **plus** a 2° minimum angular delta. The sensor fires at ~60 Hz and each emitted value triggers a full map-rotation redraw. |
| 13 | **Stable hook identity** | `useDeviceOrientation` returns a memoized object. It gets spread into the context value, so an unstable reference would fan out a re-render to every `useDeviceHeading()` consumer on every position tick. |
| 14 | **Bearing updates suspended during zoom animations** | Re-applied once on `zoomend` instead of thrashing throughout. |
| 15 | **Heavy layers can opt out while rotating** | The control fires `compassfollowchange` so heatmaps can hide during continuous rotation. |
| 16 | **`Intl.NumberFormat` cached per locale** | `formatDistance` is called on every position update by the HUD; the three formatters used to be rebuilt each time. Also de-duplicates the copy that lived in `MeasureControl`. |
| 17 | **GPS course as heading fallback** | When the device has no magnetometer, `coords.heading` (only defined while moving) drives the direction cone. |

### Location control state machine

| Mode | Behaviour | Next tap |
|---|---|---|
| `off` | Not tracking, no dot, no sensors running | → `follow` (+ `startCompass()` inside the gesture, the only place iOS grants orientation permission) |
| `follow` | Recenters on every fix, north-up | → `compass` (or a plain recenter if no usable compass) |
| `compass` | Heading-up: rotates the map to the heading, user offset to the lower third | → `follow` |
| `located` | Dot shown, map free, bearing frozen | → `follow` |

Detaching to `located` is what makes ordinary map interaction work at all — while following,
every fix recenters, so any view change the user asks for would be undone within a second.
Three signals detach: `dragstart` (manual pan), `popupopen` (every entity tap goes through a
popup), and a new custom **`followdetach`** event for affordances that move the view *without*
a popup — cluster dives (`ClusterLayer`), geocoding jumps (`GeocodingControl`), and the
waypoint's off-screen badge.

Ordering detail worth flagging: the "straighten to north" effect is declared **before** the
recenter effect on purpose. `setBearing` pivots around the container centre, and in compass
mode the user sits well below it, so straightening swings them away; recentering afterwards
computes the target under the final bearing. The reverse order left the map visibly off until
the next fix.

Angles that drive an animated CSS `rotate()` (the needle, the HUD arrow) all go through the new
`useContinuousAngle` hook, which accumulates the shortest signed delta. CSS interpolates
`rotate()` numerically, so a wrapped `[0, 360)` series sweeps the long way round every time it
crosses its boundary — `359° → 2°` animates *down* through 180°.

### Waypoint navigation

**Placing it** — long-press (`contextmenu`) on a touch device:

- **Global map**: a new entry in the existing coordinate context menu (`MapClusters/index.jsx`).
- **Fullscreen entrance map**: `WaypointControl` brings its own minimal long-press menu.

Exactly one of the two mounts per map, otherwise both would answer the same `contextmenu` event.
The entry reads "Place a waypoint here" or "Move waypoint here" — there is only ever one target.

**Where it lives** — `useWaypoint()` wraps `useLocalStorage` on a single key,
`grottocenter_waypoint`, shared by every map offering field navigation. Placing a waypoint from
an entrance page and finding it on the global map is the *same* target, not two. It is persisted
so it survives a reload in the field.

**What it renders** (`WaypointNavigation`, mounted only while a waypoint exists):

- **The pin** — a red teardrop `divIcon` (no asset-pipeline work), anchored on its bottom tip so
  the point sits exactly on the target, `zIndexOffset: 2000` to sit above the explored-cave
  badges. It stays screen-upright on a rotated map; only the guidance line rotates with the view.
- **The guidance line** — a straight polyline from the user to the pin, drawn only when a live
  position is available.
- **The HUD** — a bottom card with the straight-line distance (metric + imperial, locale-formatted)
  and the absolute bearing in degrees plus its 16-point cardinal (`142° SE`). When a compass
  heading is available, a large arrow shows the direction **relative to where the user is
  currently facing** (0 = straight ahead) — "turn this way" rather than "it's south-east". The HUD
  requests a heading itself, so the arrow works whether or not the user also turned tracking on.
- **The off-screen indicator** — when the pin's *whole bounding box* leaves the viewport, a round
  red arrow badge is pinned to the map border in its direction, computed by scaling the
  centre→waypoint vector onto an inset border rectangle. It is bearing-aware
  (`latLngToContainerPoint` already accounts for `leaflet-rotate`). Tapping it fires `followdetach`
  and then pans back to the waypoint — without the detach, the next fix would snap the view
  straight back to the user.

**Removing it** — long-press the pin → "Remove waypoint". The marker handler calls
`L.DomEvent.stopPropagation` on the *Leaflet* event so `originalEvent._stopped` is flagged and the
map's own coordinate menu doesn't open on top of ours.

**Graceful degradation without a position**: the pin, its delete menu and the off-screen indicator
stand alone; the line and the HUD appear as soon as the user turns the location control on.

**Scope** — touch-only and (on the entrance map) fullscreen-only. Long-press *is* the gesture, and
an embedded map a few hundred pixels tall is not where anyone walks to a cave.

### Fullscreen overlays

`leaflet.fullscreen` makes the **map container itself** the fullscreen element, and a browser paints
nothing outside it. MUI portals `Menu` / `Tooltip` to `document.body` by default — outside it — so
the long-press menus and control tooltips existed in the DOM but were never painted in fullscreen.

- `useMapOverlayContainer()` returns the map container while fullscreen, `undefined` otherwise
  (keeping MUI's default body portal, since `.leaflet-container` is `overflow: hidden` and would
  clip overlays on a small embedded map).
- `useIsFullscreen()` is **seeded from the map's current state**, not from future events alone: the
  fullscreen-only features mount *because* the map went fullscreen, i.e. after `enterFullscreen`
  already fired. An event-only hook reported `false` for their entire lifetime.

### API surface change on `CustomMapContainer`

```diff
- isLocateControl / isCompassControl
- isLocateControlInFullscreen / isCompassControlInFullscreen
+ isLocationControl
+ isLocationControlInFullscreen
```

`rotate` is now enabled by either flag, since the single control owns map rotation.

## 🔗 Related API PR

None — frontend only, no API change.

## 🧪 Testing

**Automated** (added in this PR):

- Unit (Vitest): `useGeolocation`, `useWakeLock`, `useContinuousAngle`, `useIsFullscreen`,
  `useMapOverlayContainer`, `utils/geo`, `LocationControl`, `UserLocationMarker`.
- E2E (Cypress): `cypress/e2e/map-location-control.cy.js` — stubs `navigator.geolocation` and
  `deviceorientationabsolute` before boot, then walks the mode machine (dot + accuracy circle on
  activation, heading-up rotation, north reset).

**Manual — best on a real phone** (a desktop exposes `DeviceOrientationEvent` but never fires it;
Chrome DevTools device emulation + the Sensors panel is the fallback):

1. **No premature prompt** — open `/map` with a `?target=` param or a previously saved position:
   no geolocation prompt should appear. Open it clean: the prompt appears once.
2. **Mode machine** — tap the control: prompt → dot + accuracy circle, map follows. Tap again:
   map rotates heading-up, the user slides to the lower third, the north badge appears. Tap again:
   back to north-up follow, badge gone.
3. **Detach** — while following, drag the map / tap an entrance / dive into a cluster / pick a
   geocoding result: tracking detaches to `located`, the dot stays, the view stops snapping back.
4. **Errors** — deny the permission: a toast appears once and tracking stops. Walk under cover so
   the fix goes stale: **no** red button, **no** toast storm; the last position stays.
5. **Backgrounding** — lock the screen for ~30s and come back: tracking resumes on its own.
6. **Wake lock** — with tracking on, the screen should not dim/lock on its own.
7. **Waypoint** — long-press the map → "Place a waypoint here". Check the pin, the line to the dot,
   the distance/bearing card, and that the arrow rotates as you turn on the spot. Pan the pin off
   screen: the red badge sticks to the border in its direction; tap it to come back. Long-press the
   pin → "Remove waypoint".
8. **Shared target** — place a waypoint on the global map, open a fullscreen entrance map: the same
   waypoint is there. Reload: still there, and **without** a new geolocation prompt.

## 📸 Previews

_TODO: add screenshots / a short screen recording of compass mode and the waypoint HUD._


ClemRz (changes requested):
This is a thorough, well-thought-out PR. The sensor architecture (ref-counted lifecycle, shared `MapLocationProvider`, `useGeolocationPermission` gating all auto-starts) is the right design and clearly documented. The follow-up geolocation fixes addressing the repeated-permission-prompt bug from field testing are exactly what this kind of PR needs before merge. A few things to address below.

### Issues (Must Fix)

- **[`LocationControl.jsx:342`]** `showNorthButton` misses the `MODE.FOLLOW` case after the user exits compass mode via a tap (the `COMPASS → FOLLOW` transition in `handleClick`). In that path `resetNorth()` is called (bearing resets to 0) so the button correctly disappears — but if the user was **dragged to `located`** while in `follow` and the map was previously rotated by a prior compass session, `needleBearing` can be non-zero while `mode === MODE.FOLLOW`. The north button never appears. The condition should include all modes where the map can be left rotated and no active rotation is happening:
  ```suggestion
  const showNorthButton =
    mode === MODE.COMPASS ||
    ((mode === MODE.LOCATED || mode === MODE.FOLLOW) && normalizedBearing !== 0);
  ```
  (For `MODE.FOLLOW`, `resetNorth` is called on entry *from COMPASS*, but a `follow → located` drag preserves the bearing, and re-entering `follow` from `located` also doesn't reset. The north button should be visible whenever the map is tilted, whatever the current mode.)

- **[`MapLocationContext.jsx:103–108`]** The `userLocation` memo spreads `...geo` but then **also** includes `active: requestCount > 0`. `requestCount` is not stable between renders — if it changes for reasons unrelated to the current consumers, it fans out a re-render to every `useUserLocation()` subscriber even though `geo` itself didn't change. Consider whether `active` should be in the context value at all (it drives the `shown` flag in `UserLocationMarker`, but the marker could derive that from `hasLocation` instead), or derive it from `isWatching` which is already stable state.

- **[`useGeolocation.js:166–173`]** The "fresh session" effect resets `fix` and `error` whenever **either** `enabled` OR `options` changes. The `options` object is memoised on `[watch, enableHighAccuracy]`, so this fires on every prop change of `enableHighAccuracy` — including accidental parent re-renders that compute the same value. A position currently being displayed would be blanked for a frame. This is a latent bug that can manifest once `enableHighAccuracy` is passed from a component that re-renders often. Consider changing the deps to `[enabled]` only, and handling the `options` change via the subscription effect that already depends on it.

- **[`WaypointControl.jsx:49–53`]** `useMapEvent('contextmenu', ...)` registers a new inline arrow function on every render, which means the previous handler is never removed (the callback reference changes). This is a `react-leaflet` gotcha: pass a stable reference (e.g., `useCallback`) or Leaflet accumulates listeners. This can result in the same context-menu event spawning multiple menu states.
  ```suggestion
  const handleContextMenu = useCallback(e => {
    if (!isTouch || !isFullscreen) return;
    e.originalEvent.preventDefault();
    setMenu({
      coords: { lat: e.latlng.lat, lng: e.latlng.lng },
      anchor: { top: e.originalEvent.clientY, left: e.originalEvent.clientX }
    });
  }, [isTouch, isFullscreen]);

  useMapEvent('contextmenu', handleContextMenu);
  ```
  The same issue is present in `MapClusters/index.jsx`'s `contextmenu` handler, though that one pre-dates this PR.

- **[`useWaypoint.js` / `MapClusters/index.jsx:436–441`]** The persisted waypoint is restored from `localStorage` without any runtime validation. A `{ lat, lng }` shape from a previous session could be corrupted (e.g. `{ lat: NaN, lng: NaN }` from a partial write), which would be passed directly into Leaflet and trigger an exception. A minimal guard in `useWaypoint` or at the mount site would prevent this:
  ```suggestion
  // In useWaypoint.js, add a validator option to useLocalStorage:
  const isValidWaypoint = v =>
    v !== null &&
    typeof v === 'object' &&
    Number.isFinite(v.lat) &&
    Number.isFinite(v.lng);

  const useWaypoint = () => {
    const [raw, setWaypoint] = useLocalStorage(WAYPOINT_STORAGE_KEY, null);
    const waypoint = isValidWaypoint(raw) ? raw : null;
    return [waypoint, setWaypoint];
  };
  ```

- **[`LocationControl.jsx:237–248`]** The `GEO_ERROR_MESSAGES` object values are used directly as i18n keys (`formatMessage({ id: GEO_ERROR_MESSAGES[geoError] })`). This works today because those strings happen to exist as keys in `en.json`, but the pattern is fragile: the full English sentence is the key, a typo in either place silently falls back to the raw string, and the tooltip/aria-label also uses the raw string as the `tooltipId` (line 397). Consider using short semantic keys (e.g. `'location.error.denied'`, `'location.error.unavailable'`, `'location.error.timeout'`) like the rest of the app.

- **[`WaypointHud.jsx:55`]** `useRequestHeading(true)` is called unconditionally, meaning the HUD always holds the compass sensor active for as long as it is mounted — even when `heading` is null (no magnetometer). The PR description notes this is intentional ("the arrow works whether or not the user also turned tracking on"), but it means mounting the HUD on a device without a compass will start and immediately error-out the orientation sensor, then keep retrying. Consider passing the condition `compassSupported && compassError !== 'unavailable'` here, or at least silencing the no-data timeout on the second+ retry.

- **[`LocationControl.jsx:302`]** The open-cycle discussion raised in the PR description (`compass → off` path is missing) is a UX issue that will surface now that auto-start can put the user into `LOCATED` without a tap. The existing `handleClick` exits `COMPASS` back to `FOLLOW`, but there is no way to dismiss tracking back to `OFF` without an error. This should be decided before merge rather than left for a follow-up — at minimum, document the decision as a code comment.

- **[`useGeolocation.js:282`]** The watchdog's `getCurrentPosition` poke uses `STALE_AFTER_MS` (15000 ms) as its timeout. That means every 5 s, once the watch goes stale, a one-shot fires and can take up to 15 s to time out — three overlapping pokes could be in-flight simultaneously. Consider using the same `10000` ms as the standard one-shot, or explicitly cancelling the previous poke before firing the next.

- **[`UserLocationMarker.jsx:57–70`]** The marker is created in one `useEffect` and updated in two separate effects. The `shown`/`location` effect runs after the `facing` effect, but if `shown` goes `true` in the same commit as `facing` changes, the cone-update effect runs first (when `el` is still null because the marker isn't on the map yet) and silently skips. This is unlikely to matter in practice since `facing` only updates on sensor events, but structuring both updates inside one effect (keyed on `[map, shown, location, facing]`) would be cleaner and remove the dependency on React's ordering guarantee.

- **[`WaypointOffscreenIndicator.jsx:31–43`]** The off-screen indicator runs `recompute` on `move`, `zoom`, `rotate`, and `resize`. On a rotating map in compass mode, this fires on every heading update (which throttles at 120 ms but still drives many recomputes). The computation involves `latLngToContainerPoint` which is cheap, but `useState` setter on every tick adds a render. Consider comparing the computed `indicator` to the previous value before calling `setIndicator`, or using `useReducer` with an equality check.

- **[`WaypointHud.jsx:64`]** `relativeBearing` wraps to `[0, 360)` by design, but `useContinuousAngle` is initialised to `null` and passes the first angle through unchanged. When `heading` changes from `null` to a real value for the first time, the arrow jumps from wherever the accumulator left off (null → first angle). This is correct per the hook's contract, but worth a note in the component comment.

- **[`MapContainer.jsx`]** The new `isLocationControl` / `isLocationControlInFullscreen` API surface works, but the prop names are slightly inconsistent with their usage: `isLocationControl` renders the controls unconditionally (always visible), while `isLocationControlInFullscreen` only renders them in fullscreen — the names don't surface that distinction. Consider `isLocationControlAlways` and `isLocationControlInFullscreen`, or document the difference in the prop comments.

### Nitpicks (Optional)

- **[`useGeolocation.js:L1`]** The module imports `useCallback` but `onPosition` and `onError` only close over refs — they could be plain functions declared with `function` outside the hook, removing the `useCallback` overhead entirely. Minor point given how often this hook is called.

- **[`WaypointLayer.jsx:36–44`]** `L.DomEvent.stopPropagation(e)` is called on the Leaflet event to stop the map's `contextmenu` from opening on top of the waypoint menu. The comment is accurate, but it's worth noting that `e.originalEvent.stopPropagation()` would *not* achieve the same thing — Leaflet's DomEvent approach is the correct one here. The comment could make that explicit.

- **[`waypointIcon.js:L1`]** The `html` variable is a template literal defined at module level (static), which is fine for a single color. If the color token ever needs to follow the MUI theme, this will need to be moved into a factory function. Not a problem now, just flagging the coupling.

- **[`cypress/e2e/map-location-control.cy.js:91–106`]** The compass-mode test fires `__setHeading(90)` twice but doesn't assert on the rotation (Leaflet's bearing or any DOM state). The test checks that the map container exists, which it always does. Adding an assertion on the north-reset button appearing (since that's the observable side-effect of compass mode being active) would make the test load-bearing.

ClemRz (commented):
This is a very well-crafted PR. The sensor architecture (shared `MapLocationProvider`, ref-counted lifecycle, permission-gated auto-start) is exactly the right design for this feature set, and the field-testing follow-up that addressed the repeated-prompt bug was handled cleanly. All items from the prior review are confirmed addressed. A few new observations below.

### Issues (Must Fix)

- **[`LocationControl.jsx:405`]** `tooltipId` defaults to the raw string `'Use my location'`, but in the error branch it falls back to `GEO_ERROR_MESSAGES[2]` (`'location.error.unavailable'`), which is a semantic i18n key. Then on line 408 `tooltipId` is also set to a raw GEO_ERROR_MESSAGES value, and that value is passed directly to `formatMessage({ id: tooltipId })` as well as used as the button's `aria-label`. While `'Use my location'` does exist in all lang files (it's used by the existing `LocateMeControl`), `'Reset to north'` and `'Compass mode'` (used as tooltip IDs on lines 410, 412) also exist in the lang files — so these work today. However, the Cypress test on line 312 directly asserts `expect(labelOf()).toBe('Use my location')`, which tests the raw English string rather than the resolved translation, meaning the test would pass even if the i18n lookup silently failed. This is a pre-existing pattern (`'Use my location'` predates this PR), so it's not a regression introduced here — no action needed on that specific key. But `'Reset to north'` (line 411) is new in this PR: verify that all 16 lang files have received this key (checking `en.json` and a few others confirms it exists, so this appears fine). No Must Fix here on this one — but see the Suggestion below.

- **[`WaypointControl.jsx:48–57`]** The `handleContextMenu` callback closes over both `isTouch` and `isFullscreen`, which themselves come from hooks and may change independently. When `isFullscreen` transitions to `true`, the component re-renders, `useIsFullscreen()` returns the new value, but `useMapEvent` receives a new stable reference only if `isTouch` or `isFullscreen` changed — this part is correct. **However**, the early-return guard inside the callback (`if (!isTouch || !isFullscreen) return`) means the function body silently no-ops if the map is not fullscreen, even though `useMapEvent` still registers it and the handler accumulates on each re-render where the deps change. This works correctly *in practice* because the deps are stable, but if `isTouch` can change (e.g. the user connects a mouse), the old handler stays registered pointing at a stale closure. The current approach with `useCallback` + the early guard is the right structure — just worth a comment explaining that the early-return guard is intentional to avoid doing work outside the activation context, rather than being a misguided workaround for a listener-leak problem.

- **[`MapLocationContext.jsx:115`]** The `userLocation` memo includes `requestAwake` and `releaseAwake` in its deps and spread (`...geo, active, enable, disable, requestAwake, releaseAwake`). These are memoised with `useCallback([], [])` so their references are stable. But `geo` is the return value of `useGeolocation`, which memoises its own result — its identity only changes when `fix` or `error` changes. Spreading it into the context memo should therefore be fine. No issue here, but worth noting the correctness depends on `useGeolocation`'s own memo invariant holding.

- **[`LocationControl.jsx:341–344`]** The comment on `showNorthButton` says "a `follow → located` drag preserves the bearing, and re-entering `follow` from `located` also doesn't reset". The condition correctly covers all these cases now (`MODE.FOLLOW` included). Good — but the old `MODE.OFF` case is not listed: if somehow tracking is turned off while the bearing is non-zero (which shouldn't happen since follow and off both call `resetNorth()`), the north button wouldn't appear. This is guarded by the `resetNorth()` call in the mode effect, so it's fine in practice. Just flagging it as a comment opportunity.

- **[`useGeolocation.js`]** The `visibilitychange` re-subscribe only fires when `Date.now() - lastFixAtRef.current >= STALE_AFTER_MS` (15s). This is a good improvement over the original unconditional re-subscribe. The comment explains the tradeoff clearly. One edge case worth documenting: on an iOS bfcache restore the `pageshow` handler always re-subscribes (no staleness check, as noted in the comment). This means a user who backgrounds the app for less than 15s and comes back via bfcache will rebuild the watch unnecessarily. This is probably acceptable given how rare bfcache restores are and how cheap a re-subscribe is compared to the dialog cost it used to incur — worth a brief comment to close the loop.

- **[`WaypointHud.jsx:80`]** `L.latLng(userLocation.lat, userLocation.lng).distanceTo(...)` is called on every render. Since `userLocation` changes reference on every fix this is unavoidable, but it constructs two throwaway `L.LatLng` objects per render. This is micro-optimisation territory for most cases, but in compass mode the HUD re-renders at every heading tick (120ms). The existing `formatDistance` formatter cache already handles the heavier work; the `distanceTo` call is cheap enough to leave as-is.

- **[`LocationControl.jsx` — open cycle]** The PR description explicitly notes the deliberate absence of a `compass → off` path, and the code comment on line ~337 documents the decision (two-tap cycle protects against accidental dismissal mid-walk). This is an acceptable product decision. Consider linking to a GitHub issue so the trade-off is tracked and the decision can be revisited after field usage data.

- **[`useIsFullscreen.js:14–16`]** Accessing `map._isFullscreen` (a private field of `leaflet.fullscreen`) is flagged with an `eslint-disable no-underscore-dangle` comment, which is correct per project convention. The comment explains why (`leaflet.fullscreen`'s own flag, only reading that covers both native and pseudo fullscreen). This is fine. Worth noting that if `leaflet.fullscreen` is updated this coupling will break — a future improvement would be to watch the `fullscreenchange` DOM event as a fallback.

### Nitpicks (Optional)

- **[`NorthResetControl.jsx:1`]** Missing semicolon after `from 'react-intl'` at the end of the import line (confirmed by reading the diff at line ~3800). The linter should catch this, but it's in a new file so it might have been missed.

- **[`PolygonMap.jsx:28`]** `useGeolocationPermission` is imported directly from `'../../../../hooks/useGeolocationPermission'` rather than from the barrel `'../../../../hooks'` (which now exports it). Consistent with the existing `useGeolocation` import pattern in the same file, so this is intentional — just noting the inconsistency for anyone who maintains this file later.

- **[`MapLocationContext.jsx:98`]** The `active` boolean is correctly derived from `requestCount > 0` and kept out of the raw count to avoid fan-out re-renders (as described in the comment). The implementation is correct. A future reader might wonder why `active` isn't just `isWatching` — a comment noting they can diverge (keep-warm grace period keeps `isWatching` true after `requestCount` drops to 0) would prevent confusion.

- **[`LocationControl.test.jsx:313`]** `messages={}` is passed to `IntlProvider`. Per the project's testing conventions, translation keys used by the component under test should be listed explicitly. With an empty messages object, `formatMessage({ id: 'Use my location' })` will fall back to the raw key string, which happens to be the expected English text — so the test passes but tests the key name rather than the translation lookup. Consider listing the handful of keys the control uses (`'Use my location'`, `'Compass mode'`, `'Recenter on your location'`, `'Reset to north'`, the error keys) explicitly in the messages prop.

ClemRz (commented):
This is an excellent, deeply considered PR. The sensor architecture (shared `MapLocationProvider`, ref-counted lifecycle, permission-gated auto-start, watchdog improvements) is exactly the right design for this feature set, and the field-testing follow-up that tackled the repeated-permission-prompt was handled cleanly. All items from the second review are confirmed addressed. A few new observations below.

### Issues (Must Fix)

- **[`MapContainer.jsx:103`]** `FullscreenOnlyControls` registers `useMapEvent('exitFullscreen', () => { … })` with an inline arrow function. This is the same listener-accumulation issue that was flagged and fixed on `WaypointControl` in the prior round — every re-render of `FullscreenOnlyControls` adds a new Leaflet listener without removing the previous one. The component's state (`isFullscreen`) changes exactly when `enterFullscreen`/`exitFullscreen` fires, so every transition leaks one handler. Wrap with `useCallback`:
  ```suggestion
  const handleExitFullscreen = useCallback(() => {
    if (typeof map.setBearing === 'function') map.setBearing(0);
  }, [map]);
  useMapEvent('exitFullscreen', handleExitFullscreen);
  ```

- **[`LocationControl.jsx:397–413`]** The `tooltipId` / `label` derivation still uses the raw English string `'Use my location'` as the default tooltip ID in the `OFF` mode. The string exists in all lang files, so it resolves correctly today. But it creates an asymmetry with the semantic keys (`'location.error.denied'`, `'Compass mode'`, etc.) introduced in this PR: a reader encountering a raw sentence-as-key has to verify by searching lang files. Consider a unified approach (`'location.use'`), or at least a comment on the line noting this key predates the semantic-key convention so future authors know it is intentional.

- **[`useGeolocation.js:153–163`]** The "fresh session" effect clears `fix` (sets it to `null`) whenever `enabled` goes `true`. This is intentional and documented — it prevents a follow consumer from snapping to a stale position. However, the `useMemo` that builds the return value depends on `[fix, error]`, so setting `fix = null` in one microtask and then receiving a new position in the next produces a single visible flash of `hasLocation: false` → `hasLocation: true`. On a device with a cached GPS fix this happens in tens of milliseconds and is imperceptible, but on a slow device this could cause the follow-recenter effect in `LocationControl` to fire twice (once to clear, once on the new fix). This is an existing trade-off, not something introduced in this PR — just flagging it as a future optimization opportunity if the double-recenter ever becomes visible.

- **[`WaypointHud.jsx:97–104`]** The HUD is rendered inside an absolutely-positioned `Box` with `bottom: theme.spacing(2)`. On a rotating map in compass mode, this overlay stays screen-fixed, which is correct. But `zIndex: 1000` is a hard-coded value rather than a MUI `zIndex` token (e.g. `theme.zIndex.tooltip`). The waypoint off-screen indicator (`WaypointOffscreenIndicator.jsx:113`) also uses `zIndex: 1000`. If MUI ever adjusts its stacking context this could clip behind other overlays. Consider using `theme.zIndex.snackbar` (1400) or `theme.zIndex.tooltip` (1500) for anything that must sit above Leaflet controls and MUI modals, or at least document why `1000` is sufficient here.

### Nitpicks (Optional)

- **[`MapContainer.jsx:103`]** Beyond the Must Fix above, the `FullscreenOnlyControls` component could benefit from a brief comment on why its `useMapEvent('exitFullscreen', …)` is registered even when `!isFullscreen` (i.e. when the component renders `null`). The same pattern is explained in `WaypointControl`'s comment about the listener staying registered outside fullscreen — worth carrying the same explanation here for consistency.

- **[`useGeolocation.js`]** The module-level comment block at the top is thorough and genuinely useful — arguably the best-documented hook in the codebase. The `NO_TIMEOUT` constant (line 9) has an excellent comment explaining the `Infinity → 0` coercion risk. One minor point: the constant name does not follow the project's `SCREAMING_SNAKE_CASE` convention for constants — it does use it, which is correct — just noting it is correctly named.

- **[`WaypointOffscreenIndicator.jsx:55–60`]** The `setIndicator(prev => (prev === null ? prev : null))` on the on-screen transition avoids a render when nothing changed. This is correct but the ternary reads a little backwards (`prev === null` → keep `prev`, else set to `null`). `setIndicator(prev => (prev !== null ? null : prev))` reads more naturally as "only update if not already null". Either form works identically — purely a readability observation.

ClemRz (approved):
All items from all three prior rounds are confirmed addressed in the current branch state. The sensor architecture is solid and the codebase is in genuinely good shape. A handful of smaller observations below.

- **[`MapClusters/index.jsx:392`]** `WaypointNavigation` is mounted inside `HydratedMap` only when `isTouch && waypoint`, which is correct for not requesting the GPS from a bare waypoint restore. However the component is wrapped in `CustomMapContainer` inside `Index`, and that container already wraps its children in a `MapLocationProvider`. `HydratedMap` is therefore always inside a provider, even for non-touch users — the provider starts subscribing to `useGeolocation` / `useDeviceOrientation` on their first `enable()`/`requestHeading()` call, so there is no wasted subscription. No issue, but the coupling is worth a brief comment near the `WaypointNavigation` import in `MapClusters` so a future reader knows why it does not need its own provider.

- **[`LocationControl.jsx:88`]** `useRequestHeading(mode !== MODE.OFF)` is called unconditionally. When tracking is in `LOCATED` mode (auto-started from an already-granted permission, user never tapped), this requests the orientation sensor at once. The PR description says the HUD handles the gating itself (`compassSupported && compassError !== 'unavailable'`), and `LocationControl` already guards compass mode behind `canUseCompass`. But the heading request here is wider than that: it will start the orientation subscription for devices where `compassSupported` is true, even before the user taps. On devices that genuinely have a magnetometer this is harmless; on Android WebViews where `DeviceOrientationEvent` fires but `computeHeading` receives `NaN`-alpha events, the `noDataTimer` in `useDeviceOrientation` will fire and set `error='unavailable'`, leaving a 3-second head-start on that error when the user never asked for it. Consider narrowing the gate to `mode !== MODE.OFF && mode !== MODE.LOCATED || userActivated`.

- **[`WaypointOffscreenIndicator.jsx:52`]** `recompute` is added as the handler for `move`, `zoom`, `rotate`, `resize` and `viewreset` via `useMapEvents`. In compass mode the map rotates at up to ~8 fps (120 ms throttle on the orientation sensor), driving `recompute` at the same rate. The equality check on `setIndicator` is good, but `latLngToContainerPoint` is called unconditionally before the equality guard — including while the pin is on-screen and the indicator is `null`. Consider adding an early-exit: if the pin is on screen and `indicator` is already `null`, bail before the projection call. The on-screen path is the steady state while actively navigating (the user is walking *toward* the pin), and eliminating the projection call in that hot path is cheap insurance.

- **[`useWaypoint.js:21`]** `isValidWaypoint` is defined as a module-level function but the `useMemo` wrapping the derived `waypoint` value takes `[raw]` as its deps. The guard itself is perfectly stable (pure function of `raw`), so the memo is correct. One edge case: if `raw` is `{ lat: 45, lng: 5, someExtraKey: 'foo' }` (a future schema with additional fields), `isValidWaypoint` passes it through, and it reaches Leaflet as the waypoint object. Any consumer that destructures only `lat`/`lng` is fine, but `WaypointLayer` passes the whole object to `waypointIcon` and `WaypointHud` spreads it as a prop shape — `PropTypes.shape({ lat, lng })` silently accepts extra keys. This is very unlikely to be a problem in practice, but documenting the "forward-compatible: extra keys are ignored downstream" assumption in the comment would save a future head-scratch.

### Nitpicks (Optional)

- **[`MapContainer.jsx:96–100`]** `FullscreenOnlyControls` renders nothing when `!isFullscreen` but is always mounted (by design — so `handleExitFullscreen` stays registered). The `useMap()` call at line 97 returns the Leaflet map instance as a dependency for `useCallback`, which is correct and cheap. Just noting the `eslint-disable no-underscore-dangle` at `useIsFullscreen.js:22` is the only place in this PR that reaches into a plugin private field — confirmed it is still isolated to that single line, good.

- **[`utils/geo.js:74`]** The `formatterCache` `Map` is module-level and never cleared. For a frontend app with a handful of locales this is a non-issue (a few KB of `Intl.NumberFormat` instances), but it behaves as a subtle memory leak if the locale can vary unboundedly (unlikely in practice since the app fixes locales). Worth a one-line comment (`// ~3 formatters × supported locales — bounded in practice`) to prevent a future "are we leaking?" audit.

- **[`LocationControl.test.jsx:207–210`]** The test `'centers the user (no offset) while north-up'` asserts on the Y coordinate of `containerPointToLatLng` calls, passing `p[1] === 300` — which is the centre of the 600px-tall mock map. This is a meaningful assertion for the center-on-user path, but it depends on the mock returning `{ x: 400, y: 300 }` from `latLngToContainerPoint`. Since `userPoint.x` is passed through unchanged, the assertion would also pass if the component accidentally computed `y + 0` (i.e. no offset) for the wrong reason. Adding `expect(point[0]).toBe(400)` in the same test (the X should also pass through for north-up centering) would close that gap, though it's a minor point.

- **[`MapLocationContext.jsx`]** The exported `useRequestUserLocation` and `useRequestHeading` are the designed API surface for consumers outside this module. Neither has a JSDoc comment documenting its return type (`void`) and side-effect contract, unlike the provider itself which is well-commented. Adding a one-liner to each would help consumers know at a glance that calling the hook is the only required step.

urien (approved):
C'est parfait, je vais faire une section dans le wiki

PR #1486, opened 2026-08-01
fix(Pwa): replace the non-functional auto-update with an explicit update button

Paul-AUB:
﻿# fix(Pwa): replace the non-functional auto-update with an explicit update button

## 🤔 What

The PWA service worker update strategy switches from `autoUpdate` (silent, background) to `prompt` (explicit, user-driven).

- `registerType: 'prompt'` in `vite.config.mjs`
- New `<UpdatePrompt>` component: registers the service worker and shows a persistent snackbar — _"A new version is available"_ + **Update** / **Later**
- Adds the periodic update check that was missing entirely (hourly + on every return to the foreground)
- Removes the now-dead service worker registration and `controllerchange` reload block from `index.jsx`
- 2 new i18n keys (`A new version is available`, `Later`) translated in all 15 languages — `Update` already existed

## 🤷‍♂️ Why

**The current auto-update never updates anything.** Not "unreliably" — structurally, it cannot work. Three independent problems stack up:

### 1. The new service worker never activates

`vite.config.mjs` combines `registerType: 'autoUpdate'` with `injectRegister: false`. But the plugin only forces `skipWaiting` when `injectRegister` is `'auto'` — [vite-plugin-pwa/dist/index.js#L874](https://github.com/vite-pwa/vite-plugin-pwa/blob/main/src/options.ts):

```js
if ((injectRegister === 'auto' || injectRegister == null) && registerType === 'autoUpdate') {
  workbox.skipWaiting = true;
  workbox.clientsClaim = true;
}
```

The condition is false here, and our own `workbox` config only sets `clientsClaim`. Confirmed on the built artifact — the only `skipWaiting` in `dist/sw.js` is the message handler Workbox always ships:

```js
addEventListener("message", s => { s.data && "SKIP_WAITING" === s.data.type && self.skipWaiting() })
```

No top-level `self.skipWaiting()`. **The new service worker installs and stays in `waiting` forever.**

And nothing sends it that message either: in `autoUpdate` mode the client-side `virtual:pwa-register` deliberately skips it —

```js
const updateServiceWorker = async () => {
  await registerPromise;
  if (!auto) { sendSkipWaitingMessage?.(); }   // `auto` is true → no-op
};
```

Consequence: the whole `controllerchange` block in `index.jsx` is dead code — `controllerchange` can never fire. In practice the update only lands once **every** tab/window of the app is closed, which for an installed PWA/TWA left in the background may simply never happen.

### 2. No update check is ever triggered

A browser only re-checks `sw.js` when a document in scope is loaded (plus a 24h safety net). GrottoCenter is a client-routed SPA: after the initial load there is no document navigation, ever. `registerSW({ immediate: true })` performs a single check per cold start and nothing after that. This is the [Periodic SW updates](https://vite-pwa-org.netlify.app/guide/periodic-sw-updates.html) recipe from the vite-plugin-pwa docs, and it was missing.

### 3. `skipWaiting` mid-session would break code splitting anyway

So "just add `skipWaiting: true`" is not the fix. `App.jsx` lazy-loads ~45 routes. If a new SW claims clients while the page is running, `cleanupOutdatedCaches: true` drops the old precache — and on Azure Static Web Apps the old hashed chunks are gone from the server too. The first navigation to a not-yet-loaded route then fails with `Failed to fetch dynamically imported module` → `PageErrorBounary`. That would trade a silent bug for an error screen.

On top of that, the old `reloadWhenHidden` heuristic reloaded the page while the tab was **hidden** — on mobile that means the app is backgrounded and often frozen (Page Lifecycle), so the reload isn't guaranteed; and when the user comes back they get either the old bundle still, or a page silently reset (map position, form input lost).

## 🔍 How

`registerType: 'prompt'` makes the new service worker install and **wait**. The running page keeps the precache its own chunks live in, so lazy routes keep resolving after a deploy. The reload happens on a click, at a moment the user picked.

**`packages/web-app/src/components/appli/UpdatePrompt.jsx`** — uses `useRegisterSW` from `virtual:pwa-register/react`:

```js
const intervalId = setInterval(checkForUpdate, UPDATE_CHECK_INTERVAL_MS); // 1h
document.addEventListener('visibilitychange', handleVisibilityChange);    // back to foreground
```

The `visibilitychange` check is the one that matters for an installed PWA resumed from the background. Both are guarded by `navigator.onLine`, `registration.update()` rejections are swallowed (offline, captive portal, deploy in flight — the next tick retries), and the effect cleans both up.

Mounted in `ApplicationShell` next to `<ErrorHandler />`, deliberately **outside** the `ErrorBoundary`: when a stale build is what crashed the app, offering the update is exactly what fixes it.

A comment in `vite.config.mjs` documents the `autoUpdate` + `injectRegister: false` trap so this doesn't regress.

### Verified on the build output, not just at the source level

`dist/assets/index-*.js` after `yarn build`:

```js
tI=`false`, rI = tI===`true`              // auto = false → prompt mode
d = async () => { await l, rI || u?.() }  // → messageSkipWaiting() IS sent
c.addEventListener(`waiting`, t)          // t registers `controlling` → location.reload()
```

and `dist/sw.js` still carries the `SKIP_WAITING` listener + `clientsClaim()`. The full chain is closed: detect → wait → click → activate → reload.

## 🔗 Related API PR

None.

## 🧪 Testing

⚠️ **Not observable on a dev server** — `devOptions.enabled: false` means `virtual:pwa-register/react` resolves to a no-op hook in development. Test on the Azure staging build or a local `yarn build && yarn preview`.

1. `yarn build && yarn preview`, open the app, let the service worker install.
2. Change any source file, `yarn build` again, keep the preview server running.
3. Switch to another tab and come back (triggers the `visibilitychange` check), or wait an hour.
4. The snackbar appears → **Update** reloads onto the new build; **Later** dismisses it for the session.

Also worth checking: navigating between lazy routes after a deploy no longer risks `Failed to fetch dynamically imported module`.

### ⚠️ One-time migration caveat

Users already running the current build won't benefit from this PR immediately: their existing service worker is stuck in `waiting` and knows nothing about the new logic. They need one full app close (or the 24h browser check) to pick it up. From that point on, the button works.

**Checks:** `yarn lint` ✅ · `yarn build` ✅ · `yarn test` ✅ (84 files, 731 tests) · `translations:check-sorted` + `translations:check-en` ✅

## 📸 Previews

Snackbar, bottom-centered, `Alert severity="info"` with the `SystemUpdateAlt` icon:

```
┌────────────────────────────────────────┐
│ ⭳  A new version                       │
│    is available       UPDATE         ✕ │
└────────────────────────────────────────┘
```

A single text action plus a close icon, rather than two text buttons — on a 360dp screen the French and German labels ("Mettre à jour", "Aktualisieren") both wrapped onto two lines, and the theme's `MuiButton` default (`variant: 'contained'`) rendered the dismiss action as a filled grey block. The update button is explicitly `variant="text"` with `whiteSpace: 'nowrap'`; the dismiss action keeps its translated label as the icon button's `aria-label`.


ClemRz (approved):
Overall this is a well-researched fix with excellent documentation. The root-cause analysis in the PR body is thorough, the architectural trade-offs are clearly justified, and the implementation follows the project's conventions. No blocking issues.

---

- **[UpdatePrompt.jsx:34-37]** `useRegisterSW` also accepts `onRegisterError` and `onOfflineReady` callbacks. The offline-ready path is a no-op here (there's nothing to surface), but a silent failure during SW registration could leave the update mechanism completely dead with no trace. Consider adding at minimum a `console.warn` in `onRegisterError` so it surfaces in production monitoring:
  ```js
  onRegisteredSW: (_swUrl, swRegistration) =>
    setRegistration(swRegistration ?? null),
  onRegisterError: err => console.warn('[UpdatePrompt] SW registration failed', err)
  ```

- **[UpdatePrompt.jsx]** No unit test is included. `useRegisterSW` can be mocked (`vi.mock('virtual:pwa-register/react')`), making it straightforward to test the two visible states: snackbar hidden when `needRefresh` is `false`, and snackbar + both actions rendered when it's `true`. Given the testing checklist in `AGENTS.md`, this feels worth adding.

- **[UpdatePrompt.jsx:63-64]** The `Snackbar` has no `onClose` prop. MUI's `Snackbar` wraps content in a `ClickAwayListener` by default; without `onClose`, clicks outside do nothing silently (which is the desired intent here — force an explicit choice). To make that intent explicit and suppress any future MUI console warning, consider passing `onClose` as a no-op, or setting `disableWindowBlurListener` if needed. Alternatively, document why `onClose` is intentionally absent, as other contributors may add it reflexively thinking it's missing.

---

### Nitpicks (Optional)

- **[headRefName]** The branch name `fix/pwa-expicit-update` has a typo — "expicit" → "explicit". Not a blocker (the branch won't persist after merge), but worth noting for future reference.

- **[UpdatePrompt.jsx:31]** The `needRefresh` destructuring is readable but slightly unusual — it destructures a `[value, setter]` tuple directly from the hook's return object. A one-line comment noting that `useRegisterSW` returns signals as `[state, setState]` pairs (matching React's `useState` convention) would help orient future readers.

PR #1487, opened 2026-08-02
tech(lint): migrate to ESLint 9 flat config

Paul-AUB:
# tech(lint): migrate to ESLint 9 flat config

## 🤔 What

Migrates the repository from ESLint 8 + `.eslintrc` to **ESLint 9 with the flat
config format**, one of the 🔴 critical items of the audit in #1391.

- Drops `eslint-config-react-app` (dead since the Vite migration)
- Replaces the five cascading `.eslintrc.js` files and `.eslintignore` with a
  single `eslint.config.mjs` at the root
- Republishes `packages/eslint-config` as a flat config array
- Removes the unused `packages/eslint-config-typescript`
- Fixes everything the reactivated ruleset uncovered

`yarn lint` ends with **0 errors**. The 219 remaining warnings are tracked
design debt, each marked with a `TODO` in the shared config.

## 🤷‍♂️ Why

`eslint-config-react-app` sets `root: true`. Because `packages/web-app/.eslintrc.js`
extended it, **the shared Airbnb + Prettier config was never applied to the web-app
package at all** — the only package that matters. `yarn lint` was green because it
was running CRA's permissive `warn`-only ruleset on 916 source files.

That had to be resolved before the ESLint 9 migration anyway: the package pins
`eslint@^8` and loads `@rushstack/eslint-patch`, which patches ESLint internals
and throws outright on ESLint 9.

So the two are the same job — turning the shared config back on is what makes the
version bump possible, and it is what produced most of the diff.

## 🔍 How

Twelve commits, meant to be read in order.

### Reactivating the shared config

| Commit | |
|---|---|
| `tech(lint): drop eslint-config-react-app` | removes the package and the `root: true` that shadowed the shared config |
| `style(lint): apply prettier and autofixable rules` | `eslint --fix`, formatting only |
| `refactor(react): drop unused React imports` | codemod, the automatic JSX runtime makes them dead |
| `tech(lint): resolve the @ alias and package exports maps` | swaps `eslint-import-resolver-node` for the TypeScript resolver: it reads the `@/*` alias and honours `exports` maps (several deps ship no `main`) |
| `test(Storybook): modernize the stories left on the legacy API` | four stories still on `storiesOf` + `addon-knobs`, which no longer parse under Storybook 10 |

### Bugs the ruleset uncovered

| | |
|---|---|
| `fix(deps): declare d3-polygon and papaparse…` | both imported directly but only present as transitive deps of `d3` / `react-papaparse`. Works by accident today; a hoisting change in either parent breaks the build |
| `refactor(Tour): extract DontShowContext…` | `AppTour` ↔ `TourTooltip` import cycle, plus a context value rebuilt on every render |
| `fix(lint): clear the mechanical airbnb violations` | includes three latent bugs: `FormProgressInfo` imported `React` as a **named** export of `react` (so the binding was `undefined`), `LayersControl` and `DesktopEntityTable` each shadowed an outer binding, and both document actions kept an `if (err.isAuthError) return;` catch that swallowed every error either way |

### Rule policy

`tech(lint): set the policy for the remaining airbnb violations` splits the 330
leftovers in two:

- **Relaxed for good** — rules fighting conventions this codebase applies on
  purpose: single-named-export modules, `continue`/`++`, and the underscore
  identifiers coming from the API wire format, Leaflet internals or test seams
  (listed one by one rather than disabling the rule).
- **Demoted to `warn` with a `TODO`** — real debt: missing prop validation,
  nested ternaries, inconsistent return paths, the a11y backlog.

Three rules are configured rather than worked around: `no-unused-vars` honours the
`_` prefix, `no-restricted-exports` stops rejecting the `export { default } from`
barrel idiom, and `react/jsx-no-duplicate-props` no longer treats MUI
`inputProps`/`InputProps` as duplicates.

### The migration itself

`eslint-config-airbnb` has no flat build and is no longer released, so it goes
through `FlatCompat`. Every other plugin is wired through its own flat export,
with `eslint-plugin-cypress` bumped to v6 for one.

Two consequences worth knowing about:

- **`.mjs` files are linted for the first time** — the old `--ext` list left them
  out, which is why `vite.config.mjs` shows up reformatted (whitespace only).
- **Flat config reports unused `eslint-disable` directives by default**; 22 stale
  ones were removed. ESLint 9 also flags unused catch bindings, now written as
  optional catch bindings (`catch { … }`).

### Target version

ESLint **9**, not 10. `eslint-config-airbnb`, `eslint-plugin-import` and
`eslint-plugin-react` all cap their `eslint` peer at `^9`, and ESLint 10 removes
`.eslintrc` support entirely. Moving to 10 means first replacing Airbnb — worth its
own issue rather than riding along here.

## 🧪 Testing

```bash
yarn install
yarn lint     # 0 errors, 219 warnings
yarn test     # 78 files, 659 tests
yarn build
```

All three pass on this branch. The pre-commit hook was also realigned: its
lint-staged glob was still the old `--ext` list, so `.mjs`/`.cjs` were skipped on
commit while CI checked them.

Reviewers: `style(lint): apply prettier and autofixable rules` and the
`vite.config.mjs` part of the flat-config commit are pure formatting and can be
skimmed. The commits worth reading closely are `fix(lint): clear the mechanical
airbnb violations`, `fix(deps): …` and `refactor(Tour): …`.

## 🔗 Related API PR

None.

## 📸 Previews

No user-facing change.

Related to #1391.


ClemRz (commented):
This is a substantial, well-structured migration. The PR description is one of the clearest I've seen for a change of this scope — the commit-by-commit narrative makes it easy to follow the reasoning. The latent bugs surfaced by the reactivated ruleset (the `React` named import, the shadowing issues, the error-swallowing catches) are genuinely good finds. A few things worth discussing before merging.

---

### Issues (Must Fix)

**`AGENTS.md` contradicts the actual config**

`AGENTS.md` was updated to say "`yarn lint` must end with **0 errors and 0 warnings**", but the shared config at `packages/eslint-config/index.mjs` (line 150) leaves `react/jsx-filename-extension` set to `'warn'` with a `// TODO` comment, and `react/no-unstable-nested-components` is silenced entirely with `'off'`. The PR description itself confirms that 219 warnings remain.

This contradiction matters because AGENTS.md is used as authoritative context by AI agents and developers joining the project. Either:
- The standard should be "0 errors; warnings are tracked debt" (and AGENTS.md should say that), or
- The warnings should be promoted to `'error'` now (and the TODO items fixed before merging).

The current state will cause any agent or contributor following AGENTS.md to fail CI before they even touch a file — or worse, to ignore warnings they should be filing as issues.

---

**`CreateDocument` / `UpdateDocument` — the new `catch` silently drops auth errors**

`packages/web-app/src/actions/Document/CreateDocument.js` (line ~95) and `UpdateDocument.js` (line ~105):

```js
.catch(() => {
  // Failures are already reported to the store above; this catch only
  // stops the rejection from escaping as an unhandled promise.
});
```

The comment justifies suppressing the re-throw from the `throw new Error(...)` on an HTTP 4xx/5xx — that part is fine. But `checkAuthStatus(dispatch)` throws an error with `isAuthError: true` on a 401. That error bypasses all the `if (response.status >= 400)` logic above and falls directly into this catch, where it is now silently discarded. Previously the code was:

```js
.catch(err => {
  if (err.isAuthError) return;
});
```

…which was already swallowing the auth error (by design — `checkAuthStatus` already dispatched `logout()` and called `location.reload()`). The new form matches that behaviour, so this is not a regression. The comment is the real issue: it claims "failures are already reported to the store above", but an auth error is **not** reported to the store by the inner `.then()` chain — it's handled by `checkAuthStatus`. The comment should acknowledge this, otherwise the next person reading it will think auth failures are covered by the dispatch they can't see.

Consider:
```js
.catch(err => {
  if (err.isAuthError) return; // checkAuthStatus already dispatched logout()
  // All other failures were dispatched to the store above; suppress
  // the re-throw so it doesn't surface as an unhandled rejection.
});
```

---

**`packages/web-app/cypress/e2e/import-observations.cy.js` — fragile `.MuiMenu-paper` selectors not addressed**

The Cypress test was reformatted (good — it was needed) but all eight `.MuiMenu-paper` class-based selectors were kept as-is:

```js
cy.get('.MuiMenu-paper').last().within(() => { ... });
```

Per the project's testing conventions, `data-testid` is preferred over `.Mui*` class selectors. This file was touched as part of the PR, making it a natural time to introduce `data-testid` attributes on the affected `<Menu>` components. This isn't blocking — the selectors worked before and still work — but since the file is already in the diff it's worth flagging.

---

**`PROJECT_SUMMARY.md` still references stale versions**

Line 20:
```
- **Storybook 8.3.5** - Component development
- **Cypress 13.15.0** - E2E testing
```

The actual versions are Storybook 10 and Cypress 15. These docs are used by AI agents for context, so stale version numbers will generate incorrect advice. Minor fix, but worth correcting since the PR already touches this file.

---

**`packages/eslint-config/index.mjs` — `no-nested-ternary` and `consistent-return` not explicitly configured**

AGENTS.md now says these rules are "enforced at `error`", but neither appears in the config — they're inherited from Airbnb and there's no comment confirming that's intentional. If a future Airbnb update or `FlatCompat` behaviour change silently relaxes them, the config becomes inconsistent with the documented policy. A one-line comment next to the relevant Airbnb extends call (or explicit `'error'` entries) would make the intent durable.

---

### Nitpicks (Optional)

**`packages/eslint-config/index.mjs` line 149**: `'react/no-unstable-nested-components': 'off'` — the TODO comment is fine, but "off" rather than "warn" means violations won't surface at all in tooling output, making it harder to track the remaining instances. Demoting to `'warn'` would make the debt visible without blocking CI. (This does increase the warning count, which circles back to the AGENTS.md issue above.)

**`packages/web-app/src/components/common/Maps/_stories.jsx` and `src/components/common/Form/FormAutoComplete.jsx`** have `// eslint-disable-next-line react/prop-types` without a preceding justification comment. These are pre-existing, but flat config now requires all surviving directives to be necessary — if they're not flagged by the new ruleset they'll be auto-removed on the next `--fix`. Worth a quick check to confirm whether they're still needed.

ClemRz (changes requested):
This is a well-executed, thoughtfully sequenced migration. The commit-by-commit narrative in the description makes the reasoning easy to follow. The latent bugs surfaced by the reactivated ruleset are genuine finds (the `React` named import, the binding shadows, the error-swallowing catches). The prior review feedback was addressed cleanly — the AGENTS.md contradiction, the misleading catch comments, the stale PROJECT_SUMMARY.md versions, and the Cypress `.MuiMenu-paper` selectors are all resolved. A few small things remain before merging.

---

### Issues (Must Fix)

- **[packages/web-app/src/components/common/Highlights/index.jsx:54]** The second `eslint-disable react/no-array-index-key` in `HighLightsLine` (the `diffSentences` block) has no justification comment preceding it. The first block in the same file (`HighLightsChar`, line 33) sets the pattern correctly. This is a direct violation of the AGENTS.md rule: _"Never use `// eslint-disable` without a written justification in the preceding comment."_ A one-liner like the first block is all it needs:
  ```suggestion
    const result = diffSentences(oldText, newText);
    // A diff chunk has no identity beyond its offset in the result, and the
    // whole list is recomputed from (oldText, newText) on every render.
    /* eslint-disable react/no-array-index-key */
  ```

---

- **[packages/web-app/src/hooks/useLanguageSync.js:30, 43, 57]** Three `react-hooks/exhaustive-deps` disables were inherited as pre-existing tech debt, but this file was touched by the PR (Prettier reformat), and the new AGENTS.md policy now requires a written justification for every surviving disable. Each one has a clear reason (the intent is to run the effect only when a specific single dependency changes, not when the full closure updates), but that reason is not written down. Consider adding a short comment above each, e.g.:
  ```js
  // Intentionally scoped to `isAuth` only: other dispatch calls are
  // stable references and re-running on each would cause a loop.
  ```

- **[packages/web-app/cypress/e2e/import-observations.cy.js:28]** `.MuiAlert-root` (unauthenticated test) and `.MuiAutocomplete-popper` (line 71, device search test) are still class-based selectors. The prior review surfaced the `.MuiMenu-paper` ones, which were fixed. These two were missed. Neither is blocking — they worked before and still work — but they go against the testing conventions and are fragile across MUI major upgrades. Worth addressing while the file is in scope.

---

### Nitpicks (Optional)

- **[packages/eslint-config/index.mjs:49]** The `import/resolver` in the shared config still points to the `node` resolver (the one that ignores `exports` maps). The root `eslint.config.mjs` overrides this for `packages/web-app` with the TypeScript resolver, so the `web-app` files are correctly resolved. But any other future workspace package that doesn't get an explicit override will silently fall back to the node resolver. A comment noting this override dependency would help future maintainers understand why the shared default is intentionally kept as `node`.

- **[packages/web-app/src/components/common/Highlights/index.jsx:1]** `import React from 'react'` is still present (needed for `React.Fragment` with a `key` prop — correct). The codemod pass removed dead React imports across 204 files but this one is genuinely needed. Consider a brief inline comment (e.g. `// React.Fragment with key requires the React namespace`) to prevent a future automated pass from accidentally stripping it.

ClemRz (commented):
This is a carefully sequenced, well-described migration. The commit-by-commit narrative makes it easy to follow, the latent bugs surfaced by the reactivated ruleset are genuine finds, and every point from both prior review rounds has been addressed cleanly. Near the finish line — just a couple of small things to close out.

---

- **[packages/eslint-config/index.mjs:119–131]** `_southWest` and `_northEast` (Leaflet `LatLngBounds` internals) are used in two production files (`Map.jsx`, `MapMassif.jsx`) and two test files, yet they don't appear in the `no-underscore-dangle` allowlist. The `Map.jsx` block disable (`/* eslint-disable no-underscore-dangle */` … `/* eslint-enable */`) was reformatted in this PR and is now in the added lines — without a preceding justification comment that the policy requires. Adding them to the allowlist alongside `_leaflet_id`, `_map`, etc. would be more consistent with the config's stated approach ("listed one by one rather than disabling the rule") and would also fix `MapMassif.jsx`'s identical pre-existing block:
  ```js
  // Leaflet / leaflet-rotate internals we have to reach into
  '_leaflet_id',
  '_map',
  '_bearing',
  '_setPos',
  '_measureDblclick',
  '_southWest',   // LatLngBounds internals (getBounds result)
  '_northEast',
  ```

---

### Nitpicks (Optional)

- **[packages/web-app/src/components/common/AutoCompleteSearch/index.jsx:72]** `// eslint-disable-next-line react/prop-types` above `InputAdornments` has no justification comment. The file was touched in this PR (React import removed, `PropTypes.any` replaced) so the directive survived the diff — but under the new policy, all surviving disables need a written reason. A one-liner is enough, e.g.:
  ```js
  // Purely internal rendering helper; its props are enforced by the
  // AutoCompleteSearch PropTypes above, not at this call-site level.
  // eslint-disable-next-line react/prop-types
  const InputAdornments = ({ isLoading, hasError }) =>
  ```

ClemRz (approved):
This is a clean, well-executed landing of all three prior review rounds. Every issue and suggestion from rounds 1 and 2 was resolved before round 3, and round 3's items were addressed on the same day. The commit history remains readable, the latent bugs that were surfaced (the `React` named import, the binding shadows, the error-swallowing catches) are all fixed, and the ESLint 9 migration itself is solid. The `_southWest`/`_northEast` allowlist addition, the `useLanguageSync` exhaustive-deps comments, the `Highlights` justification comment, and the Cypress `data-testid` selectors are all in order.

Just one small item to close out before merging.

---

- **[packages/web-app/src/components/appli/Duplicates/HydratedDocumentDuplicates.jsx:1]** The file-level `/* eslint-disable react-hooks/exhaustive-deps */` has no justification comment. This is pre-existing debt, but this PR directly touched the file — it removed the four individual `eslint-disable-next-line` directives and consolidated them under the file-level disable, which is fine. Since this file is now in the diff, the AGENTS.md policy ("Never use `eslint-disable` without a written justification in the preceding comment") applies. A one-liner before the directive is enough:
  ```js
  // Each effect is intentionally scoped to a single state flag to avoid
  // re-running on every render cycle (these are sequential HTTP-response watchers).
  /* eslint-disable react-hooks/exhaustive-deps */
  ```

---

### Nitpicks (Optional)

- **[packages/web-app/src/actions/Document/CreateDocument.js:1, UpdateDocument.js:1]** Both files still import from `isomorphic-fetch`. Removing that dependency is tracked in #1391 (phase 1 quick wins), so this is clearly out of scope here — just noting it as a reminder that these two files will need a follow-up touch when that task lands.

- **[packages/eslint-config/index.mjs]** The `eslint-plugin-storybook` peer is declared but Storybook 10 ships its own flat-config export. If a future Storybook bump breaks `storybook.configs['flat/recommended']`, the fix is straightforward (update the import shape). Worth a quick smoke-test when Storybook is eventually upgraded to confirm the config still resolves correctly.

PR #1495, opened 2026-08-04
feat(Release): publish a GitHub release on every merge and show the version in the footer

Paul-AUB:
# feat(Release): publish a GitHub release on every merge and show the version in the footer

## 🤔 What

Every merge on `develop` now publishes a versioned GitHub Release automatically, and the running version is visible in the app footer.

- **New** `cliff.toml` — declares how versions are computed and how release notes are grouped
- **New** `.github/workflows/release.yml` — tags the version and publishes the GitHub Release
- `.github/workflows/azure-static-web-apps-*.yml` — resolves the same version and injects it into the build as `VITE_APP_VERSION`
- `.github/workflows/twa-build.yml` — Android tags move to the `app-x.y.z` namespace
- `Footer.jsx` — discreet, centered `Grottocenter web vX.Y.Z` under the license bar
- `twa/README.md` — documents the two tag namespaces
- `package.json`, `packages/web-app/package.json` — aligned on 3.0.0

## 🤷‍♂️ Why

Release communication was entirely manual: changes were written up by hand on Slack, Facebook and Blogger, with nothing tying them back to the repository. Meanwhile the project already enforces conventional commits through CommitLint — the raw material for a changelog was there, and unused.

There was also no way for anyone to tell which version was actually running in production.

Two constraints shaped the design:

- **No release PR.** Releases happen on merge, not after a second review step.
- **No second deployment.** A version bump committed back to `develop` would redeploy the site for a change that alters nothing, and would prompt every PWA user to update for it.

## 🔍 How

The version is computed, injected into the build, and only tagged afterwards — never committed:

```
merge on develop
  ├─ Azure    : git cliff --bumped-version → VITE_APP_VERSION → build → deploy
  └─ Release  : git cliff --bumped-version → tag → GitHub Release
```

Both workflows run on the **same commit** and read the **same** `cliff.toml`, so the deployed version and the tag cannot disagree. They are independent: the release job never gates a deployment, so a failure there leaves production untouched and is fixed by re-running the job.

`VITE_APP_VERSION` travels as a step `env:` var through Oryx into `import.meta.env`, exactly like the existing `VITE_TURNSTILE_SITE_KEY`. Nothing is written back to the repository — hence no extra commit, no `[skip ci]`, and a single deployment per merge.

**Bump rules** — every merge produces a release, including `chore(deps)`: `!`/`BREAKING CHANGE` → major, `feat` → minor, anything else → patch. `filter_unconventional = false` keeps non-conventional commits counting rather than vanishing.

**Tag namespaces** — `tag_pattern` accepts bare semver only. This is what keeps the Android tags out of the web version history, and the matching change in `twa-build.yml` means a web release no longer triggers a full Android build. The `app-` prefix is stripped before the version reaches Gradle, since `versionName` must stay strict semver.

**Re-run safety** — the release job skips publication when the computed tag already exists, so re-running it is harmless.

## 🔗 Related API PR

None.

## 🧪 Testing

Both derivations were verified against this repository:

```bash
npx git-cliff@2 --bumped-version     # 3.0.0 → "nothing to bump" until a commit lands
npx git-cliff@2 --bump --unreleased  # grouped notes, correct sections
```

Bump rules were checked on a throwaway repository across the five cases: chore-only → patch, `feat` → minor, `feat` + `fix` → minor, `!` → major, and `BREAKING CHANGE` footer → major.

To review the footer: run the app and scroll to the bottom of the homepage — it should read `Grottocenter web v0.0.0-dev` locally, and the tagged version once deployed.

Note that this PR's own preview deployment will display a `-preview` suffix, which is intended: a preview must never claim a version number that has not been tagged.

## 📸 Previews

Footer license bar, bottom of the homepage:

```
[CC-BY-SA]  Unless stated otherwise, the CC-BY-SA license applies…
[ODBL]      The ODBL license applies to all data that is not copyrighted.
                       Grottocenter web v3.0.0
```

## ⚠️ Note for the reviewer

The first release will be **3.1.0**, computed from the `3.0.0` tag already pushed on `develop`.

`package.json` is no longer updated automatically — the git tag is now the source of truth for the version, and the file only serves as a readable reference. There is no `CHANGELOG.md`: the notes live in the GitHub Releases.

One known limitation: if two merges land within a couple of minutes, the second build may compute its version before the first tag exists and display a number one patch behind. It corrects itself on the following merge. Removing it entirely would mean serialising release and deployment, which would make deployments depend on the release tooling — a trade rejected on purpose.

### DIFF
--- a/.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml
+++ b/.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml
@@ -18,12 +18,36 @@ jobs:
       - uses: actions/checkout@v5
         with:
           submodules: true
+          # Tags and full history: next-version.js derives the version from the
+          # commits since the last release, and a shallow clone has neither.
+          fetch-depth: 0
+      # The version the Release workflow is about to tag, derived from the same
+      # cliff.toml on the same commit so the deployed bundle and the tag always
+      # agree. Nothing is written to the repository — it only travels as an env
+      # var into the build, the way VITE_TURNSTILE_SITE_KEY already does.
+      - name: Resolve the app version
+        id: version
+        run: |
+          set -euo pipefail
+          VERSION="$(npx --yes git-cliff@2 --bumped-version)"
+          if ! printf '%s' "$VERSION" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::git-cliff returned an unusable version ('$VERSION')"
+            exit 1
+          fi
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            # A preview is not a release: never let it claim a version number
+            # that has not been tagged.
+            VERSION="${VERSION}-preview"
+          fi
+          echo "value=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "Building GrottoCenter $VERSION"
       - name: Build And Deploy
         id: builddeploy
         uses: Azure/static-web-apps-deploy@v1
         env:
           NODE_VERSION: 20
           VITE_TURNSTILE_SITE_KEY: ${{ secrets.VITE_TURNSTILE_SITE_KEY }}
+          VITE_APP_VERSION: ${{ steps.version.outputs.value }}
         with:
           azure_static_web_apps_api_token: ${{ secrets.AZURE_STATIC_WEB_APPS_API_TOKEN_ORANGE_ROCK_0D4F87503 }}
           repo_token: ${{ secrets.GITHUB_TOKEN }} # Used for Github integrations (i.e. PR comments)
--- a/.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml
+++ b/.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml
@@ -9,6 +9,11 @@ on:
     branches:
       - develop
 
+env:
+  # Pinned to a major so a git-cliff release cannot silently change how versions
+  # are computed between two deployments. Keep in sync with release.yml.
+  GIT_CLIFF: git-cliff@2
+
 jobs:
   build_and_deploy_job:
     if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.action != 'closed')
@@ -29,7 +34,7 @@ jobs:
         id: version
         run: |
           set -euo pipefail
-          VERSION="$(npx --yes git-cliff@2 --bumped-version)"
+          VERSION="$(npx --yes "$GIT_CLIFF" --bumped-version)"
           if ! printf '%s' "$VERSION" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+$'; then
             echo "::error::git-cliff returned an unusable version ('$VERSION')"
             exit 1
--- a/.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml
+++ b/.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml
@@ -23,8 +23,8 @@ jobs:
       - uses: actions/checkout@v5
         with:
           submodules: true
-          # Tags and full history: next-version.js derives the version from the
-          # commits since the last release, and a shallow clone has neither.
+          # Tags and full history: git-cliff derives the version from the commits
+          # since the last release, and a shallow clone has neither.
           fetch-depth: 0
       # The version the Release workflow is about to tag, derived from the same
       # cliff.toml on the same commit so the deployed bundle and the tag always
--- a/.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml
+++ b/.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml
@@ -9,6 +9,15 @@ on:
     branches:
       - develop
 
+# Two deployments to the same target must not race: on two quick merges the
+# slower upload could land after the newer one. Keyed on the target rather than
+# flat, so previews — which each own a distinct environment — queue per PR
+# instead of behind production, and a PR teardown waits for that PR's upload.
+# Never cancel in progress: an interrupted upload leaves the target half-built.
+concurrency:
+  group: azure-deploy-${{ github.event.pull_request.number || 'production' }}
+  cancel-in-progress: false
+
 env:
   # Pinned to a major so a git-cliff release cannot silently change how versions
   # are computed between two deployments. Keep in sync with release.yml.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,10 @@ on:
   push:
     branches: [develop]
 
-  # Allows you to run this workflow manually from the Actions tab
+  # Allows you to run this workflow manually from the Actions tab. No inputs:
+  # a manual run always re-derives the version from the tip of develop, so
+  # this is only useful to retry after a transient failure, not to target a
+  # specific commit or override the computed version.
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,12 +11,6 @@ on:
   push:
     branches: [develop]
 
-  # Allows you to run this workflow manually from the Actions tab. No inputs:
-  # a manual run always re-derives the version from the tip of develop, so
-  # this is only useful to retry after a transient failure, not to target a
-  # specific commit or override the computed version.
-  workflow_dispatch:
-
 permissions:
   contents: write
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,88 @@
+# Publishes a release on every merge into develop — no release PR, no version
+# bump commit. The tag is the source of truth for the version; the deployed
+# bundle carries the same number because the Azure workflow derives it from the
+# same cliff.toml, on the same commit.
+#
+# This job never blocks a deployment: it runs alongside the Azure one, so a
+# failure here leaves production untouched and is fixed by re-running the job.
+name: Release
+
+on:
+  push:
+    branches: [develop]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+permissions:
+  contents: write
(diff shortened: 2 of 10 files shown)


ClemRz (changes requested):
Overall this is well-designed: the version derivation approach (same `cliff.toml`, same commit, no write-back) is sound, and the concurrency guard + idempotency check on the release job are exactly the right safety nets. Comments are thorough and the `set -euo pipefail` discipline is consistent throughout. A few things to address before merge.

### Issues (Must Fix)

- **[twa/README.md:153, 185]** `release-please` is mentioned twice as the tool that creates the `X.Y.Z` tags and ships the `CHANGELOG.md`, but this PR introduces `git-cliff` (not release-please), and the PR description explicitly states *"There is no `CHANGELOG.md`: the notes live in the GitHub Releases"*. The table row and the inline reference both need correcting — otherwise contributors will search for a release-please workflow that doesn't exist and expect a `CHANGELOG.md` that will never appear.

  Line 153 — suggest:
  ```
  | `X.Y.Z`     | git-cliff (`.github/workflows/release.yml`)        | the web app (release notes in GitHub Releases) |
  ```
  Line 185 — suggest:
  ```
     the web one `.github/workflows/release.yml` publishes under the bare `X.Y.Z` tag.
  ```

- **[.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml:32]** The Azure workflow hardcodes `git-cliff@2` inline, while `release.yml` centralises it in a `GIT_CLIFF` env var. If the major version ever needs bumping (2→3), this spot would be easy to miss. Consider extracting it to an `env:` block at the workflow level, matching the pattern already used in `release.yml`.

  ```yaml
  env:
    GIT_CLIFF: git-cliff@2
  ```
  Then use `npx --yes "$GIT_CLIFF" --bumped-version` in the `run` block.

- **[.github/workflows/release.yml:15]** The `workflow_dispatch` trigger has no inputs. This means a manual run always re-derives the version from the tip of the default branch — useful for a re-run, but there's no way to target a specific commit or override the computed version. That's fine for the primary use-case (re-run after a transient failure), but the limitation is worth a short comment so future maintainers understand why no `version` input exists.

### Nitpicks (Optional)

- **[cliff.toml:37]** The `commit_parsers` list covers all types from `commitlint.config.js` except `revert`. `^revert` is already captured (line 40 in the diff), so this is fine — just confirming the coverage is intentional. ✓

- **[.github/workflows/release.yml:88]** `make_latest: true` unconditionally marks every web release as the latest GitHub release. If someone ever pushes a patch for an older minor while a newer minor is already out, this would demote the newer release. For a project that releases linearly on `develop` this is the right default, but it's worth knowing the trade-off.

ClemRz (commented):
All three must-fix items and both suggestions from the previous review have been addressed — the README was corrected, the `GIT_CLIFF` env var was extracted, and the `workflow_dispatch` comment was added. A few things remain before merge.

### Issues (Must Fix)

- **[.github/workflows/twa-build.yml:14]** `release-please` is still named as the tool that creates the web tags. This PR introduces `git-cliff`, not release-please, and the same incorrect reference was already caught and fixed in `twa/README.md` in the previous round — it was just missed in the YAML comments. Suggest:
  ```
  # semver (1.0.0) belongs to git-cliff (.github/workflows/release.yml), which
  # tags the web app on every merge into develop. Android ships on its own cadence,
  ```

- **[.github/workflows/twa-build.yml:118]** Same stale attribution one comment down:
  ```
  # bare semver tag — that one is git-cliff's web release.
  ```

- **[.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml:26]** The `fetch-depth: 0` comment says `next-version.js derives the version`, but no such file exists in the repository — it's `git-cliff` that does this. Suggest:
  ```
  # Tags and full history: git-cliff derives the version from the commits since
  # the last release, and a shallow clone has neither.
  ```

- **[.github/workflows/release.yml:18]** `workflow_dispatch` has no `branches` filter, so it can be triggered manually from any ref — a feature branch, a tag, anything. For a re-run-after-failure use case that's fine, but it also means someone can accidentally publish a release from a non-develop tip. Consider adding a note to the existing comment, or — if you want a hard guard — a job-level step that checks `GITHUB_REF` and exits early when it's not `refs/heads/develop`.

### Nitpicks (Optional)

- **[packages/web-app/src/pages/homepage/Footer.jsx:370]** `Grottocenter web v{appVersion}` is a hardcoded English string. The architecture convention flags hardcoded text as a pattern to avoid. In practice the product name and version number aren't localised, and the Footer already has a few hardcoded brand strings, so this is a low-stakes gap — just worth noting for consistency.

ClemRz (changes requested):
All must-fix items from the previous two rounds are resolved: the `release-please` and `next-version.js` stale references are gone, the `GIT_CLIFF` env var is extracted correctly, and the README table is accurate. The design holds up well — same `cliff.toml`, same commit, no write-back, solid concurrency guard.

One regression landed in the last batch of commits, and a minor convention note:

### Issues (Must Fix)

- **[.github/workflows/release.yml]** `workflow_dispatch` was added in `82204e93` (with the correct explanatory comment requested in the previous round), then silently removed in `0735cbd7`. Without it there is no way to re-run the release job after a transient failure short of waiting for the next push to `develop` — which means a failed release either stays broken until the next merge, or requires force-pushing a dummy commit to trigger a rerun. The PR description itself calls out "fixed by re-running the job" as the recovery path, which is no longer possible. Suggest restoring it with the comment from `82204e93`:

  ```yaml
  on:
    push:
      branches: [develop]

    # No inputs: a manual run always re-derives the version from the tip of
    # develop, so this is only useful to retry after a transient failure,
    # not to target a specific commit or override the computed version.
    workflow_dispatch:
  ```

### Nitpicks (Optional)

- **[git history]** Both fixup commits use `fix: PR comments` with no scope, which does not conform to the project's CommitLint convention (`type(scope): description` — scope is required). Since this branch will likely be squash-merged it won't survive into `develop`, but worth aligning for consistency. Consider something like `fix(Release): address review comments` before merge.

ClemRz (changes requested):
The overall design is solid — same `cliff.toml`, same commit, no write-back, race-free concurrency guard, idempotent re-run logic. The README and YAML comment corrections are accurate. A must-fix from the previous round came back in the last commit, and two smaller items are worth addressing.

### Issues (Must Fix)

- **[`.github/workflows/release.yml`]** `workflow_dispatch` was restored in `82204e93` (with the explanatory comment requested in round 2), then removed again in `f9a085a7`. The PR description explicitly states the recovery path is "fixed by re-running the job" — that is only possible with `workflow_dispatch`. Without it, a transient failure (network blip, npm registry hiccup, GitHub API timeout) leaves that release permanently unpublished until the next push to `develop`. The idempotency guard (`git rev-parse --verify`) already makes re-running safe. Please restore:

  ```yaml
  on:
    push:
      branches: [develop]

    # No inputs: a manual run always re-derives the version from the tip of
    # develop, so this is only useful to retry after a transient failure,
    # not to target a specific commit or override the computed version.
    workflow_dispatch:
  ```

- **[`.github/workflows/azure-static-web-apps-orange-rock-0d4f87503.yml`]** The Azure `build_and_deploy_job` does not have a `concurrency` guard. On two quick merges to `develop`, both jobs can race: they will compute the same `--bumped-version`, inject it into their bundle, and both deploy — no harm done to the tag (the release workflow serialises that), but the second Azure deploy may overwrite the first before it is fully warm. The release workflow already handles this with `concurrency: group: release / cancel-in-progress: false`. A matching guard on the Azure workflow would keep the deployment queue deterministic:

  ```yaml
  concurrency:
    group: azure-deploy
    cancel-in-progress: false
  ```

- **[`packages/web-app/src/pages/homepage/Footer.jsx:363`]** `Grottocenter web v{appVersion}` is a hardcoded English string. This was noted as a low-stakes nitpick in the previous round and the author confirmed the Footer already has other hardcoded brand strings, so this is a conscious exception — but the right fix would be a translation key such as `app-version-label` with `defaultMessage="Grottocenter web v{version}"`. Worth tracking if i18n completeness ever becomes a hard gate.

### Nitpicks (Optional)

- **[git history]** Both fixup commits (`6c20ba42`, `f9a085a7`) use `fix: PR comments` with no scope, which does not conform to CommitLint (`type(scope): description` — scope is required). Since the branch will likely be squash-merged this won't reach `develop`, but aligning before merge (e.g. `fix(Release): address review comments`) keeps the log clean and avoids the pre-commit hook firing on future checkouts of this branch.

- **[`cliff.toml:43`]** `tech` and `refactor` both map to `<!-- 6 -->🔧 Technical`. Duplicate group keys are resolved at render time by grouping, so this works correctly — just confirming the intent is to surface both commit types under the same changelog section. ✓

ClemRz (approved):
The concurrency guard on the Azure workflow and the `GIT_CLIFF` env var extraction from the previous round are both in place — those look good.

- **[`packages/web-app/src/pages/homepage/Footer.jsx:357`]** The version `Typography` has no top margin. The preceding `LicenseLine` only has `sx={{ mb: 0.25 }}` (2px), so the version line sits very close to the ODbL text. A small `mt` — e.g. `mt: 1` (8px) — would visually separate it from the legal notices and make the colophon role clearer. The MUI shorthand `sx={{ mt: 1, ... }}` (or `theme.spacing(1)`) would be appropriate here.

### Nitpicks (Optional)

- **[`packages/web-app/src/pages/homepage/Footer.jsx:363`]** `Grottocenter web v{appVersion}` is still a hardcoded English string — noted as low-stakes in previous rounds and the author confirmed it's a conscious exception. Flagging only for completeness: the natural fix would be a translation key like `app-version-label` with `defaultMessage="Grottocenter web v{version}"`. No blocker.

- **[git history]** The fixup commits use `fix: PR comments` (no scope), which doesn't conform to commitlint (`type(scope): description`). Since the branch will almost certainly be squash-merged this won't reach `develop`, but aligning before merge (e.g. `fix(Release): address review comments`) avoids the pre-commit hook firing on future checkouts. No blocker.

urien (approved):
Merci @Paul-AUB pour ce nouveau service
C'est tout a fait satisfaisant

PR #1497, opened 2026-08-04
fix(document): use authorsOrganization consistently instead of authorsGrotto

Paul-AUB:
# fix(document): use authorsOrganization consistently instead of authorsGrotto

## 🤔 What

- Renames every remaining `authorsGrotto` reference to `authorsOrganization` in the Document form:
  - `Provider.jsx` (default attributes, validation, initial values normalization)
  - `AuthorsSection/index.jsx`
  - `types.js` PropTypes shape
  - `DuplicatesHandler/DocumentsHandler.jsx`
  - `documentTypeHelpers.js` payload fields
- Removes the now-obsolete workaround comment explaining the read/write field name mismatch.

closes #1496

## 🤷‍♂️ Why

The API used to read organization-authors as `authorsOrganization` but expected them back as `authorsGrotto` on write, forcing the form to translate between the two names internally. That asymmetry no longer exists, so keeping two names around was just a source of confusion and bugs.

## 🔍 How

Straight rename of the field across the Document form state, validation, PropTypes, duplicate-merge handling, and payload building — no behavioral change beyond dropping the now-unnecessary translation.

## 🔗 Related API PR

https://github.com/GrottoCenter/grottocenter-api/issues/1729

## 🧪 Testing

- Create/edit a document and add an organization as author, save, and confirm it persists correctly.
- Merge two duplicate documents that both have organization authors and confirm the merged result keeps them.

## 📸 Previews

_N/A — internal field rename, no visual changes._

### DIFF
--- a/packages/web-app/src/components/appli/EntitiesForm/Document/Provider.jsx
+++ b/packages/web-app/src/components/appli/EntitiesForm/Document/Provider.jsx
@@ -21,7 +21,7 @@ export const defaultDocAttributes = {
   datePublication: '',
   creatorComment: '',
   authors: [],
-  authorsGrotto: [],
+  authorsOrganization: [],
   editor: null,
   library: null,
   type: -1,
@@ -71,7 +71,7 @@ const checkFormValidation = document => {
   )
     isValid = false;
 
-  if (document.authors.length + document.authorsGrotto.length === 0)
+  if (document.authors.length + document.authorsOrganization.length === 0)
     isValid = false;
   if (!isDocumentPagesFormatValid(document.pages)) isValid = false;
   if (document.identifier && !document.identifierType) isValid = false;
@@ -115,10 +115,7 @@ const normalizeInitialValues = values => {
     ...rest,
     selectOptionAuthorizationDocument: option ?? null,
     files: (files ?? []).map(f => ({ ...f, state: IS_INTACT })),
-    // The API reads organization-authors as `authorsOrganization` but expects
-    // them back as `authorsGrotto` on write; the form only ever uses the
-    // write-side name internally.
-    authorsGrotto: authorsOrganization ?? []
+    authorsOrganization: authorsOrganization ?? []
   };
 };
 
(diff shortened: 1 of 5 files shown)


ClemRz (approved):

- **[`AuthorsSection/index.jsx:213`]** The `required` prop change (`required` → `required={value.length === 0}`) is a UX improvement — removing the asterisk once at least one author is selected is correct behaviour. However, this behavioural change is **not mentioned in the PR description**, which says "no behavioral change beyond dropping the now-unnecessary translation." Consider adding a note to the PR description, and verify that MUI's `filled` variant TextField still renders the label correctly when `required` toggles from `true` to `false` at runtime (no layout shift or label overlap).

- **[`DuplicatesHandler/DocumentsHandler.jsx:155–157`]** The `onlyIfFilled` guard already handles the empty-array case correctly. Just a heads-up: `newItems` from `retrieveFromObjectCollection(authorsOrganization)` is silently discarded (the comment explains why — no atomic create-and-attach endpoint yet). Consider a `/* newItems intentionally unused */` inline comment or destructuring as `_newItems` to make the suppression explicit and avoid a future "dead code" question from reviewers.

### Nitpicks (Optional)

- **[`Provider.jsx` — PR description]** The "Related API PR" field reads `[_N/A_](https://github.com/GrottoCenter/grottocenter-api/issues/1729)` — the link points to the backend issue (not N/A). Worth fixing to avoid confusion for anyone tracing history.

- **[`DuplicatesHandler/DocumentsHandler.jsx:124`]** The local variable rename from `previousAuthorsGrotto` → `previousAuthorsOrganization` is correct. One minor note: the surrounding variable `authorsOrganization` (the state array) and the destructured `previousAuthorsOrganization` now share a very similar name in close proximity. The current scoping makes it unambiguous, but a brief inline comment like `// pre-existing orgs only (see comment above)` would make the intent clear at a glance.

---

Overall: the rename is complete and consistent across all five files — `grep` confirms zero remaining `authorsGrotto` occurrences. The `filterDocumentPayload` change in `documentTypeHelpers.js` ensures `authorsOrganization` reaches the API for both create and update paths. The `normalizeInitialValues` fix correctly maps the read-side field to the form state. Nice, focused PR.

PR #1498, opened 2026-08-05
feat(offline): make the app honest and usable without a connection

Paul-AUB:
# feat(offline): make the app honest and usable without a connection

## 🤔 What

Second pass on the offline experience, now that the service worker from the PWA
work is in place. The app no longer *pretends* to be online: every screen that
can fail without a network now says what is actually going on, and every control
that cannot work is disabled with a reason rather than left to fail.

- **Offline is an `info`, not an `error`** — new `FetchErrorState` replaces the
  red "data is not available" alert on the 8 entity pages (Entrance, Cave/Network,
  Massif, Organization, Person, Country, Region, Document) when the failure is a
  network one.
- **No dead Retry buttons** — new `useRefetchOnReconnect` hook repairs the view
  by itself the moment the connection returns; the Retry button is only rendered
  when pressing it can actually succeed.
- **Disabled, never hidden** — new `OfflineDisabled` wrapper explains why a
  control is unavailable (tooltip). Applied to the write and search actions:
  login, entity creation/edition, document upload, advanced search, cave /
  organization / document search forms, and the Account page panels.
- **The map tells the truth** — `OfflineDetailNotice` explains an empty map at
  detail zoom offline ("this area was never visited online"), instead of letting
  it read as "there are no caves here". Map tiles are now cached by the service
  worker, so relaunching the PWA offline shows the map it showed yesterday.
- **One snackbar stack** — the notification system is unified on a single
  notistack provider; `UpdatePrompt` no longer renders its own MUI `<Snackbar>`.
- **Session expiry warning moved to a snackbar** — closes #1489.
- **Storage protection** — an installed PWA/TWA asks for eviction-proof storage,
  and the Account page shows a "Protected" chip when the browser granted it.

## 🤷‍♂️ Why

Offline was technically working (the SW cached things) but the UI was still
written for a permanent connection, so the field experience was misleading in
three different ways:

1. **Cached data looked broken.** A page whose fetch failed offline showed the
   same red error as a genuine server failure. The data exists, it just hasn't
   been downloaded — that is not an error, and telling a caver "this cave is not
   available" makes them think the record is gone.
2. **Controls failed silently.** Save, upload, search, login — all clickable
   offline, all guaranteed to fail, none of them saying why. The user gets an
   opaque failure and no way to understand it.
3. **The map lied.** Above the clustering threshold the map draws per-tile
   markers. Offline, an area never visited returned nothing and the map rendered
   empty — indistinguishable from a karst with no caves in it. Worse, the tile
   cache was in-memory only, so relaunching the app offline lost everything.

Underground and in the field is *the* use case for this app, and it is exactly
where there is no signal.

### Telling "no network" apart from "the server said no"

`utils/networkError.js` — `isNetworkError(error)` normalises the four error
shapes this codebase stores in Redux (raw `Error`, `makeErrorMessage()` output,
the structured API `{ code, message, details, status }`, and the bare string a
few reducers keep). A numeric `status` wins outright and returns `false`: a 404
fetched before losing signal is still a 404 offline, and mislabelling it would
send the user hunting for a network problem that doesn't exist. Engine wording
differs for the same failure (Chrome "Failed to fetch", Firefox "NetworkError…",
Safari "Load failed"), hence the regex.

`FetchErrorState` pairs that with `useOnlineStatus`: "Failed to fetch" can also
be raised **while online** (API down, DNS, CORS), and promising "it will load
when you're back online" to someone who never left would be wrong. It only
claims offline when we actually are.

### Repairing instead of retrying

`useRefetchOnReconnect(refetch, enabled)` fires only on a real offline → online
transition, never on mount. `refetch` is kept in a ref so the usual inline arrow
call site doesn't re-run the effect every render and loop.

### Guarding actions

`OfflineDisabled` only adds the *explanation* — callers pass `disabled`
themselves, so they keep their own logic (e.g. on the Account page, the
Join/Add toggle blocks **opening** the panel offline but always allows closing
it). The `<span>` wrapper is required, not cosmetic: MUI sets
`pointer-events: none` on a disabled button, so the Tooltip would never fire.

### Service worker (`vite.config.mjs`)

- **New `api-map-tiles` rule** (`StaleWhileRevalidate`): map viewport data is
  requested on a fixed slippy-tile grid, so the same tile always yields the same
  URL — which is what makes it cacheable at all. `cacheKeyWillBeUsed` drops the
  `zoom` param (sent by the client, read by no `/geoloc` controller) so we keep
  one entry per tile rather than one per tile×zoom pair. Rule ordering matters
  and is documented inline: it must sit after the two bulk rules and before
  `api-get`.
- **Bulk coordinates `maxEntries` 5 → 40.** The massif page fetches
  `entrancesCoordinates` with a per-massif bbox, so every visited massif adds an
  entry to the same LRU. At 5, browsing two massifs evicted the world-wide
  entrance dataset and emptied the global map offline.
- **`handlerDidError` on both tile rules** returns an empty SVG. `CacheFirst`
  otherwise rejects with `no-response` per missing tile — dozens per pan,
  drowning the console; the area now just renders as plain map background.

### Notifications

`SnackbarProvider` is configured once (bottom-center — bottom-left would sit
under the 240px persistent SideMenu, `preventDuplicate`, `dense` below `sm`).
`maxSnack` is 4 because three snackbars can be `persist: true` at the same time
(offline, SW update, session expiry) and notistack dismisses the oldest instead
of queueing once every slot is persistent. `useNotification` now forwards
notistack options and exposes `onClose`, so nothing needs `useSnackbar`
directly anymore.

### #1489

The admin session-expiry warning was a full-width `Alert` rendered between
`AppBar` and `SideMenu`, outside `MainWrapper` — and only `MainWrapper` carries
`margin-left: theme.sideMenuWidth`, so the persistent Drawer covered the first
240px of the banner and cut the sentence in half on desktop. It is now a
snackbar (`position: fixed`, no layout offset to get wrong), with the threshold
logic extracted into `useSessionExpiry` so it is testable without a real clock.
Both layout traps are now documented on the `ApplicationLayout` tree.

## 🔗 Related API PR

None — frontend only.

## 🧪 Testing

New unit tests: `networkError`, `FetchErrorState`, `OfflineDisabled`,
`useRefetchOnReconnect`, `NetworkStatusNotifier`, `SessionExpiryNotifier`,
`FormContainers`, `SectionCreateButton`, `CustomIcon`, `mapTileCache`,
`offlineCache`, `actions/Intl`. Updated: `UpdatePrompt`, `Locations`.

Three of them cover a trap rather than a behaviour, and are the ones to keep if
the files are ever refactored:

- `actions/Intl` — an empty `messages[locale]` must **not** count as loaded,
  otherwise a locale whose fetch failed offline stays broken until a reload.
- `CustomIcon` — the disabled entity badge is dimmed by colour, never by
  `opacity`; the test reads the CSS Emotion actually emitted, since jsdom's
  `getComputedStyle` does not resolve `.Mui-disabled &`.
- `offlineCache` — the baseline is sampled repeatedly because `caches.delete()`
  resolves before the bytes are reclaimed; the test feeds a decreasing quota
  sequence and expects the lowest reading to win.

Manual, with DevTools → Network → Offline (or airplane mode on a phone):

1. **Entity pages** — open an entrance online, go offline, reload: the page
   comes from cache. Open an entrance *never visited*: blue "Not available
   offline" notice, no Retry button. Go back online without touching anything —
   the content fills in on its own.
2. **Actions** — offline, hover Save / Add / Upload / Search / Login: disabled
   with an "Unavailable offline" tooltip. Nothing disappears from the layout.
   Same at the *door* of each write flow: Contribute in the side menu, "Create
   an entrance here" in the map context menu, "New Message". On a disabled
   entity-creation icon, check the small badge stays opaque — the icon must not
   show through it.
3. **Paginated lists** — open Documents / Notifications / Messages online, go
   offline and reload: the list is still there (the API answers `206`, which the
   Cache API refuses as-is; the SW now stores it as a `200` and keeps
   `Content-Range`, so the pagination count stays right).
4. **Messages** — offline, the conversation list and an opened conversation come
   from cache; switch to **Archived** or page forward (never fetched, so never
   cached): "Not available offline" notice, and the list fills in by itself once
   the connection is back. Archive / Unarchive and Send are disabled with the
   offline tooltip, not left to fail.
5. **Map** — browse an area online, then offline: tiles and markers are still
   there after a full reload (install the PWA to test the real launch path).
   Pan to a never-visited area at detail zoom: the notice explains the empty map
   instead of it reading as "no caves here". Then, **without leaving the page**,
   go back online: the grey basemap squares re-request themselves and the
   clusters/markers refill — no navigation away and back, no manual pan.
6. **Snackbars** — trigger a SW update while offline: both notices stack, none
   overlaps. Toggle the connection: "You are offline" is persistent and
   dismissible, "You are back online" auto-hides, and it does **not** fire on a
   language change.
7. **Language** — switch language offline: it works if that `/lang/*.json` was
   cached, and otherwise leaves the current language in place instead of
   dropping the UI to raw message ids. Back online, switching to that same
   language works on the first try.
8. **Account → Offline data** — the size reported is the offline copy, not the
   whole origin (the ~9 MB app precache is excluded once a first Clear has
   established the floor). "Protected" chip when the browser granted persistent
   storage. Clear drops the counter to 0 straight away, removes the runtime
   caches, and leaves the app shell still working offline.

`yarn lint` — 0 errors (the 8 remaining warnings are pre-existing, in files this
branch does not touch).

## 📸 Previews

_To be added._


ClemRz (changes requested):
# Review

This is a substantial, well-thought-out PR. The offline UX strategy is sound — distinguishing network errors from server errors, auto-healing on reconnection, disabling (never hiding) network-dependent actions, and unifying the notification stack. The code is clean, the tests are thorough, and the inline documentation is unusually good. A few things to tighten up before landing.

---

### Issues (Must Fix)

- **[`packages/web-app/src/components/common/SectionCreateButton.jsx:32`]** Typo in the JSDoc comment: `"The tooltip i` (split mid-sentence — line wrapping artefact). While not a runtime bug, this sits in a public component's doc block and will mislead any future reader.
  ```suggestion
   * Hence a single `isBlocked` driving all three of the wrapper, the tooltip and
   * the button: they cannot drift apart. The tooltip is blanked rather than left
  ```

- **[`packages/web-app/src/pages/Account/index.jsx` — `EditActions`, `PersonalInfoSection`, `EmailSecuritySection`, `PreferencesSection`]** The Account page's "Save changes", "Change password", and "Save preferences" buttons are not guarded by `OfflineDisabled`. All three call `updateAccount()` (a PATCH) offline, which will fail silently after a spinner. The PR explicitly guards every other write action in the app via `OfflineDisabled` — this is the only section left out. Consider wrapping `EditActions`'s submit button (or the component itself) with `OfflineDisabled`, matching the pattern applied to login, compose, and the search forms.

- **[`packages/web-app/src/pages/Messages/ConversationDetail.jsx:719+`]** The `OfflineDisabled` wrapping the `InputRow` uses `fullWidth` so the wrapper doesn't collapse the row — that's correct. However, the `SendButton` inside the row is **not** given `disabled={!isOnline}`, so submitting via Enter key on the keyboard (`handleSend` via `onKeyDown`) is still reachable while offline regardless of the `OfflineDisabled` tooltip. The guard should also add `disabled={!isOnline}` to `SendButton`, consistent with how `ComposeDialog` treats its Send button.

---

- **[`packages/web-app/src/hooks/useSessionExpiry.js`]** The hook returns `expiresAt` in milliseconds (documented) but nothing in the codebase currently uses that field — `SessionExpiryNotifier` only reads `isExpiringSoon` and `userId`. If it's intended as a future API, a short `// TODO: expose to callers once the UI shows the exact time` would prevent it being treated as dead code. If it's genuinely not needed yet, consider not exporting it to avoid confusion.

- **[`packages/web-app/src/utils/networkError.js:42`]** The `navigator.onLine` fallback for unrecognised error shapes is broad: any error object that doesn't match the known shapes (e.g. a Redux-side string error from an unrelated middleware) will be treated as a network error when the device happens to be offline. The existing tests cover the documented shapes well, but this catch-all could produce false positives for edge cases. Worth a comment clarifying that this is intentional for the unrecognised shape, not a bug.

- **[`packages/web-app/vite.config.mjs` — `api-get` / 206-rewrite plugin`]** The `cacheWillUpdate` plugin that rewrites 206 → 200 is essential for pagination to work offline, and the comment explains the tradeoff well. One small follow-up: the `Content-Range` header is copied from the original 206, but the rewritten response has `status: 200` — if any caller checks `response.ok` or `response.status` before reading `Content-Range`, it will work, but worth a comment noting that this is deliberate (status 200 + Content-Range is non-standard but Cache-API-storable).

- **[`packages/web-app/src/components/appli/UpdatePrompt.jsx:57–62`]** `updateAction` is defined at module scope to avoid `react/no-unstable-nested-components`, but it creates a new factory function reference on every call. The snackbar key (`UPDATE_SNACKBAR_KEY`) prevents re-enqueue, so this is benign, but a note in the comment explaining why the factory returning a named function component is correct here (vs. a component defined at the outer scope) would help future maintainers avoid refactoring it back into the component body.

- **[`packages/web-app/src/pages/Map.jsx`]** `refetchMapViewport()` is called as a plain function (not dispatched), because `mapTileCache` holds its own dispatch reference. This is documented in `Map.js` but the call site in `Map.jsx` has no comment explaining why this non-thunk pattern is correct. Consider adding a brief inline note so the asymmetry between `loadMapData()` (dispatched) and `refetchMapViewport()` (direct call) is immediately clear to a reader.

- **[`packages/web-app/src/components/common/NetworkStatusNotifier.test.jsx:124`]** The test "translates the offline message when the catalogue lands after mount" passes `undefined` as `messages` to the tree helper, which triggers an `IntlProvider` with no messages object. The test currently asserts `countTexts(baseElement, 'offlineIndicator') === 1` before the messages land — this relies on `<FormattedMessage id="offlineIndicator" />` rendering the raw id as text when no messages are available. It's correct, but a comment would clarify this is intentional test infrastructure rather than an accidental missing prop.

---

### Nitpicks (Optional)

- **[`packages/web-app/src/components/common/FetchErrorState.jsx:33`]** The comment "Shares OfflineDisabled's label rather than carrying its own" is a bit cryptic — it means the same i18n key (`offlineActionUnavailable`) is reused as the Alert title. Worth expanding slightly: `// Reuses 'offlineActionUnavailable' as the title rather than a separate key — the two surfaces said the same thing in practice.`

- **[`packages/web-app/src/utils/offlineCache.js` — `rememberOfflineBaseline`]** The recursive async implementation (calling itself to sample multiple times) is creative, but iterative sampling with a loop + `sleep` would be easier to reason about stack depth on slow devices. Not a bug at the 8-default sample count, just worth noting.

- **[`packages/web-app/src/components/common/OfflineDisabled.jsx`]** The `disabled` prop defaults to `true` ("blocked whenever offline"), but this could silently wrap a toggle-style button that *should* remain usable (the `isBlocked = !isOnline && !isOpen` pattern documented in the JSDoc). The default is correct for the common case and well-documented — just reinforcing that callers *must* read the doc before omitting `disabled`.

ClemRz (approved):
# Follow-up Review

All three must-fix items from the previous review are addressed:
- `SectionCreateButton.jsx` JSDoc typo — fixed.
- `Account/index.jsx` save buttons — `OfflineDisabled` now wraps the `EditActions` submit across `PersonalInfoSection`, `EmailSecuritySection`, and `PreferencesSection`.
- `ConversationDetail.jsx` Enter-key path — `SendButton` carries `disabled={!isOnline}` and `RoundInput` is `disabled={!isOnline}`, so keyboard submission is fully blocked.

All suggestions are implemented too. The code quality is high, the inline documentation is genuinely useful, and the test coverage is thorough. A few small things to note below.

---

- **[`packages/web-app/src/components/common/Maps/MapClusters/index.jsx`]** `showOfflineDetailNotice` checks `visibleMarkers.length > 0` to guard against blaming the cache for a user-hidden map — that is correct and well-commented. One edge case worth verifying manually: the clustering threshold (`MARKERS_LIMIT`) is only relevant when `isMarkersMode` is true, and that flag is already part of the guard (`&& isMarkersMode`), so there is no issue there. Just flagging for a quick sanity-check in the test matrix that toggling all layers off offline does not trigger the notice.

- **[`packages/web-app/src/utils/offlineCache.js` — `rememberOfflineBaseline`]** The recursive sampling approach is creative and works at 8 samples. One forward-looking concern: if the function is called on a device that is already throttled or has a large precache, 8 × 300 ms = 2.4 s runs in the background and the browser may interrupt it on navigation. It is fire-and-forget so there is no visible failure, but consider adding a `try/catch` around the recursive body in case `getStorageUsage` throws on a navigated-away context where storage is no longer accessible. Currently an unhandled rejection in the recursive tail would be swallowed by the `if (usage == null) return` guard, so it is already resilient — this is just belt-and-braces.

- **[`packages/web-app/src/pages/Messages/ConversationDetail.jsx` — `reloadMessages` dependency]** `reloadMessages` is defined with `useCallback` depending on `[dispatch, conversationId]`, then used both in a `useEffect` (replacing the old one) and in `useRefetchOnReconnect`. The switch from `dispatch(...)` directly inside the effect to a `useCallback`-wrapped loader is the right pattern. Worth a quick look: if `conversationId` changes (user navigates to a different thread), `reloadMessages` changes reference → the effect fires → correct. `useRefetchOnReconnect` also picks up the new reference via its ref sync. No issue found; noting it because the pattern is less common and worth confirming in review.

- **[`packages/web-app/src/components/common/NetworkStatusNotifier.jsx`]** The `hasBeenOffline` ref is reset to `false` on every "back online" path. That means if the app transitions `online → offline → online` in a single mounted lifecycle, the "back online" snackbar fires; but a subsequent `online` event (no intervening `offline`) is silently ignored. This is correct behaviour — just noting for reviewers that the guard is intentional, as it could look like a bug.

---

### Nitpicks (Optional)

- **[`packages/web-app/src/components/common/FetchErrorState.jsx:32`]** The comment "Reuses the i18n key `offlineActionUnavailable` … the two surfaces said the same thing in every language but English" could be read as "the English translations are different" when what is meant is "the key now covers both". Minor wording, no impact.

- **[`packages/web-app/src/pages/Messages/index.jsx:394`]** The comment on the mobile `Fab` states "The Fab stays absolutely positioned against the relative parent: OfflineDisabled's span is static, so it never becomes the containing block." This is accurate but slightly non-obvious — the `OfflineDisabled` span has `display: inline-flex` and no `position`, so it is indeed out of the positioning flow. The comment explains the correct reasoning; just a note that this is load-bearing and should not be changed to `display: block` in a future refactor without retesting the FAB position.

- **[`packages/web-app/vite.config.mjs`]** The `cacheKeyWillBeUsed` hook that strips the `zoom` param from map tile cache keys is well-documented. The one thing not stated: if `mapTileCache.js` ever stops sending `zoom`, the hook becomes a no-op rather than breaking anything, so the coupling is one-directional and safe. Worth a note in the comment for future readers.

PR #1503, opened 2026-08-09
feat(ux): rework navigation, tables, documents and administration experience

### LINKED ISSUE #1304: Show notification indicators for documents pending validation
## Is your feature request related to a problem? Please describe.

Moderators have no visual indication that documents are awaiting validation. They must manually navigate to the document validation page to discover pending submissions. This makes it easy to overlook new documents that need review.

## Describe the solution you'd like

When a moderator is logged in and there are documents pending validation, display notification indicators in three places:

1. **User avatar/initials** (top bar): a small dot overlay indicating "something needs attention."
2. **Dashboard link** (in the user dropdown menu): a dot next to the link.
3. **Document Validation button/link** (on the dashboard): a badge showing the exact count of pending documents (e.g., `3`).

### Behavior

- On login (or when navigating to the dashboard), call the new API endpoint `GET /api/v1/documents/count-pending-validation` (see companion API issue).
- If `count > 0`, show the dot on the avatar and dashboard link, and the numeric badge on the Document Validation button.
- If `count === 0`, hide all indicators.
- Only fetch and display for users with the **Moderator** or **Administrator** role.
- No real-time polling needed — fetching on login and on dashboard navigation is sufficient.

### Visual reference

- **Avatar dot**: small colored circle (e.g., red or accent color) positioned at the top-right corner of the avatar/initials, similar to standard notification badge patterns.
- **Dashboard link dot**: same small dot next to the "Dashboard" text in the dropdown menu.
- **Document Validation badge**: numeric badge (e.g., pill-shaped) showing the count, placed next to or on the Document Validation button/link.

## Describe alternatives you've considered

- Showing a badge with count everywhere. Rejected in favor of progressive disclosure: dot for general awareness, count only where the user takes action.
- Real-time polling or WebSocket updates. Overkill for this use case — document submissions are infrequent enough that fetch-on-navigation is sufficient.

## Additional context

- Depends on the companion API issue for `GET /api/v1/documents/count-pending-validation` on `GrottoCenter/grottocenter-api`.
- The indicators should only be visible to Moderators and Administrators, matching the permission model of the validation endpoint.

Paul-AUB:
# feat(ux): rework navigation, tables, documents and administration experience

A broad UX pass over the application: navigation (desktop + mobile), search
tables, the documents section, and the administration area. 26 commits, mostly
independent and reviewable one by one.

Closes #1493 · Closes #1375 · Closes #1304

### 🧭 Navigation & menus

- **Desktop side menu is now foldable**, and keeps its menu buttons always
  accessible when collapsed.

  image
  image

- **Mobile side menu** follows the gesture fluidly again (it used to jump, fixed).
- **Quick search moved into the app bar on mobile** — reachable in one click
  instead of opening a dedicated screen.
  
  
image
image

  
- **Simpler menu for regular users**: notifications are directly accessible, and
  the old dashboard is now reserved for governance/admin profiles.
  
  
image
  
- Feedback button no longer covers the footer.

image

### 📊 Search tables

- **All table controls moved to the top, sticky and always visible** while
  scrolling (closes #1375).
- Sticky headers render correctly when scrolling (no more overlap/jitter).

image

### 📄 Documents

- **Sort option for document lists** on entrances, caves and massifs
  (closes #1493).

- Images and files take much less vertical space in document sections;
  thumbnails are denser and more readable.

image

- **PDF previews now work in the PWA/app** (they silently failed before).
- Lightbox: **reset-zoom button**, and the caption shows the document title
  instead of the raw file name.

image

- Document type no longer overlaps other UI elements.

image

- Publication authorization is shown with an icon.

### 🛡️ Moderation & administration

- **Indicator on the moderator avatar** when an action is required
  (pending documents, duplicates) — closes #1304.

image

- **Administration/management tools page reworked.**
- Responsible-organization form is now inline in the page, consistent with the
  other forms, and surfaces its error messages properly.

image

### 📱 Mobile & PWA

- Buttons collapse to **icons + tooltips on mobile** to save space, with burger
  menus for overflow actions.

image

- App resize and pull-to-refresh gestures are prevented in the PWA.

### 🎨 Visual consistency

- **Destructive actions** are consistently identified with red color + icon.

image

- "Set as visited" is now a **split button**, far more visible.

image

- Title boldness slightly reduced.

image

- All occurrences of `GrottoCenter` renamed to `Grottocenter`.

### 📚 Docs

- README reworked for new contributors and better project presentation.

## 🤷‍♂️ Why

Several usability issues had accumulated, especially on mobile and on the
long-scrolling pages:

- **Table controls scrolled out of view**, so on a long result list you had to
  scroll back to the top to change anything (#1375).
- **Document lists had no ordering**, which made them hard to browse once an
  entrance or a massif accumulated many files (#1493).
- **Moderators had no way to know** that something was waiting for them without
  opening the dashboard (#1304).
- **Mobile screens wasted a lot of space** on text buttons and on a quick search
  that took two interactions to reach.
- The **dashboard mixed regular-user concerns with governance tools**, which made
  it confusing for the large majority of users who only needed notifications.
- Destructive actions looked exactly like neutral ones — a real risk on a
  collaborative database.

### New shared building blocks

| File | Purpose |
| --- | --- |
| `hooks/useIsDesktopLayout.js` | Single source of truth for the desktop/mobile switch |
| `hooks/useMeasuredWidth.js` | Measure elements to drive responsive collapsing |
| `hooks/useSideMenuOffset.js` | Share the foldable side-menu offset with the layout |
| `components/common/AppBar/MobileSearch.jsx` | In-app-bar quick search on mobile |
| `components/common/EntityTable/ToolbarActionButton.jsx` | Unified sticky-toolbar action button |
| `components/common/DocumentSortSelect.jsx` | Document list sort control |
| `components/common/PdfPreview/index.jsx` | PDF rendering that works in the PWA |
| `utils/pdfViewerSupport.js` | Feature detection for native PDF embedding |
| `utils/documentSort.js` | Generalized sorting (replaces `documentChildrenSort.js`) |

### Notable points

- **Sticky toolbars** build on the existing `FixedContent` scroller: the toolbar
  and the table header stack via measured heights rather than hardcoded offsets,
  so they stay correct whatever the toolbar wraps to.
- **PDF preview in the PWA**: the standalone display mode does not honor the
  native PDF plugin, so `pdfViewerSupport.js` detects it and falls back to an
  explicit renderer instead of an empty frame.
- **Moderator indicator** is backed by two new Redux slices,
  `PendingDocumentsCount` and `DuplicatesCount`, following the project's
  action/reducer convention.
- **`documentChildrenSort` → `documentSort`**: the old helper only handled
  document children; it was generalized and its tests extended rather than
  duplicated.
- **`Entry/Science` and `Network/Science` were merged** into a single
  `appli/Science` component — they had drifted into near-duplicates.
- Responsive action bars go through `Layouts/ResponsiveActions.jsx`, which
  collapses labels to tooltips and overflows into a burger menu.

## 🔗 Related API PR

_None — this PR is frontend-only and needs no API change._

## 🧪 Testing

`yarn lint` → **0 errors**. `yarn test` → **967 tests / 105 files, all passing**
(including new `documentSort` and `AppLink` suites, and an extended
`DesktopEntityTable` suite).

Worth exercising manually:

1. **Desktop** — fold/unfold the side menu, check the menu buttons stay reachable
   and the content reflows.
2. **Mobile** — open the quick search from the app bar; swipe the side menu open
   and closed and check it tracks the finger.
3. **Search tables** — scroll a long result list and confirm the controls and the
   header stay pinned.
4. **Documents** — sort a document list on an entrance/cave/massif; open an image
   in the lightbox and use reset-zoom; check the caption shows the title.
5. **PWA** — install the app, open a document with a PDF, confirm it renders;
   check pull-to-refresh no longer triggers.
6. **Moderator account** — confirm the avatar indicator appears when documents
   are pending validation, and disappears once handled.
7. **Regular user account** — confirm the simplified menu and that the dashboard
   is not offered.

## 📸 Previews

_To be added._


ClemRz (changes requested):
### Issues (Must Fix)

- **[README.md:24]** The new Getting Started section references `REACT_APP_API_URL`, which is a dead CRA-era variable. Client code only reads `VITE_`-prefixed vars (see `architecture.md`). New contributors following this doc will edit the wrong key and see no effect.
  ```suggestion
  By default the app uses the production API as backend. To change it, edit the `VITE_API_URL` variable in `packages/web-app/.env`.
  ```

---

- **[packages/web-app/src/components/common/DocumentsList/ImageLightbox.jsx:198]** The wheel-zoom floor is `0.5`, meaning the user can zoom the image down to half size inside the lightbox, below the "fit to frame" baseline. `PdfJsViewer` has an explicit `MIN_ZOOM = 1` for the same reason. Consider aligning to `1` as the floor so the image can never become smaller than the viewport it fills — or add a comment explaining why sub-1 zoom is intentional here.

- **[packages/web-app/src/components/common/AppBar/User.jsx:68–70]** `fetchPendingDocumentsCount` fetches **all** unvalidated documents (no role filter on the frontend), relying on the API to restrict the response to moderators. The issue spec (#1304) requests showing the indicator for Moderators *and* Administrators. The current `permissions.isModerator` guard would silently skip Administrators. Consider `permissions.isModerator || permissions.isAdmin`.

- **[packages/web-app/src/actions/PendingDocumentsCount.js]** The action queries `getDocumentsUrl` with `{ isValidated: false }`, which fetches the general unvalidated document list. The companion API issue (#1560) describes a dedicated `GET /api/v1/documents/count-pending-validation` endpoint. The PR description acknowledges this is a temporary workaround using existing APIs. Worth a `// TODO: replace with /count-pending-validation when the API endpoint lands` comment so the technical debt is traceable.

- **[packages/web-app/public/lang/*.json (ar, bg, ca, de, el, id, it, ja, nl, pt, ro)]** The five new administration description keys (`CSV Import description`, `Document validation description`, `Duplicates tool description`, `Manage users description`, `Management tools access restricted`) are present in all language files but left with their English values in every locale except `fr`. `fr` is fully translated. This is consistent with the project's translation workflow (AI-translate after merge), but it's worth ensuring the translation workflow runs before the first production release — these strings appear in a gated page (management tools), so the impact is limited.

- **[packages/web-app/src/components/common/AppBar/User.jsx:62–65]** `bothCountsSucceeded` requires both counts to have resolved before `showPendingDot` can ever be true. If one of the two fetches fails (network blip, API error), the dot is permanently suppressed for the whole session. Consider making the condition additive: show the dot if *at least one* count succeeded and has a non-zero value.

### Nitpicks (Optional)

- **[packages/web-app/AGENTS.md]** The `packages/web-app/README.md` is mentioned in the root AGENTS file as a reference for the web-app, but that README still references `REACT_APP_API_URL` (unchanged in this PR). Worth a follow-up to keep the two in sync.

- **[packages/web-app/src/components/common/EntityTable/DesktopEntityTable.jsx:84–89]** `getScrollParent` walks up the DOM and reads a computed style per node. For a typical card layout this is a handful of iterations, but it runs on every page change and sort. Not a real concern at current scale, but caching the result in a `useRef` (resetting on mount) would be cleaner than rediscovering it on every navigation.

- **[packages/web-app/src/components/common/PdfPreview/index.jsx]** `hasNativePdfViewer` is called as a `useState` initializer: `useState(hasNativePdfViewer)` — note no parentheses. This is the correct lazy-initializer pattern and works, but it differs from the documented behavior elsewhere in the file where the comment says "Read once per mount". The comment is accurate, but `useState(() => hasNativePdfViewer())` would be the more conventional spelling of a lazy initializer; passing the function reference directly also works because `hasNativePdfViewer` takes no arguments. No change needed — just noting it is unconventional.

- **[packages/web-app/src/components/common/DocumentsList/DocumentsList.jsx:sort reset]** When `sortOrder` changes, `page` is reset to 1 (correct). When `documents` changes (new entity loaded), `page` is NOT reset. If a user views page 3 of entity A, navigates to entity B which has fewer total pages, they land on a blank page 3. Consider a `useEffect` that resets `page` when `sortedDocuments` changes identity (i.e. when `documents` prop changes), or use a key on the component.

ClemRz (commented):
### Follow-up review — after de6d4604

All **Issues (Must Fix)** from the first review are resolved. Checking the updated code:

---

**`PendingDocumentsCount` TODO** ✅ — the comment now points at `GET /api/v1/documents/count-pending-validation` and the API issue. Good.

**Additive dot logic** ✅ — `countOf` is per-source and the two are summed, so one failing fetch no longer suppresses the other. The logic is clean.

**`ImageLightbox` zoom floor** ✅ — `MIN_ZOOM = 1` / `MAX_ZOOM = 5` are module-level constants, consistent with `PdfPreview`. All four zoom paths (wheel, pinch, double-tap, reset) go through them.

**Untranslated administration keys** ✅ — all 13 language files now carry proper translations for the five new keys (including `es` and `he` which were missed in the first pass).

**`DocumentsList` page clamp** ✅ — `Math.min(page, Math.max(totalPages, 1))` clamps without resetting, which also handles the "unlink the last document on the last page" case.

---

- **[`packages/web-app/src/components/common/AppBar/User.jsx:62`]** The pending dot is still gated on `permissions.isModerator` only, while the "Management tools" menu item is shown to `isAdmin || isModerator || isLeader`. An administrator who opens the menu and sees the dot next to "Management tools" gains useful signal; a Leader whose duplicate dashboard card is hidden gains nothing. The deliberate decision to exclude admins is documented in the thread, but it might be worth calling that asymmetry out in an inline comment for the next reader — it looks like a bug otherwise.

- **[`packages/web-app/src/components/common/EntityTable/DesktopEntityTable.jsx:328–333`]** The mount-only `useEffect` that calls `onPageChange` to sync the stored `rowsPerPage` with the parent carries a bare `// eslint-disable-next-line react-hooks/exhaustive-deps` with no written justification. This was pre-existing on `develop`, but this PR touches the surrounding file extensively — worth adding "intentional mount-only sync; adding deps would loop" or equivalent while in the neighbourhood.

### Nitpicks (Optional)

- **[`packages/web-app/src/components/common/PdfPreview/index.jsx:305`]** `useState(hasNativePdfViewer)` passes the function reference as a lazy initializer — documented in the first review as unconventional but correct. Paul chose to leave it, which is fine. The comment in the file explains the intent.

- **[`packages/web-app/src/utils/documentSort.js`]** `DEFAULT_DOCUMENT_SORT_ORDER` is exported but `DocumentsList` deliberately ignores it and uses `ADDED_DESC` instead (with a comment explaining why). Consider either removing the export or renaming it to `DEFAULT_COLLECTION_SORT_ORDER` to make the distinction explicit.

ClemRz (changes requested):
### Follow-up review — after c555518d

All **Issues** and **Suggestions** from the previous two rounds are now addressed cleanly. This pass covers the three items landed in c555518d (the `User.jsx` comment, the `DesktopEntityTable` eslint-disable justification, the `DEFAULT_COLLECTION_SORT_ORDER` rename) plus a fresh read of the remaining code.

---

### Issues (Must Fix)

- **[`AGENTS.md:14-15` and `README.md:3-4`]** Several GitHub repository URLs were changed from `github.com/GrottoCenter/` to `github.com/Grottocenter/` (lower-case `c`). GitHub is case-insensitive for web browsing, but **`package.json:8`** was also updated to `git@github.com:Grottocenter/grottocenter-front.git` — Git remote URIs **are** case-sensitive on the server side, and the actual GitHub org slug is `GrottoCenter` (capital C). A `git push` using this SSH remote will fail with "Repository not found" on a fresh clone. The badge URLs in `README.md` use `Grottocenter` too, so the CI badge images will 404 until the org is renamed.
  ```suggestion
  "repository": "git@github.com:GrottoCenter/grottocenter-front.git",
  ```
  The prose rename (`GrottoCenter` → `Grottocenter`) in display text, comments, and translation values is correct and intentional. Only the **URL paths** that reference the org need to stay `GrottoCenter`.

---

- **[`packages/web-app/src/components/common/AppBar/User.jsx:24`]** The two new action imports (`fetchDuplicatesCount`, `fetchPendingDocumentsCount`) and the three hook imports still use three-segment relative paths (`../../../actions/…`, `../../../reducers/…`, `../../../hooks`) while the file already uses `@/` for `UserAvatar`. The `AGENTS.md` convention says all new code must use `@/` for paths that would otherwise require two or more `../` segments. These new imports added in this PR are exactly that case.
  ```suggestion
  import { fetchDuplicatesCount } from '@/actions/DuplicatesCount';
  import { fetchPendingDocumentsCount } from '@/actions/PendingDocumentsCount';
  import REDUCER_STATUS from '@/reducers/ReducerStatus';
  import { usePermissions, useUserProperties } from '@/hooks';
  ```

- **[`packages/web-app/src/components/common/EntityTable/DesktopEntityTable.jsx:328–333`]** The justification comment added in this round (`Mount-only: syncs the parent to the rowsPerPage restored from localStorage…`) is clear and correct — the eslint-disable is now explained. One minor nit: the `eslint-disable-next-line` sits after the comment rather than directly above the offending line. In most codebases this still works (the rule fires on the `useEffect` call on the next line), but the canonical placement is the line immediately before the flagged code, not before the comment that explains it. Not a lint error, just a readability preference.

- **[`packages/web-app/src/utils/documentSort.js:17`]** `DEFAULT_COLLECTION_SORT_ORDER` is exported but `DocumentsList.jsx` deliberately ignores it and defaults to `ADDED_DESC` (with an explanatory comment). The export is now correctly named and the comment in `documentSort.js` explains the intended consumer (collection children). Consider also adding a brief note in the export's JSDoc — or at the call site in `DocumentDetails/index.jsx` where it is used — confirming that `DEFAULT_COLLECTION_SORT_ORDER` is `PUBLICATION_DESC` so the reader does not have to trace the constant. This is cosmetic and purely optional given the existing comments.

- **[`packages/web-app/src/components/common/AppBar/index.jsx:19`]** `usePermissions` is imported from `../../../hooks` but is not used anywhere in `AppBar/index.jsx` (permissions are used in `User.jsx`, not here). This is a dead import that will cause a lint warning if the `no-unused-vars` rule catches it. Worth removing.

---

### Nitpicks (Optional)

- **[`packages/web-app/src/components/common/DocumentsList/DocumentsList.jsx:97–99`]** The page clamping (`Math.min(page, Math.max(totalPages, 1))`) is correct and the comment explaining it is exactly right. One micro-note: `page` here is React state (1-based) while `currentPage` is the clamped version used for rendering. Since `setPage` is never called with the clamped value, navigating away and back will reset the page because `page` state was never updated to match `currentPage`. This is the correct and intentional behaviour (as noted in the PR description), but a reader unfamiliar with the pattern may wonder why `setPage(currentPage)` isn't called — a one-line comment ("state left intentionally unclamped; next navigation or entity load resets it naturally") would pre-empt that question.

- **[`packages/web-app/src/hooks/useMeasuredWidth.js`]** The hook is a near-exact mirror of `useMeasuredHeight.js`. Consider whether a shared `useMeasuredDimension(dimension: 'width' | 'height')` would be cleaner, or whether the current two-file approach is intentional for readability. Not blocking — the duplication is minimal and both hooks are well-documented.

- **[`packages/web-app/src/components/common/AppBar/MobileSearch.jsx:95–104`]** The `focusout`/`onBlur` dismiss is well thought-out and the comment explaining why `focus` beats a click-away listener is valuable. Worth confirming (if not already done) that the same `event.relatedTarget` check works when a user taps the back button on Android (which may not trigger `focusout` in all WebViews). Low risk for the PWA given the `isSearchOpen → false` guard in the `isDesktop` effect, but worth a note in the TODO if you run into it.

ClemRz (changes requested):
### Follow-up review — after c5875695

All **Issues** and **Suggestions** from the three previous rounds are resolved. This pass covers the changes since `c555518d` (the `AGENTS.md`/`README.md`/`package.json`/`User.jsx`/`DocumentsList.jsx`/`documentSort.js` commit) and a fresh read of the full diff.

---

### Issues (Must Fix)

- **[`packages/web-app/src/conf/externalLinks.js:30`]** `githubLink` was changed to `'https://github.com/Grottocenter'` (lowercase `c`). This is a URL path, not display text — GitHub org slugs are case-sensitive on the server side, and the actual org is `GrottoCenter`. The same concern was raised and fixed for `package.json` in the previous round; `externalLinks.js` was not in scope then but was touched now. Any deep-link that hands this URL to a machine (OAuth redirect, API call) will resolve to the wrong target on a case-sensitive host.
  ```suggestion
  '*': 'https://github.com/GrottoCenter'
  ```

- **[`packages/web-app/public/lang/_README.md:26`]** The wiki link was changed to `https://github.com/Grottocenter/grottocenter-front/wiki/…` — same lowercase-`c` issue. `_README.md` is the translation contributor guide; a wrong link there sends contributors to a 404 (or a case-folded redirect that may not resolve on all hosting configurations).
  ```suggestion
  See [our dedicated wiki page](https://github.com/GrottoCenter/grottocenter-front/wiki/Translation-workflow) for more information.
  ```

---

- **[`packages/web-app/src/components/appli/EntitiesForm/Document/formElements/SearchBar.jsx:152`]** The inline comment URL was also changed to `https://github.com/Grottocenter/grottocenter-front/issues/58`. Comments and issue URLs are not machine-executed, so this is not blocking — but keeping all three in sync (code URL, README, package.json) avoids confusion about which casing is canonical.

- **[`packages/web-app/src/components/appli/OrganizationAssociation/AssociationForm.jsx:55–62`]** The info alert that warns "a new organization will be created" now fires when `!hasExactMatch` rather than when `results.length === 0`. Because the quick-search is fuzzy and almost always returns partial matches, the old condition almost never showed the alert even when the user was clearly about to create something new. The new condition is a real improvement. One edge case worth being aware of: a user who types `"GUCEM"` when the org is stored as `"gucem"` will see `hasExactMatch = true` (case-insensitive compare) and the alert is suppressed — yet the submit will create a duplicate rather than link the existing one, because `selectedOrg` is only set by clicking a list item, not by the exact-match. The submit path needs the user to *select* the item; typing the exact name is not enough. This is existing behaviour (the PR did not regress it), but since the alert was specifically meant to guide users away from accidental creation, it is worth a comment.

- **[`packages/web-app/src/components/common/AppBar/User.jsx:68–70`]** The fetch fires only when `permissions.isModerator` is true, but the dot depends on `isModerator` too — so the count reducers always start at `IDLE` for a non-moderator. The `countOf` helper short-circuits on anything that isn't `SUCCEEDED`, so the dot correctly stays hidden. No bug here, but the `countOf` helper and `hasPendingTasks` are evaluated even when `permissions.isModerator` is false (rendering the `showPendingDot = false` calculation redundant). Consider early-returning `const showPendingDot = false` for non-moderators to make the intent explicit, or moving the `useEffect` dependency guard up so the derived values are clearly unreachable.

---

### Nitpicks (Optional)

- **[`packages/web-app/src/components/common/AppLink.jsx:55–60`]** The `import.meta.env.DEV` dev-mode check works in Vite and Vitest (which sets the same env), so the runtime guard is fine. Worth a quick note that this is a Vite-specific idiom rather than `process.env.NODE_ENV === 'development'`, so a future non-Vite consumer would not get the warnings. Already implied by the codebase's full Vite adoption, but worth calling out for the next contributor who maintains this file.

- **[`packages/web-app/src/components/common/DocumentsList/DocumentsList.jsx:159–160`]** The `isWide` flag is derived per-document in the same `useMemo` that computes `allImages`, which is correct — sorting changes the order and the gallery threshold must follow the sorted order, not the original one. The comment `// A gallery earns the full row: past the threshold its tiles no longer fit one card` is clear. One observation: the full-width span uses `gridColumn: '1 / -1'`, which spans all *explicit* columns. In a single-column layout (mobile) this is a no-op as noted, but in a two-column layout a document with exactly `GALLERY_MIN_IMAGES` images will span both columns even if the adjacent column still has room for a shorter card. That is the intended design ("earn the full row"), just flagging it is an explicit trade-off, not an accident.

- **[`.kiro/steering/`]** The `architecture.md` and `product.md` steering files still say `GrottoCenter` (capital C) in several places. If the rename is intentional project-wide, these should be updated too — but steering files are usually updated in a dedicated tech PR rather than alongside a feature.

ClemRz (approved):
### Follow-up review — fourth round

All **Issues** and **Suggestions** from the previous three rounds are resolved. This pass covers the full diff after the latest commit.

---

### Nitpicks (Optional)

- **[public/lang/en.json (and all other lang files)]** The old capitalized pagination keys `"Next Page"`, `"Previous Page"` are still present in every language file, but nothing in the source uses them anymore — the new `"Next page"` / `"Previous page"` (lowercase `p`) took over in `DesktopEntityTable` and `PdfPreview`. They are inert dead weight that `yarn translations:find-unused` would surface. Worth removing in a follow-up (or right now, since the lang files are already touched in this PR).

---

Everything else looks solid. The new architecture is well thought-out:

- **New hooks** (`useIsDesktopLayout`, `useMeasuredWidth`, `useMeasuredHeight`, `useSideMenuOffset`) are clean, well-documented, and genuinely DRY — they address the root cause of the tablet layout drift rather than patching each symptom.
- **`AppLink` value-based routing** (deciding internal vs. external from the URL value, not the prop name) correctly fixes the side-menu-as-new-tab regression, and the regression tests for it are convincing.
- **`PdfPreview`** cancellation logic is thorough — both the `isCancelled` flag guard and `loadingTask?.destroy()` in the cleanup are correct. The `useState(hasNativePdfViewer)` lazy-initializer pattern is intentional and correct.
- **`documentSort`** handles the full edge case set (undated documents, locale-aware collation with fallback, unknown order) and the test coverage matches.
- **`DuplicatesCount` / `PendingDocumentsCount` reducers** follow the project pattern exactly and are registered in `GCReducer`. The `showPendingDot` narrowing to moderator-only is a good UX call.
- **Permission check** for the "Management tools" menu item is intact — all three roles (admin, moderator, leader) still reach the dashboard, matching `Dashboard.jsx`'s own guard.
- **`FeedbackButton` footer detection** via `IntersectionObserver` is a clean way to avoid the footer overlap without hardcoding heights. `tabIndex={-1}` when hidden is a good a11y touch.

Good work on the PR. 🎉

PR #1504, opened 2026-08-09
fix(twa): edge-to-edge readiness, R8 optimization and AGP 9 migration

Paul-AUB:
## 🤔 What

Clears the three warnings the Play Console raised on the published bundle, and brings the Android build up to date.

- `androidbrowserhelper` 2.6.2 → **2.7.2**, which makes the splash screen edge-to-edge
- **minSdk 21 → 23** (Android 6.0) — forced by the above, see below
- R8: adds `shrinkResources`, `proguardFiles`, `proguard-rules.pro`, explicit full mode
- **AGP 8.9.1 → 9.3.1**, **Gradle 8.11.1 → 9.7.0**, and the AGP 9 DSL migration
- Drops `jcenter()`, shut down since 2022
- `res/raw/keep.xml` — new, see the resource-shrinking note below
- `generateShorcutsFile` no longer rewrites a tracked source file on every build

**APK −53% (1.41 MB → 654 KB), AAB −52% (1.54 MB → 747 KB)** versus `develop`.

## 🤷‍♂️ Why

Play Console reported, on the published bundle:

1. Deprecated edge-to-edge window APIs under Android 15 — `Window.setStatusBarColor`, `setNavigationBarColor`, `getStatusBarColor`
2. Resource shrinking not enabled
3. R8 optimization not enabled, and AGP below 9.0

**On (1) — none of it was our code.** All three call sites Google listed live inside `androidbrowserhelper` 2.6.2 (June 2023). Our Android code is three trivial classes and touches no window API, so there was nothing for `enableEdgeToEdge()` to be added to. Version 2.7.2 (June 2026) paints the splash screen through androidx `ProtectionLayout`/`ColorProtection` and opts in via `WindowCompat.enableEdgeToEdge()` — which *is* the `enableEdgeToEdge()` Google's message asks for, just inside the library.

**On (3) — `minifyEnabled true` was set, but there was no `proguardFiles` block at all.** Without one AGP injects no default configuration, so R8 ran on the AAPT-generated rules only and its optimizer never engaged. That is what the Console reports as "optimization is not enabled".

### The dependency bump forces minSdk 23

`androidbrowserhelper` 2.7.x declares `minSdk 23` and the manifest merger refuses to go below it. The only alternative Google offers is `tools:overrideLibrary`, which its own message describes as "may lead to runtime failures" — it trades a build error for a crash on Android 5.x. So the floor moves to Android 6.0.

⚠️ **Worth a look at the Play Console install base before merging.** Android 5.0/5.1 should be a rounding error in 2026, and a TWA needs a recent Chrome anyway, but the number is yours to check.

### What is actually left in the release DEX

Disassembled the shipped APK to confirm the fix rather than trusting the changelog. Two call sites remain, both benign:

| Call site | Status |
| --- | --- |
| `LauncherActivity.a()` | androidx `WindowCompat.enableEdgeToEdge()`, inlined by R8 — the documented pre-API-35 compat path, guarded on `SDK_INT` |
| `WebViewFallbackActivity.onCreate` | guarded to `SDK_INT <= 34`, and unreachable anyway with `fallbackType: customtabs` |

The two `Utils.set*BarColor` helpers Play named are **gone** from the DEX: nothing in 2.7.2 calls them any more, so R8 strips them. `WebViewFallbackActivity` will probably still show up in the Console's report — the analysis is static and sees the guarded call — but it never executes.

### AGP 9

`buildFeatures { resValues = true }` is the load-bearing line. AGP 9 flips `android.defaults.buildfeatures.resvalues` to `false`, and this project routes practically everything it displays through `resValue` — app name, launch URL, system bar colors, provider authority. **Without the opt-in the build still succeeds** and produces an app whose resources resolve to nothing.

Also migrated: `compileSdkVersion`/`minSdkVersion`/`targetSdkVersion` → `compileSdk`/`minSdk`/`targetSdk`, `lintOptions` → `lint`, `task clean` → `tasks.register`, and the `package=` attribute removed from `AndroidManifest.xml` (AGP 9 rejects it rather than ignoring it). `versionCode`/`versionName` were deliberately left in space-assignment syntax so the CI `sed` in `twa-build.yml` keeps matching. Zero Gradle deprecation warnings remain.

### One real regression, caught and fixed

AGP 9 enables optimized resource shrinking, which traces resources and classes together — worth another 37% off the APK on its own. It also silently removed `string/update_chrome_toast`: `androidbrowserhelper` resolves that one through `Resources.getIdentifier()`, by name, so no compiled reference points at it. The out-of-date-Chrome toast simply stopped appearing (`ChromeUpdatePrompt` treats a missing resource as "skip the toast", so no crash). `res/raw/keep.xml` restores it.

`string/fullScopeUrl` is also shrunk away, in this PR and already before it — it is read only by Meta Quest builds and `isMetaQuest` is `false`.

### Build hygiene

`generateShorcutsFile` was generated by Bubblewrap as a bare `task { … }`, so its body ran during Gradle's configuration phase and rewrote the tracked `shortcuts.xml` on every invocation — `./gradlew tasks` included, stripping its license header each time. It is now registered with the work in `doLast` and emits the header, so generated and committed match and builds leave the working tree clean.

### Keeping this from being undone

Bubblewrap 1.25.0 — the latest — still pins `androidbrowserhelper:2.6.2` and still generates an AGP 8 project with the unoptimized R8 setup. A `bubblewrap update` would silently revert all of this and bring the warnings back. `twa/README.md` now lists everything that has to survive a regeneration.

## 🧪 Testing

Verified locally with a full signed release build (AGP 9.3.1 / Gradle 9.7.0 / JDK 17), throwaway keystore:

- `./gradlew clean :app:bundleRelease :app:assembleRelease` — green, no deprecation warnings
- Packaged manifest: `minSdkVersion 23`, `targetSdkVersion 36`, launchable activity and label intact
- Resource table: every `resValue` resolves — `bool/enableNotification` = true, `integer/splashScreenFadeOutDuration` = 300, `color/backgroundColor` = `#ff5d4037`; splash, launcher and notification icons present
- DEX: all three app classes kept; the only `Window` bar-colour call sites are the two listed above

**Still needed before publishing** — I could not do these here:

- [ ] Install the release APK on a **physical Android 15 or 16 device** and check the splash screen draws correctly behind the system bars, then the TWA opens with no Chrome URL bar
- [ ] Confirm notification delegation and the geolocation permission prompt still work — R8 full mode is newly effective here and `androidbrowserhelper` ships no consumer ProGuard rules of its own
- [ ] Check the Android 5.x install base in the Play Console against the `minSdk` bump

## 📸 Previews

Size comparison, both built from a clean tree with the same throwaway keystore:

| Artifact | `develop` | This branch | Δ |
| --- | --- | --- | --- |
| APK | 1 409 143 B | 656 814 B | −53.4% |
| AAB | 1 539 671 B | 746 686 B | −51.5% |

### DIFF
--- a/twa/README.md
+++ b/twa/README.md
@@ -314,6 +314,31 @@ bubblewrap update            # apply twa-manifest.json changes to the project
 > Bubblewrap version emits. Google Play requires the target API to be at most one
 > year behind the latest Android release (**API 36 minimum from 2026-08-31**), so
 > check that line after every regeneration.
+>
+> ⚠️ The same goes for everything below — Bubblewrap 1.25 still generates an AGP 8
+> project with an unoptimized R8 setup, so a regeneration silently undoes all of it
+> and the Play Console warnings come back:
+>
+> In `app/build.gradle`:
+>
+> - `androidbrowserhelper:2.7.2` — Bubblewrap still pins 2.6.2, and 2.7.x is what
+>   makes the splash screen edge-to-edge
+> - `minSdk 23` — 2.7.x will not merge below it
+> - `shrinkResources`, `proguardFiles …-optimize.txt`, `proguard-rules.pro`
+> - AGP 9 DSL — `compileSdk`/`minSdk`/`targetSdk`, `lint {}`,
+>   `buildFeatures { resValues = true }`
+>
+> Elsewhere:
+>
+> - `build.gradle` — `com.android.tools.build:gradle:9.3.1`, `mavenCentral()`
+>   instead of the shut-down `jcenter()`
+> - `gradle.properties` — `android.enableR8.fullMode=true`
+> - `app/src/main/res/raw/keep.xml` — a new file Bubblewrap never generates
+> - `app/src/main/AndroidManifest.xml` — no `package=` attribute (AGP 9 rejects it)
+>
+> `buildFeatures { resValues = true }` is the dangerous one: without it AGP 9 builds
+> an app whose every generated string and color resolves to nothing, and the build
+> still succeeds.
 
 Then **commit the regenerated project** so CI builds the updated app:
 
--- a/twa/README.md
+++ b/twa/README.md
@@ -300,14 +300,19 @@ bubblewrap update            # apply twa-manifest.json changes to the project
 >     }
 >     buildTypes {
 >         release {
->             minifyEnabled true
->             signingConfig signingConfigs.release   // ← add this line
+>             minifyEnabled true                     // ← Bubblewrap emits only this
+>             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+>             shrinkResources = true
+>             signingConfig = signingConfigs.release
 >         }
 >     }
 > }
 > ```
 >
-> Without it the release build is unsigned and Play rejects it.
+> Without the `signingConfig` the release build is unsigned and Play rejects it;
+> without the other two the Play Console reports the app as unoptimized. Bubblewrap
+> regenerates the `release` block with `minifyEnabled` alone, so all three lines have
+> to come back by hand.
 >
 > ⚠️ Same problem for **`targetSdkVersion`**: it is not exposed in
 > `twa-manifest.json`, so `bubblewrap update` resets it to whatever the installed
--- a/twa/README.md
+++ b/twa/README.md
@@ -300,7 +300,7 @@ bubblewrap update            # apply twa-manifest.json changes to the project
 >     }
 >     buildTypes {
 >         release {
->             minifyEnabled true                     // ← Bubblewrap emits only this
+>             minifyEnabled = true                   // ← Bubblewrap emits `minifyEnabled true`
 >             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
 >             shrinkResources = true
 >             signingConfig = signingConfigs.release
@@ -332,6 +332,11 @@ bubblewrap update            # apply twa-manifest.json changes to the project
 > - `shrinkResources`, `proguardFiles …-optimize.txt`, `proguard-rules.pro`
 > - AGP 9 DSL — `compileSdk`/`minSdk`/`targetSdk`, `lint {}`,
 >   `buildFeatures { resValues = true }`
+> - `tasks.register('generateShortcutsFile') { doLast { … } }` and
+>   `tasks.named('preBuild') { dependsOn('generateShortcutsFile') }` — Bubblewrap
+>   emits a bare `task generateShorcutsFile { … }` (note the typo) whose body runs
+>   during Gradle's configuration phase, so it rewrites `shortcuts.xml` on every
+>   invocation. The registered form makes it lazy and only writes when it runs.
 >
 > Elsewhere:
 >
--- a/twa/app/build.gradle
+++ b/twa/app/build.gradle
@@ -55,7 +55,10 @@ android {
     namespace "org.grottocenter.twa"
     defaultConfig {
         applicationId "org.grottocenter.twa"
-        minSdkVersion 21
+        // 23 (Android 6.0), not Bubblewrap's default of 21: androidbrowserhelper 2.7.x
+        // declares minSdk 23 and the manifest merger refuses anything lower. Forcing it
+        // with tools:overrideLibrary would only move the failure to runtime.
+        minSdkVersion 23
         targetSdkVersion 36
         versionCode 4
         versionName "4"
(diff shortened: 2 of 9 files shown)


ClemRz (commented):
This is a well-researched and thorough migration. The PR body alone is a reference document — the disassembly verification, the `fullScopeUrl` / `update_chrome_toast` shrinking analysis, and the `buildFeatures { resValues = true }` diagnosis are exactly the level of depth this kind of change requires. The README additions are genuinely useful for the next person who runs `bubblewrap update`. A few things worth addressing before the physical device testing checklist is cleared.

- **[twa/app/build.gradle:178]** `minifyEnabled true` uses the old Groovy DSL property form while `shrinkResources = true` and `signingConfig = signingConfigs.release` on the lines below use the AGP 9 property assignment syntax. In AGP 9 the canonical form for the minification flag is `isMinifyEnabled = true` (Kotlin DSL parity — `minifyEnabled` still works as a setter but it will emit a deprecation warning in a future toolchain). Aligning the block now avoids a follow-up migration:
  ```suggestion
              isMinifyEnabled = true
  ```

- **[twa/app/proguard-rules.pro:1–15]** The comment says _"Everything reachable is either declared in AndroidManifest.xml (activities…), which R8 keeps automatically"_, but in R8 full mode the manifest entry keeps the class itself reachable, not necessarily its no-arg constructor — which is exactly why `DelegationService` needs the explicit `<init>()` rule below. The `locationdelegation.PermissionRequestActivity` is resolved by Chrome Custom Tabs / TWA machinery via `PackageManager` (intent resolution), not via direct `new`-allocation, so the class survives. But given that `enableNotifications: true` and `locationDelegation: enabled: true` are both active, and the PR body already flags this as untested, it is worth adding a comment cross-referencing the open testing item so a future reader knows the omission is intentional, not an oversight:
  ```suggestion
  # locationdelegation.PermissionRequestActivity is resolved by Chrome/TWA via
  # PackageManager (intent resolution), not instantiated from code, so R8's activity
  # rules derived from the manifest are sufficient. If the geolocation permission
  # prompt stops working, add:
  #   -keep class com.google.androidbrowserhelper.locationdelegation.PermissionRequestActivity { <init>(); }
  -keep class com.google.androidbrowserhelper.trusted.DelegationService { <init>(); }
  -keep class org.grottocenter.twa.DelegationService { <init>(); }
  ```

### Nitpicks (Optional)

- **[twa/app/build.gradle:202, 257]** The task name `generateShorcutsFile` (missing the first `t` in *Shortcuts*) is inherited from the original Bubblewrap template. The task is being meaningfully refactored in this PR anyway, so it is a low-cost opportunity to fix the typo: `generateShortcutsFile`. It appears in both the `tasks.register(…)` call and the `dependsOn` reference, and in the README checklist should the author decide to document it.

- **[twa/README.md]** The `bubblewrap update` re-add checklist in the README is very thorough — one small gap: `tasks.register('generateShorcutsFile')` and `tasks.named('preBuild') { dependsOn(…) }` (the configuration-phase fix from this PR) are not in the list. If someone regenerates with Bubblewrap they will get the bare `task { … }` form back, which silently rewrites `shortcuts.xml` on every Gradle invocation. Worth adding a bullet alongside the others.

ClemRz (approved):
All four suggestions from the previous review have been applied correctly — `minifyEnabled = true`, the `locationdelegation` cross-reference comment, the `generateShortcutsFile` typo fix, and the README checklist update. The overall quality of this PR remains excellent: the diagnosis, disassembly verification, and `buildFeatures { resValues = true }` catch are thorough and well documented.

A few remaining things worth a look.

- **[twa/app/proguard-rules.pro:25]** `-keepattributes SourceFile,LineNumberTable` preserves line numbers in stack traces but also leaves the actual source file names in the DEX (e.g. `LauncherActivity.java`). For an open-source project this is not a leak concern, but the standard companion rule suppresses the class-internal name while keeping Play Console symbolication intact:
  ```suggestion
  -keepattributes SourceFile,LineNumberTable
  -renamesourcefileattribute SourceFile
  ```
  Without `renamesourcefileattribute`, the `.SourceFile` attribute in the DEX contains the real filename; with it, every class reports `"SourceFile"` literally, which R8's retrace tool still resolves correctly via the mapping file.

### Nitpicks (Optional)

- **[twa/app/build.gradle:54, 65, 69–70]** `compileSdk 36`, `minSdk 23`, `targetSdk 36`, and `applicationId "org.grottocenter.twa"` use the old Groovy method-call form while the lines around them (`namespace = "..."`, `shrinkResources = true`, `minifyEnabled = true`, `signingConfig = signingConfigs.release`) use the AGP 9 property-assignment form. AGP 9 accepts both, so this is purely a style note — aligning the whole block to the assignment form would make it consistent for the next person who reads it.

- **[twa/app/build.gradle:209–210]** The assert error messages for `url` and `icon` are swapped — a pre-existing bug inherited verbatim from the Bubblewrap template:
  ```
  assert s.url != null, 'Missing `icon` in shortcut #' + i   // says icon, checks url
  assert s.icon != null, 'Missing `url` in shortcut #' + i   // says url, checks icon
  ```
  Since `shortcuts: []` is currently empty this never fires, but since the task body was meaningfully refactored in this PR it is a low-cost fix worth making:
  ```suggestion
          assert s.url != null, 'Missing `url` in shortcut #' + i
          assert s.icon != null, 'Missing `icon` in shortcut #' + i
  ```

PR #1506, opened 2026-08-11
Feat/1404 sensitivity lock

### LINKED ISSUE #1404: Admin-controlled lock on entrance and massif sensitivity user interface
# Requirements Document

## Introduction

Add an `isSensitiveLocked` boolean field to both entrances and massifs in the GrottoCenter frontend. When an administrator locks the sensitivity of an entrance or massif, only administrators can subsequently change its `isSensitive` value. The lock is a deliberate, symmetric admin action — it can freeze sensitivity in either state (on or off). Additionally, massif cascade operations (marking/unmarking a massif as sensitive) must skip entrances whose sensitivity is locked.

This document covers the **frontend portion only** (`grottocenter-front` React SPA). It assumes the backend API already supports the `isSensitiveLocked` field on entrance and massif create, update, and find endpoints, and that the backend cascade logic skips locked entrances.

## Glossary

- **Entrance_Form**: The React Hook Form–based entrance create/edit form rendered by the `EntranceDetail` component in `components/appli/EntitiesForm/Entrance/EntranceDetail.jsx`.
- **MassifSensitivityControl**: The admin-only component in `components/appli/EntitiesForm/Massif/MassifSensitivityControl.jsx` that manages massif sensitivity with a preview/confirm cascade flow.
- **Properties_View**: The `Properties` component in `components/appli/Entry/Properties.jsx` that displays entrance attributes on the detail page.
- **Entrance_Detail_Page**: The `Entry` component in `components/appli/Entry/index.jsx` that renders the entrance detail page including the sensitivity warning.
- **EntrancePropTypes**: The PropTypes shape definition in `types/entrance.type.js` that describes the entrance data structure used across components.
- **MassifTypes**: The PropTypes shape definition in `types/massif.type.js` that describes the massif data structure.
- **Translation_Files**: The JSON translation files in `public/lang/` (15 languages) used by React Intl for internationalization.
- **isSensitiveLocked**: A boolean attribute on an entrance or massif. When `true`, only administrators can change the `isSensitive` value of that entity. Defaults to `false` (unlocked).
- **Lock_Toggle**: A UI control (padlock icon button) adjacent to the sensitivity switch, visible only to administrators, that sets or unsets `isSensitiveLocked`.
- **Administrator**: A user with the "Administrator" role as determined by the `usePermissions` hook.
- **Cascade_Operation**: The process by which marking or unmarking a massif as sensitive propagates the `isSensitive` value to all entrances within the massif polygon.

### Requirement 1: Add isSensitiveLocked to type definitions

**User Story:** As a developer, I want the entrance and massif PropTypes to include `isSensitiveLocked`, so that components consuming these entities have a consistent type contract.

#### Acceptance Criteria

1. THE EntrancePropTypes shape in `entrance.type.js` SHALL include an `isSensitiveLocked` property defined as `PropTypes.bool` (optional, not required).
2. THE MassifTypes shape in `massif.type.js` SHALL include an `isSensitiveLocked` property defined as `PropTypes.bool` (optional, not required).
3. WHEN a component consuming EntrancePropTypes or MassifTypes does not provide `isSensitiveLocked`, THE PropTypes validation SHALL pass without warnings.

### Requirement 2: Entrance sensitivity switch respects lock state

**User Story:** As a caver, I want the sensitivity switch on the entrance form to be disabled when the entrance is locked, so that I cannot accidentally change a locked entrance's sensitivity.

#### Acceptance Criteria

1. IF an entrance has `isSensitiveLocked` set to `true` and the current user is not an Administrator, THEN THE Entrance_Form SHALL disable the `isSensitive` switch.
2. IF an entrance has `isSensitiveLocked` set to `true` and the current user is an Administrator, THEN THE Entrance_Form SHALL keep the `isSensitive` switch enabled.
3. IF an entrance has `isSensitiveLocked` set to `false`, THEN THE Entrance_Form SHALL apply the standard disable logic where non-admin users cannot unmark a sensitive entrance but can still mark a non-sensitive entrance as sensitive.
4. IF the `isSensitive` switch is disabled due to `isSensitiveLocked` being `true`, THEN THE Entrance_Form SHALL replace the existing sensitivity alert with an informational message indicating that an administrator has locked the sensitivity of this entrance.
5. IF the `isSensitive` switch is disabled due to `isSensitiveLocked` being `true` and the entrance is also marked as sensitive (`isSensitive` is `true`), THEN THE Entrance_Form SHALL display the lock message and hide the coordinate fields, consistent with the existing behavior for sensitive entrances.

### Requirement 3: Admin lock toggle on entrance form

**User Story:** As an administrator, I want a padlock toggle next to the entrance sensitivity switch, so that I can lock or unlock the sensitivity state of an entrance.

#### Acceptance Criteria

1. WHEN the current user is an Administrator, THE Entrance_Form SHALL render a Lock_Toggle control adjacent to the `isSensitive` switch.
2. WHEN the current user is not an Administrator, THE Entrance_Form SHALL not render the Lock_Toggle control.
3. WHEN an Administrator activates the Lock_Toggle, THE Entrance_Form SHALL set the form value `entrance.isSensitiveLocked` to `true`.
4. WHEN an Administrator deactivates the Lock_Toggle, THE Entrance_Form SHALL set the form value `entrance.isSensitiveLocked` to `false`.
5. THE Lock_Toggle SHALL display a locked padlock icon when `entrance.isSensitiveLocked` is `true` and an unlocked padlock icon when `entrance.isSensitiveLocked` is `false`.
6. WHILE `entrance.isSensitiveLocked` is `true`, THE Entrance_Form SHALL disable the `isSensitive` switch so that its value cannot be changed by any user.
7. WHEN the Entrance_Form loads with an existing entrance, THE Lock_Toggle SHALL reflect the persisted value of `entrance.isSensitiveLocked` (active if `true`, inactive if `false`).

### Requirement 4: Massif sensitivity switch respects lock state

**User Story:** As an administrator, I want the massif sensitivity switch to be disabled when the massif is locked, so that accidental changes to a locked massif's sensitivity are prevented.

#### Acceptance Criteria

1. WHEN a massif has `isSensitiveLocked` set to `true` and the current user is an Administrator, THE MassifSensitivityControl SHALL disable the sensitivity switch and preserve its current checked state reflecting the massif's `isSensitive` value.
2. WHEN a massif has `isSensitiveLocked` set to `true`, THE MassifSensitivityControl SHALL display a visible informational message below the switch explaining that the massif sensitivity is locked and must be unlocked before changes can be made.
3. WHEN a massif has `isSensitiveLocked` set to `false`, THE MassifSensitivityControl SHALL keep the sensitivity switch enabled for administrators.
4. IF the massif's `isSensitiveLocked` property is undefined or null, THEN THE MassifSensitivityControl SHALL treat the lock state as `false` and keep the sensitivity switch enabled for administrators.

### Requirement 5: Admin lock toggle on massif sensitivity control

**User Story:** As an administrator, I want a padlock toggle in the massif sensitivity panel, so that I can lock or unlock the sensitivity state of a massif.

#### Acceptance Criteria

1. THE MassifSensitivityControl SHALL render a Lock_Toggle control adjacent to the sensitivity switch only when the current user has the Administrator role.
2. WHEN an Administrator activates the Lock_Toggle, THE MassifSensitivityControl SHALL dispatch an action to set `isSensitiveLocked` to `true` on the massif.
3. WHEN an Administrator deactivates the Lock_Toggle, THE MassifSensitivityControl SHALL dispatch an action to set `isSensitiveLocked` to `false` on the massif.
4. THE Lock_Toggle SHALL visually indicate the current lock state (locked icon when `isSensitiveLocked` is `true`, unlocked icon when `isSensitiveLocked` is `false`).
5. WHILE `isSensitiveLocked` is `true` on the massif, THE MassifSensitivityControl SHALL disable the sensitivity switch.
6. IF the lock or unlock action fails, THEN THE MassifSensitivityControl SHALL display an error notification indicating the failure reason and revert the Lock_Toggle to its previous state.
7. WHILE a lock or unlock action is in progress, THE MassifSensitivityControl SHALL disable the Lock_Toggle until the action completes or fails.

### Requirement 6: Massif cascade skips locked entrances with user feedback

**User Story:** As an administrator performing a massif cascade, I want to be informed about entrances that were skipped due to locks, so that I know the cascade did not affect all entrances.

#### Acceptance Criteria

1. WHEN a massif sensitivity cascade operation completes and the backend response includes a skipped locked entrance count greater than 0, THE MassifSensitivityControl SHALL display the count of affected entrances and the count of skipped locked entrances in the success toast notification.
2. WHEN a massif sensitivity cascade preview is requested and the backend response includes a locked entrance count greater than 0, THE MassifSensitivityControl SHALL include the locked entrance count in the confirmation dialog body alongside the count of entrances that will be affected.
3. WHEN a massif sensitivity cascade preview is requested and the backend reports that all entrances in the massif are locked, THE MassifSensitivityControl SHALL display a message in the confirmation dialog body indicating that no entrances will be affected by the cascade.
4. IF all entrances in a massif are locked, THEN THE MassifSensitivityControl SHALL still allow the administrator to confirm the cascade operation to mark the massif itself as sensitive.

### Requirement 7: Display lock status on entrance detail page

**User Story:** As a visitor viewing an entrance detail page, I want to see whether the entrance sensitivity is locked, so that I understand the governance status of that entrance.

#### Acceptance Criteria

1. WHEN an entrance has both `isSensitive` set to `true` and `isSensitiveLocked` set to `true`, THE Entrance_Detail_Page SHALL display a lock indicator (padlock icon) alongside the existing sensitivity warning (SensitiveCaveWarning component).
2. WHEN an entrance has `isSensitiveLocked` set to `true` but `isSensitive` set to `false`, THE Entrance_Detail_Page SHALL not display the lock indicator (since the sensitivity warning itself is not shown).
3. WHEN an entrance has `isSensitiveLocked` set to `false` or undefined, THE Entrance_Detail_Page SHALL not display a lock indicator.
4. THE lock indicator SHALL use a localized tooltip explaining that sensitivity is locked by an administrator.

### Requirement 8: Include isSensitiveLocked in API payloads

**User Story:** As a developer, I want the form-to-API transformers to include `isSensitiveLocked`, so that the lock state is sent to the backend on create and update.

#### Acceptance Criteria

1. WHEN entrance form data is transformed for an API request, THE entrance transformer SHALL include the `isSensitiveLocked` boolean field from `data.entrance.isSensitiveLocked` in the output object produced by `makeEntranceData`.
2. IF `isSensitiveLocked` is `undefined` or `null` in the entrance form data, THEN THE entrance transformer SHALL omit the field from the output object so that the serialized JSON does not contain the key and the API applies its default.
3. WHEN a massif lock toggle action is dispatched, THE corresponding action creator SHALL send a PATCH request to the massif API endpoint with a body containing the `isSensitiveLocked` boolean field set to the new lock state value.

### Requirement 9: Add translation keys for lock UI labels

**User Story:** As a user browsing in any supported language, I want the lock-related labels and messages to appear in my language, so that the interface remains consistent and understandable.

#### Acceptance Criteria

1. THE Translation_Files SHALL include a key whose value describes toggling the lock state of sensitivity (conveying both "Lock sensitivity" and "Unlock sensitivity" semantics via the toggle tooltip).
2. THE Translation_Files SHALL include a key whose value conveys an informational message explaining why the sensitivity switch is disabled on a locked entrance.
3. THE Translation_Files SHALL include a key whose value conveys an informational message explaining the locked state within the massif sensitivity control.
4. THE Translation_Files SHALL include a key whose value describes a cascade notification indicating that locked entrances were skipped during the sensitivity cascade operation.
5. THE Translation_Files SHALL include a key whose value describes the lock indicator tooltip displayed on the entrance detail page.
6. WHEN a translation key is added to the English master file, THE Translation_Files for all other supported languages SHALL include the same key with the English value as a placeholder until translated.
7. WHEN translation keys are added or modified, THE Translation_Files SHALL have their keys sorted in alphabetical order.

dawoldo:
# Let administrators lock entrance and massif sensitivity

## 🤔 What

Adds an `isSensitiveLocked` flag to entrances and massifs, so an administrator can freeze a
sensitivity value and prevent anyone else from changing it. Closes #1404.

- New admin-only padlock toggle (`SensitivityLockToggle`) next to the sensitivity switch on
  both the entrance form and the massif sensitivity panel.
- A locked entrance disables its "Restricted access entrance" switch for non-admins and
  explains why with an info alert.
- A locked massif disables its sensitivity switch **for everyone, admins included** — it has
  to be unlocked first (this mirrors the API, which returns `403` on `mark-sensitive` /
  `unmark-sensitive` while the massif is locked).
- The massif cascade preview dialog and success toast now report how many entrances were
  skipped because their sensitivity is locked, including the "everything is locked, only the
  massif itself will change" case.
- Sensitive entrance detail pages show a padlock next to the existing restricted-access
  warning when the lock is on.
- `isSensitiveLocked` added to `EntrancePropTypes` and `MassifTypes`, to the entrance
  form/API transformers, and to all 15 translation files.

## 🤷‍♂️ Why

Sensitivity is a governance decision, not a personal preference: some entrances are marked
restricted (or deliberately left unrestricted) because of a land-manager agreement or local
legislation, and today any contributor can flip that back on the next edit. There was also no
way to protect an individual entrance from a massif-wide cascade.

The lock is deliberately symmetric — it freezes sensitivity in either state, so it protects
both "this cave must stay protected" and "this cave must stay visible" decisions.

## 🔍 How

**Entrance form** — `EntranceDetail` used to derive one boolean for two unrelated things.
It now separates them:

- `areCoordinatesHidden = !isAdmin && initialIsSensitive` — unchanged behaviour, hides the
  coordinate section and blocks unrestricting.
- `isLockedForUser = isSensitiveLocked && !isAdmin` — the new lock rule.

The switch is disabled when either is true, and a single `sensitivityAlert` object picks the
message by priority (lock → can't-unrestrict → the standard sensitivity warning), which keeps
the "locked **and** sensitive" case correct: lock message shown, coordinates still hidden.

The padlock is a `Controller` on `entrance.isSensitiveLocked` rendered only for admins, so a
non-admin submit never registers the field and the transformer leaves the key out of the
payload entirely (`data.entrance.isSensitiveLocked != null` guard in `makeEntranceData`) —
the API then keeps its own value. `hasEntranceChanged` compares the flag too, so toggling
only the padlock still counts as a change.

**Massif panel** — `MassifSensitivityControl` keeps optimistic local lock state, dispatches
the new `setMassifSensitiveLock` thunk (`PUT /massifs/:id` with `{ isSensitiveLocked }`), and
reverts the toggle plus raises an error notification if the call fails. The toggle is disabled
while the request is in flight. The 403 handling was extracted into a shared
`notifyActionError` used by both the cascade and the lock paths.

`previewSensitiveMassif` now resolves `{ count, lockedCount }` instead of a bare number, and
`markMassifSensitive` carries `skippedLockedCount` through to the success toast. The two
backend counts are disjoint by construction (`is_sensitive_locked = false` vs `= true` in the
same query), so "will mark {count} … {lockedCount} skipped" does not double-count.

Unmarking intentionally reports no counts: the API's `unmark-sensitive` never cascades to
entrances, it only clears the massif.

**Note on the spec** — the requirements doc says the entrance switch should be disabled for
*any* user when locked (Req 3.6), which contradicts Req 2.2 and the doc's own introduction.
This PR follows 2.2: admins keep editing a locked entrance directly, and only massifs require
an explicit unlock. The doc also asks for a `PATCH`; the API route is `PUT /api/v1/massifs/:id`,
so `PUT` is what's implemented.

## 🔗 Related API PR

grottocenter-api#1713 — *Feat/1684 massif sensitivity status lock* (API issue #1684).

**Already merged into API `develop`**, so the Azure Static Web Apps stage site should work
end-to-end as soon as the API stage deploy is up to date. Verified against the API `develop`
tree: `toEntrance` / `toMassif` both expose `isSensitiveLocked`, `massif/update` strips the
field for non-admins, and `coerceBool` returns `undefined` for absent params (Waterline drops
undefined keys), so a partial `{ isSensitiveLocked }` body cannot clobber other massif fields.

## 🗄️ Database migrations

⚠️ **This feature does not work without a DB migration.** It is in the API repo, not here:

`sql/9_16_2026_06_29_entrance_massif_is_sensitive_locked.sql` (on API `develop`)

- adds `is_sensitive_locked bool NOT NULL DEFAULT false` to `t_entrance`, `h_entrance`,
  `t_massif` and `h_massif`;
- recreates `histo_update_entrance()` and `histo_update_massif()` so the new column is
  archived in history, mirroring how `is_sensitive` is handled.

The column additions are guarded by `information_schema` checks and the whole file runs in a
single transaction, so it is idempotent and safe to re-run.

Apply in numeric order — `sql/9_17_2026_07_09_fix_history_triggers.sql` (fix #1723) also
recreates `histo_update_entrance()` with `is_sensitive_locked` in its `INSERT`, so running
9_17 on a database that never got 9_16 leaves a trigger referencing a missing column and
breaks *every* entrance update.

No search-index work is needed: `is_sensitive_locked` is not part of the Typesense schema, so
no collection re-creation or reindex.

## 🧪 Testing

Local checks on this branch: `yarn lint` → 0 errors, `yarn test` → 84 files / 773 tests pass,
`yarn translations:check-sorted` and `yarn translations:check-en` → pass, completeness report
→ 15/15 languages at 100 % (real translations, not English placeholders).

New unit tests cover `SensitivityLockToggle` (icon/label per state, click emits the opposite
state, disabled while pending) and the transformer (key omitted when absent, coerced to
boolean, change detection).

Manual scenarios — you need one admin account and one regular account:

**Entrance**
1. As admin, edit an entrance, click the padlock, save. Reopen: the padlock is still closed.
2. As a regular user, edit that entrance: the sensitivity switch is disabled and an info
   alert says an administrator locked it.
3. Lock an entrance that is *also* sensitive, then reopen as a regular user: lock message
   shown **and** coordinate fields hidden.
4. As admin, confirm the sensitivity switch stays usable on a locked entrance.
5. Unlock, save, and confirm a regular user can mark/unmark again per the usual rules.

**Massif**
6. As admin, open a massif with entrances and lock it: the sensitivity switch greys out and
   an info alert appears below it.
7. Unlock, then toggle sensitivity on: the confirm dialog reports both the affected count and
   the skipped-locked count. Lock a couple of the massif's entrances first to see it.
8. Lock *every* entrance in the massif, then cascade: the dialog says no entrance will be
   affected but still lets you mark the massif itself.
9. Toggle the padlock while offline / with the API stopped: the toggle reverts and an error
   notification appears.

**Detail page**
10. As admin, open a sensitive + locked entrance: a padlock appears on the restricted-access
    warning with a localized tooltip. Note this warning is admin-only today (non-admins get
    `SensitiveLocationPlaceholder` instead), so the indicator follows it.


ClemRz (changes requested):
Great PR overall — the feature is well-scoped, the PR description is thorough, all 15 translation files are updated and sorted, tests cover the new component and the transformer, and the diff is clean. The architecture (optimistic lock toggle with revert, shared `SensitivityLockToggle`, alert priority logic) is solid.

Two issues need fixing before merge.

---

### Issues (Must Fix)

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/transformers.js:33 / index.jsx:51]** **Non-admin form submission inadvertently unlocks a locked entrance.**

  `defaultEntranceValues.isSensitiveLocked` is `false` (not `null`). React Hook Form keeps the field value at its default even when the `Controller` isn't mounted (since `shouldUnregister` defaults to `false`). So when a **non-admin** submits, `data.entrance.isSensitiveLocked` is `false`, the `!= null` guard in `makeEntranceData` passes, and the PUT payload contains `isSensitiveLocked: false` — silently unlocking a previously locked entrance.

  The PR description says "a non-admin submit never registers the field and the transformer leaves the key out of the payload entirely", but React Hook Form's default behaviour contradicts this assumption.

  Two complementary fixes are needed:

  1. **In `defaultEntranceValues`** — default the field to `null` instead of `false` so unregistered fields produce `null` rather than `false`, letting the `!= null` guard do its job:
  ```suggestion
    isSensitiveLocked: null,
  ```

  2. **In `hasEntranceChanged`** — the comparison `Boolean(undefined) !== Boolean(undefined)` already resolves to `false` with the change above, so no separate fix is needed there. However, to be safe and explicit, consider guarding the comparison so that an absent key in `entranceDataFmt` (when the API key was omitted) never triggers a false change:
  ```suggestion
    ('isSensitiveLocked' in entranceDataFmt
      ? Boolean(entranceDataFmt.isSensitiveLocked) !==
        Boolean(originalEntranceValues.isSensitiveLocked)
      : false) ||
  ```

  Without at least fix (1), an admin locking an entrance can have the lock silently cleared by any non-admin editing and saving the same entrance.

- **[packages/web-app/src/components/appli/Entry/SensitiveCaveWarning.jsx:22]** **Keyboard/assistive-technology users cannot access the lock tooltip.**

  `LockOutlinedIcon` is a non-interactive `svg`. MUI's `Tooltip` only fires on `focus` and `mouseenter`, and an `svg` is not focusable by default. The `aria-label` makes it meaningful to screen readers, but keyboard users can never reach it.

  Wrap the icon in a `<span tabIndex={0}>` (or a `<IconButton>` if a click action ever makes sense here) so the tooltip is keyboard-reachable and `aria-label` works end-to-end:
  ```suggestion
          <Tooltip title={lockLabel}>
            <span tabIndex={0} style={{ display: 'inline-flex', alignItems: 'center' }}>
              <LockOutlinedIcon aria-label={lockLabel} fontSize="small" />
            </span>
          </Tooltip>
  ```

---

- **[packages/web-app/src/actions/Massif/SetMassifSensitiveLock.js:34]** The `setMassifSensitiveLock` thunk re-throws on non-auth errors instead of dispatching a structured failure action. The component handles the error correctly (optimistic revert + toast), but the pattern diverges from every other massif action creator in the repo (`updateMassif`, `loadMassif`, etc.) which dispatch a `*_FAILURE` action. This matters if any future middleware or DevTools tooling needs to observe all failure paths. Consider dispatching a `SET_MASSIF_SENSITIVE_LOCK_FAILURE` action before re-throwing (or instead of throwing):
  ```javascript
  export const SET_MASSIF_SENSITIVE_LOCK_FAILURE =
    'SET_MASSIF_SENSITIVE_LOCK_FAILURE';

  // in the catch:
  .catch(error => {
    dispatch({ type: SET_MASSIF_SENSITIVE_LOCK_FAILURE, error });
    throw error; // still let the component catch it for the optimistic revert
  });
  ```

- **[packages/web-app/src/components/appli/EntitiesForm/Massif/MassifSensitivityControl.jsx:51]** `notifyActionError` is defined inside the component render function (after the early-return `if (!isAdmin || !massif?.id) return null`). This is fine functionally since the early return means it is only called when both guards pass, but the function definition order is a bit unusual — consider moving it above the early return to signal that it is a utility, not an admin-path detail.

- **[packages/web-app/src/components/appli/EntitiesForm/Massif/MassifSensitivityControl.jsx:99-122]** The success toast `count` and `lockedCount` are read from two sources: `result?.count ?? preview?.count ?? 0`. If the API returns `{ count: 0, skippedLockedCount: 5 }`, `result.count` is `0`, which is falsy but valid — `?? 0` handles this correctly since `??` is not `||`. This is fine as written, but worth a comment noting that `0` is a valid count.

- **[packages/web-app/src/components/common/SensitivityLockToggle.jsx:22]** The `aria-pressed` attribute on the `IconButton` correctly communicates toggle state to assistive technologies. Consider also adding `role="switch"` or at minimum a visible label variant for the two states so low-vision users (who may not process icon shape as reliably as colour) also get the context. The existing tooltip already covers hover/focus, but it requires pointer or focus proximity — a visible text label (even `sr-only`) is more robust.

---

### Nitpicks (Optional)

- **[packages/web-app/src/actions/Massif/SetMassifSensitiveLock.js:1]** The comment block on the function is thorough. One small inconsistency: the module-level JSDoc says `PUT /api/v1/massifs/:id` but other action files in the same folder don't include the HTTP verb in the JSDoc — they describe the semantic intent. Minor, no action needed.

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/EntranceDetail.jsx:43-46]** The double-negative `const isSensitiveLocked = !!useWatch(...)` is idiomatic and clear. A small note: `useWatch` for `isSensitiveLocked` will re-render `EntranceDetail` on every lock toggle. Since this component also watches `isSensitive`, `latitude`, and `longitude` (via individual `useWatch` calls and `nearbyEntrances`), and mounts a map, the renders are already happening for those changes — so this is fine as-is.

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/transformers.sensitiveLock.test.js]** Tests are excellent and cover the important null/undefined omission and boolean coercion cases. Consider adding a test for the non-admin scenario specifically (form submitted with `isSensitiveLocked: null` should omit the key) once fix (1) above is applied — it would prevent a regression of the original issue.

Paul-AUB (changes requested):
(See my precedent message)

ClemRz (approved):
The two must-fix issues from the previous review have been fully addressed in the follow-up commits. The implementation is now correct and the PR is in good shape.

### Previous issues — resolved ✅

- **Non-admin unlock bug** — `useController` in `EntranceDetail` now has no `defaultValue`, so a locked entrance stays locked when a non-admin saves it. The `sensitiveLockSubmit.test.jsx` regression test pins this exactly.
- **Keyboard accessibility** — `SensitiveCaveWarning` wraps the lock icon in a `<Box component="span" tabIndex={0} role="img" aria-label={lockLabel}>`, making it focusable and meaningful to assistive technologies. `SensitiveCaveWarning.test.jsx` asserts both the accessible name and the tab stop.

### Paul-AUB's UI/UX concerns — addressed ✅

- **Cancel/Confirm button hierarchy** — `Cancel` is now `outlined`, `Confirm` is `warning` + `contained`.
- **Lock fires immediately without saving** — The massif lock is now part of the form's RHF state and dispatched only inside `onSubmit`, not on toggle.
- **Sensitivity panel position** — `MassifSensitivityControl` comes after the map/fields with an explanatory comment.
- **Consistent UI across entrance and massif** — Both entities now share `SensitivitySection` and `SensitivityLockToggle`, giving an identical visual language.
- **Lock icon colour** — Warning colour is applied when locked (`color={isLocked ? 'warning' : 'default'}`).
- **Lock on massif creation** — Works now: the form accepts the lock value, `postMassif` runs first to get the id, then `setMassifSensitiveLock` runs in the same submit handler.

---

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/transformers.js:75]** `hasEntranceChanged` compares `isSensitiveLocked` unconditionally. When the transformer **omits** the key (e.g. field value is `null`/`undefined`), `entranceDataFmt.isSensitiveLocked` is `undefined` and `Boolean(undefined)` evaluates to `false`. Against an original value of `true` this would be a false-positive "changed" detection, causing an unnecessary PUT. In practice this can't happen with the current form wiring (the `useController` default always reads from the RHF tree, which is populated by `{ ...defaultEntranceValues, ...entranceValues }`), but the function's comment says *"keep in sync with the entranceValues prop shape"* — a future field addition could accidentally trip this. Consider mirroring the coordinate-omission guard:
  ```javascript
  ('isSensitiveLocked' in entranceDataFmt
    ? Boolean(entranceDataFmt.isSensitiveLocked) !==
      Boolean(originalEntranceValues.isSensitiveLocked)
    : false) ||
  ```

- **[packages/web-app/src/actions/Massif/SetMassifSensitiveLock.js:34]** The action still re-throws non-auth errors without dispatching a failure action first. The massif form's `onSubmit` try/catch handles this correctly for the on-save flow. However, since `SET_MASSIF_SENSITIVE_LOCK_SUCCESS` is now imported by `MassifReducer`, a symmetrical `SET_MASSIF_SENSITIVE_LOCK_FAILURE` would complete the Redux action triad and make the failure path observable in DevTools. Low priority but worth noting for consistency with the rest of the massif action file family.

---

### Nitpicks (Optional)

- **[packages/web-app/src/components/appli/EntitiesForm/Massif/index.jsx:140]** `notifyCascadeResult` is defined as a `const` inside the component body rather than as a `useCallback`. Since it's only called inside the `async onSubmit`, this is harmless (no stale-closure risk), but `eslint-plugin-react-hooks`'s exhaustive-deps rule may flag it on a future refactor. Wrapping it in `useCallback` pre-empts that.

- **[packages/web-app/src/components/common/SensitivitySection.jsx:26-30]** `onLockChange = undefined` as a default prop while the prop type says `PropTypes.func` (no `.isRequired`) is intentional — `showLock` gates whether it's ever called. Consider adding a comment to the propTypes block explaining the pairing, so a future reader doesn't add `.isRequired` by reflex.

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/sensitiveLockSubmit.test.jsx:17]** The test harness mirrors `defaultEntranceValues.isSensitiveLocked: false` correctly. The comment *"the value can only come from the form defaults"* is accurate and helpful — keep it. Consider also noting that the test intentionally doesn't import from `index.jsx` to stay unit-focused.

Paul-AUB:
✅ Fixed — `hasEntranceChanged` now ignores `isSensitiveLocked` when the transformer intentionally omits that key, preventing a false-positive update. A regression test covers an omitted lock state against an originally locked entrance.

⏭️ Not fixed — the suggested Redux `SET_MASSIF_SENSITIVE_LOCK_FAILURE` action is no longer applicable. The legacy action was removed during the React Query migration; `useSetMassifSensitiveLock` now exposes mutation errors through TanStack Query, and the form handles `mutateAsync` failures.

⏭️ Not fixed — `notifyCascadeResult` remains a plain local function. It is neither passed to a child nor used as a hook dependency, so `useCallback` would add dependency bookkeeping without providing a stable-identity benefit.

✅ Fixed — documented why `onLockChange` is optional in `SensitivitySection`: `showLock` gates whether the callback can be used.

✅ Fixed — documented that the lock-submit test keeps its minimal defaults local intentionally so the harness remains unit-focused.

Validation: full ESLint passes, and the targeted sensitivity-lock suite passes (12 tests).

urien (approved):
Je viens de faire un test utilisateur et tout fonctionne
En particvulier le fait de bloquer a Non sensible une cavité touristique dans un massif "sensible", ce qui est la demande de la NSS
https://orange-rock-0d4f87503-1506.westeurope.azurestaticapps.net/ui/entrances/10464

ClemRz (approved):
The previous review rounds addressed the two must-fix issues (non-admin unlock bug and keyboard accessibility) and the UI/UX concerns Paul raised. The code is in good shape. A few items still worth calling out below.

- **[packages/web-app/src/components/appli/EntitiesForm/Entrance/sensitiveLockSubmit.test.jsx:41]** The admin `Controller` in the test harness keeps `defaultValue={false}`, whereas the real `EntranceDetail` uses `useController` with **no** `defaultValue` (the post-fix state). The comment correctly notes that the persisted value comes from the RHF tree, and the "keeps a locked entrance locked" test still passes because `{ ...defaultEntranceValues, ...entranceValues }` resolves `isSensitiveLocked` to `true` before the Controller mounts. The two wirings are functionally equivalent here, but a reader comparing the harness to `EntranceDetail` may wonder why `defaultValue={false}` is there at all. Consider dropping it from the harness to stay in sync with the real component:
  ```javascript
  // Drop the defaultValue — matches EntranceDetail's useController({no defaultValue})
  <Controller
    name="entrance.isSensitiveLocked"
    control={control}
    render={({ field: { value, onChange } }) => (
      <button type="button" onClick={() => onChange(!value)}>
        toggle lock
      </button>
    )}
  />
  ```

- **[packages/web-app/src/components/appli/EntitiesForm/Massif/MassifReactQueryFlow.test.jsx:65]** `MassifSensitivityControl` is mocked down to a single `onSensitiveChange` button. The test covers the cascade and error paths well, but the lock-on-create path (`setMassifLock.mutateAsync` called after `createMassif.mutateAsync` when `nextLocked` is true) has no integration test. Consider adding a scenario: create a massif with the sensitivity lock toggled on, confirm that `setMassifLock.mutateAsync` is called with the new id after `createMassif.mutateAsync` resolves.

- **[packages/web-app/src/components/appli/EntitiesForm/Massif/index.jsx:152]** `notifyCascadeResult` reads `result?.skippedLockedCount` from the API response but `preview?.lockedCount` from the preview object. The asymmetric property names (`skippedLockedCount` vs `lockedCount`) are intentional (the preview returns `lockedCount`; the mark-sensitive response returns `skippedLockedCount`), but a comment here would save the next reader from second-guessing the mismatch:
  ```javascript
  // preview uses `lockedCount`; the mark response uses `skippedLockedCount`.
  const lockedCount = result?.skippedLockedCount ?? preview?.lockedCount ?? 0;
  ```

- **[packages/web-app/src/components/appli/Entry/index.jsx:379]** The lock indicator in `SensitiveCaveWarning` is only shown when `entrance.isSensitive && isAdmin`. Per requirement 7.1 in issue #1404, the indicator was originally specified for "a visitor", but the PR description notes it is admin-only because non-admins get `SensitiveLocationPlaceholder` instead. The implementation is intentional and documented in the PR — just worth noting in a code comment so a future contributor doesn't move the block without realising the non-admin path is already handled elsewhere:
  ```javascript
  {/* Lock indicator is admin-only: non-admins get SensitiveLocationPlaceholder
      and never reach this code path. */}
  {entrance.isSensitive && isAdmin && (
    <SensitiveCaveWarning isLocked={!!entrance.isSensitiveLocked} />
  )}
  ```

### Nitpicks (Optional)

- **[packages/web-app/src/hooks/queries/usePreviewSensitiveMassif.js:25]** Removing `staleTime` means `fetchQuery` with no `staleTime` will by default treat data as stale immediately — which is the intended "always refetch" semantics. This is correct as written and confirmed by the test. The JSDoc comment explains it clearly.

- **[packages/web-app/src/components/common/SensitivitySection.jsx:72]** `{children}` slot after the padlock is a nice escape hatch. Its `PropTypes.node` type accepts anything including interactive elements that could interfere with the switch row's flex layout. This is fine for now; just keep it in mind if future callers pass something taller than a button.

- **[packages/web-app/src/components/appli/EntitiesForm/Massif/index.jsx:192-197]** The early-return guard (`if (data.massif.geogPolygon?.coordinates?.length === 0) return;`) silently swallows the submit without user feedback when a polygon was somehow set to empty. This existed before this PR, but now that the submit path is more involved (cascade + lock), a `return` here with no notification could confuse an admin who toggled the lock and hit save. Pre-existing issue, no action needed from this PR.

Paul-AUB (approved):

Paul-AUB:
✅ Fixed — removed the test-only `Controller.defaultValue` so the entrance lock harness now matches `EntranceDetail` exactly.

✅ Fixed — added a massif creation regression test that enables the sensitivity lock, verifies the lock mutation receives the newly created massif ID, and asserts creation completes before locking.

✅ Fixed — documented the intentional `lockedCount` / `skippedLockedCount` difference between the preview and mutation responses.

✅ Fixed — documented why the sensitive entrance lock indicator is admin-only: non-admins follow the `SensitiveLocationPlaceholder` path.

Validation: full ESLint passes, and the targeted entrance/massif sensitivity suites pass (2 files, 8 tests).

ClemRz:
Looks ready to be merged.

PR #1507, opened 2026-08-11
feat(home): promote the mobile app and account creation on the home page

### LINKED ISSUE #884: [HOME PAGE] Info et promotion de l'appli mobile
###Ill convient de traiter ce ticket la solution mise en oeuvre pouvant être très différente de ce qui est indiqué.
### Attendre que l'app soit sur l'app store d'Apple

**Avoir un accès facile à l'appli mobile**

À terme (quand les tests auront été terminés et l'appli validée) il faudra prévoir sur la page d'accueil de Grottocenter un lien de téléchargement vers celle-ci ou a minima une fenêtre de présentation mentionnant la procédure pour pouvoir la récupérer.

En effet,  même si je n'en suis pas, il y a pas mal de monde qui peut télécharger une appli depuis Internet sur son smartphone et l'ajouter directement sans passer par l'installation d'un fichier .apk.

La fenêtre d'accueil de droite étant déjà pas mal chargée en infos je pense qu'il vaudrait mieux ajouter un item cliquable dans le menu de gauche qui renverrait vers une page spécifique à l'appli mobile dans la fenêtre de droite.

Visuel possible : 

Image10

Paul-AUB:
# feat(home): promote the mobile app and account creation on the home page

## 🤔 What

Two additions to the home page, plus a small consistency pass on its section titles.

- **Mobile app promotion**
  - New `AppPromoBanner` section with the official "Get it on Google Play" badge, linking to `org.grottocenter.twa`.
  - New `IosInstallPrompt` that explains the Safari `Share → Add to Home Screen` flow, since iOS has no install prompt of its own.
  - Both are hidden when the visitor already browses from the installed PWA / Android TWA (new `useCanPromoteApp` hook).
- **Account creation**
  - New `JoinCommunity` section listing four concrete benefits (edit caves, track explored entrances, rate caves, message cavers) with a single prominent "Create an account" CTA.
  - The discreet signup button that used to sit at the bottom of `Welcome` is removed — it is superseded by this section.
- **Polish**
  - Home page section titles moved from `h5` to `h3` so they follow the heading scale used elsewhere.
  - Rounded, slightly raised primary buttons (header CTA, "Donate now") for a more inviting call to action.
- 16 language files updated with the new keys.

## 🤷‍♂️ Why

The Android app and the PWA install were essentially undiscoverable: nothing on the site pointed to them, so visitors had no way of knowing Grottocenter can be used offline in the field — which is exactly where cavers need it.

On the account side, the previous signup button was a plain text button lost at the bottom of the "Free access" block and gave no reason at all to create an account. Stating the benefits explicitly makes the value obvious before asking for the commitment.

## 🔍 How

- `useCanPromoteApp` reuses the same display-mode heuristic as `src/index.jsx`'s `ensurePersistentStorage` gate (`display-mode: standalone` / `navigator.standalone`), plus a `document.referrer` check against `android-app://org.grottocenter.twa` to catch the Android TWA. The value is computed once via lazy `useState` since it cannot change during a session.
- The Play Store badge is loaded from Google's own CDN rather than a local copy: their branding guidelines require the asset to be used as published, not reproduced or restyled.
- `IosInstallPrompt` uses `react-ios-pwa-prompt` (new dependency). The library already self-gates on iOS + non-standalone and throttles reappearance through `localStorage`; only the copy is wired to `react-intl` so it stays translated. Configured with `promptOnVisit={1}` and `timesToShow={2}` to stay discreet.
- `JoinCommunity` renders nothing for authenticated users, so the home page stays unchanged once logged in.

## 🔗 Related API PR

None — front-end only, no API change involved.

## 🧪 Testing

- [ ] **Android / desktop**: the Play Store banner and the "Join the community" section should show on the home page when logged out; the banner should disappear when opening the site from the installed app (TWA), and "Join the community" should disappear once logged in.
- [ ] **iOS**: @ClemRz — could you kindly take this one on Safari? 🙏 I have no iOS device to check it on. What needs a look: the `Add to Home Screen` prompt appearing on the first visit, its copy being correctly translated, it not reappearing after being dismissed twice, and it staying hidden once the app is installed to the Home Screen.

## 📸 Previews

image

-------------

--------

image


urien (approved):
C'est très engageant,
Je trouve que c'est un moyen astucieux de présenter les services disponibles

ClemRz (changes requested):
Great addition overall — the intent is clear, the components are focused and well-scoped, i18n coverage is thorough across all 16 languages, and the PWA-detection heuristic mirrors the existing `ensurePersistentStorage` gate correctly. One real UX bug to address before merge.

### Issues (Must Fix)

- **[`AppPromoBanner.jsx` / `useCanPromoteApp.js`]** The Play Store banner is shown to iOS users. `useCanPromoteApp` only gates on "already running as installed PWA / TWA" — it has no platform check. An iOS Safari visitor who hasn't installed the app gets `canPromote = true` and therefore sees the Google Play badge, even though Google Play doesn't exist on iOS. The `IosInstallPrompt` (which self-gates on iOS) handles that side correctly, but the Play Store banner needs to be Android-only.

  `react-device-detect` is already in the project and `isIOS` is already used in `SideMenu/index.jsx`, so the fix is a one-liner in `AppPromoBanner`:
  ```suggestion
  import { isIOS } from 'react-device-detect';
  // …
  if (!canPromote || isIOS) return null;
  ```
  Alternatively, gate in `useCanPromoteApp` with a third condition and rename the hook if its semantics widen.

- **[`AppPromoBanner.jsx:48-52`]** `variant="h6" component="p"` renders at the h6 visual scale (14px/600) but is semantically a paragraph — the `component` override is correct and intentional. Consider `variant="subtitle1"` or `variant="body1"` with `fontWeight={600}` instead, since the h-scale variants carry heading semantics in the theme token naming even when overridden via `component`. Minor, but it avoids a reader having to think twice about why a heading variant is on a `<p>`.

- **[`JoinCommunity.jsx:72-82`, `Header.jsx:CTARow`, `Association.jsx:donate button`]** The "lift on hover" button style (`borderRadius: 999`, `boxShadow: theme.shadows[3]`, `transition: 'transform 0.15s ease, box-shadow 0.15s ease'`, `'&:hover': { transform: 'translateY(-2px)', boxShadow: theme.shadows[6] }`) is now defined in three places. Consider extracting it as a shared `styled` button (e.g. `AttractiveButton`) in a `homepage/shared.js` module to avoid drift when one instance needs updating.

- **[`AppPromoBanner.jsx`]** The Play Store badge is fetched from `https://play.google.com/intl/…` at render time. The code comment explains this is required by Google's branding guidelines, which is fine. Worth adding a brief note that this makes every page load issue an external HTTP request to Google's CDN, which may be blocked by aggressive privacy extensions or strict CSPs. A `loading="lazy"` attribute on the `` would at least defer the fetch until the banner scrolls into view on long-viewport devices.

### Nitpicks (Optional)

- **[`IosInstallPrompt.jsx:9`]** The component always renders `<PWAPrompt>` unconditionally. The library self-gates on iOS + non-standalone, so nothing is shown on other platforms — but it does execute and potentially read `localStorage` on every home-page render across all platforms. This is low-cost today; just document it with a brief comment so future readers know the iOS check is inside the library, not missing.

- **[`AppPromoBanner.jsx`]** `copySubtitle` in `IosInstallPrompt` is hardcoded `"grottocenter.org"` — this is intentional (it's a domain, not UI copy) but a short comment would help.

ClemRz (approved):
Second pass after b29f3694 — all six points from the previous review are addressed cleanly. The iOS/Play Store guard is in the right place, `AttractiveButton` is well-extracted, the CDN comment and `loading="lazy"` are there, and the `IosInstallPrompt` documentation is clear. Good work.

- **[`AppPromoBanner.jsx:53`]** `aria-label` on the `<Banner>` section is identical to the visible text already rendered inside it by the `<Typography component="p">`. The better pattern (already used in `JoinCommunity`, `LatestBlogNewsSection`, etc.) is to give the heading an `id` and use `aria-labelledby` — that avoids the duplication and keeps the accessible name and the visible text in sync automatically. Since the "heading" here is a `<p>` (correct — it's not structural), you'd just add `id="app-promo-title"` to it and switch to `aria-labelledby="app-promo-title"`.

- **[`useCanPromoteApp.js:8`]** The hook's return value is computed once via lazy `useState` (documented, intentional). Worth noting in the comment that `display-mode` can technically change mid-session on Chrome for Android (the browser hides/shows the browser frame when the user scrolls), but the once-computed value is still correct for this use-case because it represents the *launch* context, not the current chrome visibility. A brief note would prevent a future reader from questioning the `useState` vs `useMemo` choice.

### Nitpicks (Optional)

- **[`IosInstallPrompt.jsx`]** `react-ios-pwa-prompt` is added with a `^` range (`^2.0.6`), consistent with the rest of `package.json`. The library has a single maintainer and last published July 2024, so consider tracking it in `yarn outdated` checks if the iOS Safari install flow changes with future iOS releases.

- **[`index.jsx:62`]** `<IosInstallPrompt />` is placed outside `<Box component="main">` — this is correct given `PWAPrompt` renders a fixed overlay, so it shouldn't be inside the main landmark. A brief comment would save a future reader from wondering whether it was accidentally left outside.

PR #1509, opened 2026-08-12
fix(document): reset validation selection between batches

### LINKED ISSUE #1508: Le nombre de documents validés se cumule entre plusieurs séries de validations
## Décrire le bug
Dans le processus de validation des documents, le nombre annoncé de documents validés se cumule lorsqu'on procède à plusieurs séries de validations indépendantes.

## Comportement attendu
Le compteur doit revenir à zéro après chaque série de validations indépendante et annoncer uniquement le nombre de documents validés lors de la série en cours.

## Contexte
Le problème se produit lors de plusieurs séries successives de validations de documents.

Paul-AUB:
# fix(document): reset validation selection between batches

## 🤔 What

- Make the document validation table selection controllable by its parent.
- Reset desktop and mobile table selections between independent validation batches.
- Add regression tests for controlled selection on both table layouts.
- Closes #1508.

## 🤷‍♂️ Why

The validation page cleared its selected document IDs after a successful batch,
but the table retained an internal copy of the previous selection. Selecting a
document for the next batch therefore restored the old IDs and displayed a
cumulative document count.

## 🔍 How

`EntityTable` now accepts an optional `selectedIds` prop and forwards it to its
desktop and mobile implementations. When this prop is provided, the table uses
the parent-owned selection; existing callers without it keep the previous
internal-state behavior.

The document validation page passes its own `selectedIds`, so clearing that
state after a successful request also clears the table selection.

## 🔗 Related API PR

Not applicable.

## 🧪 Testing

- `yarn lint`
- `yarn test` — 106 test files and 970 tests passed
- Targeted controlled-selection tests for desktop and mobile layouts

## 📸 Previews

Not applicable; this fixes state synchronization without changing the visual
design.

### DIFF
--- a/packages/web-app/src/components/common/EntityTable/DesktopEntityTable.jsx
+++ b/packages/web-app/src/components/common/EntityTable/DesktopEntityTable.jsx
@@ -246,7 +246,7 @@ const DesktopEntityTable = ({
   setEntityColumns,
   isLoading,
   onSelected,
-  selectedIds,
+  selectedIds: controlledSelectedIds,
   pageRows,
   nbTotalRows,
   onRowClick,
@@ -275,7 +275,10 @@ const DesktopEntityTable = ({
   const [order, setOrder] = useState('');
   const [orderBy, setOrderBy] = useState('');
   const [internalSelectedIds, setInternalSelectedIds] = useState([]);
-  const selected = selectedIds ?? internalSelectedIds;
+  const selected = controlledSelectedIds ?? internalSelectedIds;
+  const isSelectionControlled = controlledSelectedIds !== undefined;
+  const onSelectedRef = useRef(onSelected);
+  onSelectedRef.current = onSelected;
   // Sticky results toolbar + sticky table header both pin to the same card
   // scroller; the header sits just below the toolbar, at its measured height
   // (see hook). 0 when the toolbar isn't rendered (shouldHideFooter).
@@ -347,7 +350,7 @@ const DesktopEntityTable = ({
     } else {
       newSelected.splice(selectedIndex, 1);
     }
-    if (selectedIds === undefined) setInternalSelectedIds(newSelected);
+    if (!isSelectionControlled) setInternalSelectedIds(newSelected);
     onSelected(newSelected);
   };
 
@@ -398,7 +401,7 @@ const DesktopEntityTable = ({
     if (event.target.checked) {
       newSelected = pageRows.map(n => n.id);
     }
-    if (selectedIds === undefined) setInternalSelectedIds(newSelected);
+    if (!isSelectionControlled) setInternalSelectedIds(newSelected);
     onSelected(newSelected);
   };
 
@@ -408,7 +411,10 @@ const DesktopEntityTable = ({
     setOrder('');
     setOrderBy('');
     setInternalSelectedIds([]);
-  }, [isNewQuery]);
+    if (isSelectionControlled && onSelectedRef.current) {
+      onSelectedRef.current([]);
+    }
+  }, [isNewQuery, isSelectionControlled]);
 
   const colSpan = visibleColumns.length + (onSelected ? 1 : 0);
 
--- a/packages/web-app/src/components/common/EntityTable/DesktopEntityTable.jsx
+++ b/packages/web-app/src/components/common/EntityTable/DesktopEntityTable.jsx
@@ -246,6 +246,7 @@ const DesktopEntityTable = ({
   setEntityColumns,
   isLoading,
   onSelected,
+  selectedIds,
   pageRows,
   nbTotalRows,
   onRowClick,
@@ -273,7 +274,8 @@ const DesktopEntityTable = ({
   );
   const [order, setOrder] = useState('');
   const [orderBy, setOrderBy] = useState('');
-  const [selected, setSelected] = useState([]);
+  const [internalSelectedIds, setInternalSelectedIds] = useState([]);
+  const selected = selectedIds ?? internalSelectedIds;
   // Sticky results toolbar + sticky table header both pin to the same card
   // scroller; the header sits just below the toolbar, at its measured height
   // (see hook). 0 when the toolbar isn't rendered (shouldHideFooter).
@@ -345,7 +347,7 @@ const DesktopEntityTable = ({
     } else {
       newSelected.splice(selectedIndex, 1);
     }
-    setSelected(newSelected);
+    if (selectedIds === undefined) setInternalSelectedIds(newSelected);
     onSelected(newSelected);
   };
 
@@ -396,7 +398,7 @@ const DesktopEntityTable = ({
     if (event.target.checked) {
       newSelected = pageRows.map(n => n.id);
     }
-    setSelected(newSelected);
+    if (selectedIds === undefined) setInternalSelectedIds(newSelected);
     onSelected(newSelected);
   };
 
@@ -405,7 +407,7 @@ const DesktopEntityTable = ({
     setPage(0);
     setOrder('');
     setOrderBy('');
-    setSelected([]);
+    setInternalSelectedIds([]);
   }, [isNewQuery]);
 
   const colSpan = visibleColumns.length + (onSelected ? 1 : 0);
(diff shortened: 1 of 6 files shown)


urien (approved):
Cela me semble très bien fonctionner

ClemRz (changes requested):
The fix is well-scoped and the controlled/uncontrolled pattern is cleanly applied. A few things worth addressing before merging.

### Issues (Must Fix)

- **[MobileEntityList.test.jsx]** The `IntlProvider` wrapper is used without a `messages` prop throughout the file. Per the project testing convention (see `testing.md` and the pattern in `DesktopEntityTable.test.jsx` and `ExportFormatDropdown.test.jsx`), translation keys rendered by the component under test must be explicitly listed in a `messages` object. Even though `<Translate>` falls back to `defaultMessage` and the test currently passes, an empty messages object produces `MISSING_TRANSLATION` console warnings and breaks the established pattern. The column label (`'Name'`) is rendered via `<Translate>` inside `MobileEntityCard` when body columns are present — and if a future test adds body columns without messages, it will silently produce noise. Add a `messages` object (at minimum `{ Name: 'Name' }`) and pass it via `messages={messages}`, consistent with every other test in the directory.

- **[DesktopEntityTable.jsx:407–411 / MobileEntityList.jsx:187–193]** When `isNewQuery` fires in controlled mode, `setInternalSelectedIds([])` is a no-op for the visible state (since `selected` reads from `selectedIds`), but the parent is never notified to clear its own `selectedIds`. This works correctly for the current consumer (`DocumentValidationPage`, which already calls `setSelectedIds([])` in `loadDocuments`), but it is a silent contract: any future consumer that wires up `selectedIds` without also clearing it on `isNewQuery` will end up with a stale visual selection. Consider whether `onSelected([])` should be called from within the `isNewQuery` effect when `selectedIds !== undefined`, or at least document the expectation in a comment.

- **[DesktopEntityTable.test.jsx:213–249 / MobileEntityList.test.jsx:29–59]** The regression tests assert that `onSelected` is called with the correct next selection (`[1]` then `[2]`), which validates the callback side. They do not assert the visual/controlled state — i.e., that the checkbox for `id: 1` is rendered as checked after `rerender(..., selectedIds: [1])`. Adding a `expect(checkboxes[...]).toBeChecked()` assertion after the first rerender would make the test a true end-to-end validation of the controlled rendering path, not just the callback path.

- **[MobileEntityList.jsx:221–236]** `handleToggle`'s deps include both `controlledSelectedIds` and `selectedIds`. Since `selectedIds = controlledSelectedIds ?? internalSelectedIds`, the two are not independent in controlled mode (`selectedIds` changes whenever `controlledSelectedIds` does). In controlled mode this is fine, but in uncontrolled mode `selectedIds` tracks `internalSelectedIds`, which is already updated by `setInternalSelectedIds` before the effect fires. Listing `selectedIds` directly (rather than the two constituents) and removing `controlledSelectedIds` as a separate dep would make the relationship clearer — though the current code is functionally correct.

### Nitpicks (Optional)

- **[MobileEntityList.test.jsx:1]** `import React from 'react'` is absent, which is consistent with the rest of the codebase (JSX transform). No action needed — just noting it is intentionally absent, unlike the main component file which still imports it.

- **[DesktopEntityTable.jsx:350 / DesktopEntityTable.jsx:401]** The guard `if (selectedIds === undefined)` is correct but asymmetric with `MobileEntityList`, which checks `if (controlledSelectedIds !== undefined)`. The semantics are identical; the naming difference (`selectedIds` vs `controlledSelectedIds` for the prop alias) is the cause. Both are fine, but if the files are ever refactored together, standardising the alias name would reduce the mismatch.

ClemRz (approved):
All issues from the previous review have been addressed. The controlled/uncontrolled pattern is cleanly applied in both table layouts, the test `messages` objects are properly populated, the visual/checked assertions are in place, and `onSelected([])` is correctly called from the `isNewQuery` effect when the parent owns the selection.

A few things still worth discussing before merging.

- **[MobileEntityList.jsx:217–221]** The `internalSelectedIds` effect still calls `onSelectedRef.current(internalSelectedIds)` whenever `internalSelectedIds` changes. In controlled mode, `handleToggle` short-circuits and calls `onSelected` directly without ever updating `internalSelectedIds`, so the effect is correctly idle. However, the code now has two notification paths (effect + direct call in `handleToggle`) that differ only by which array they send. This is safe today, but the asymmetry between the controlled and uncontrolled paths is non-obvious. Consider a comment like:

  ```javascript
  // Uncontrolled path only: controlled mode notifies via handleToggle directly
  // and never writes to internalSelectedIds, so this effect stays idle.
  useEffect(() => {
    if (!hasInteracted.current) return;
    if (onSelectedRef.current) onSelectedRef.current(internalSelectedIds);
  }, [internalSelectedIds]);
  ```

- **[DocumentValidation/index.jsx]** The `isNewQuery` prop is not wired up on `EntityTable` in this page, so the `isNewQuery`-triggered `onSelected([])` path added to both table implementations is never exercised by the only current consumer of controlled mode. The bug is genuinely fixed (via `setSelectedIds([])` in `loadDocuments`), but the `isNewQuery` callback path is untested end-to-end in a real page. This is fine as a building block, but worth flagging: any future consumer relying on `isNewQuery` to drive the reset (rather than clearing `selectedIds` directly) should be aware that both the prop update and the callback call will fire, and the parent must be idempotent against the double reset.

- **[DesktopEntityTable.jsx / MobileEntityList.jsx — `isNewQuery` effect]** `isSelectionControlled` is derived synchronously from `controlledSelectedIds !== undefined`, so it cannot change independently of the component re-rendering with a new prop. Including it in the `isNewQuery` effect's dependency array is technically correct but adds a dep that in practice only changes at the same time as the component's prop set. A short comment to that effect would help a future reader wondering whether the combined condition is guarding something subtle.

### Nitpicks (Optional)

- **[MobileEntityList.jsx:217–221 / DesktopEntityTable.jsx:409–417]** `isSelectionControlled` is in the `isNewQuery` effect deps, which is correct per exhaustive-deps rules. The linter would flag its absence. No action needed — just confirming the dep is there for the right reason, not by accident.

- **[MobileEntityList.test.jsx / DesktopEntityTable.test.jsx]** Both `isNewQuery` reset tests pass `selectedIds: [1]` but the component never renders a checkbox as checked (since `isNewQuery: true` fires the effect synchronously on mount before the first paint). Asserting `expect(onSelected).toHaveBeenCalledWith([])` is the correct check for this case. Fine as-is.

Paul-AUB:
✅ Fixed — clarified that the `internalSelectedIds` effect is the uncontrolled notification path; controlled mode notifies directly through `handleToggle` and leaves that effect idle.

⏭️ Not fixed — `DocumentValidationPage` continues to clear its controlled `selectedIds` directly in `loadDocuments`. Wiring `isNewQuery` there as well would add a redundant callback-driven reset without changing the behavior.

⏭️ Not fixed — no extra comment was added for the `isSelectionControlled` dependency. The effect reads that value, so keeping it in the dependency array is required and already self-explanatory; as the review notes, no action is needed.

Validation: both affected Vitest suites pass (15 tests), and `yarn lint` passes.

urien:
@BTHOM tu as 3 documents pour valider cette fonctionnalité

BTHOM:
C'est fait.Le bug est décrit dans un autre messageBTHEnvoyé depuis l'application Mail Orange----------------De : "Frédéric Urien" ***@***.***>Envoyé : lundi 17 août 2026 21:04À : GrottoCenter/grottocenter-front ***@***.***>Cc : BTHOM ***@***.***>; Mention ***@***.***>Objet : Re: [GrottoCenter/grottocenter-front] fix(document): reset validation selection between batches (PR #1509)urien left a comment ***@***.*** tu as 3 documents pour valider cette fonctionnalité—Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today! You are receiving this because you were mentioned.Message ID: ***@***.***>

Paul-AUB:
Je remet ici la procédure de reproduction de bug :

- Sur trois documents test qui étaient à valider, j'en ai d'abord validé 2 en même temps ; le message affiché sur le bouton vert était correct, à savoir "Valider 2 documents".
- J'ai ensuite validé le troisième document ; le message affiché sur le bouton vert était incorrect, à savoir "Valider 3 documents" (alors que l'affichage correct aurait dû être "Valider 1 document").
- En outre, aucun message indiquant le succès de chaque validation ne s'affiche ; les documents concernés disparaissent simplement de la liste, sans que la validation ne soit confirmée.

Le bug est confirmé en prod.

J'ai reproduit la séquence sur cet environement de staging avec la correction, et le bug est corrigé : 






Concernant le dernier point (absence de message de succès en cas de validation), j'ouvre https://github.com/GrottoCenter/grottocenter-front/issues/1518 pour le corriger plus tard afin de libérer des places en staging.

PR #1512, opened 2026-08-13
refactor(dropzone): unify file selection

Paul-AUB:
# refactor(dropzone): unify file selection

## 🤔 What

- Replace `react-dropzone-uploader` with headless `react-dropzone` encapsulated by `FileSelectorInput`.
- Centralize file type, size, count, disabled-state, and rejection handling while preserving the existing MUI and responsive UI.
- Migrate document attachments, geometry imports, and CSV imports to the shared selector.
- Keep Papa Parse for CSV parsing and preserve CSV import reset, validation, replacement, and removal behavior.
- Remove the unavailable organization-logo uploader.
- Add React Testing Library coverage and Storybook states for the shared selector and CSV flow.

Closes #1511.

## 🤷‍♂️ Why

The application had multiple file-selection implementations with inconsistent validation between the native picker and drag-and-drop. Centralizing generic selection behavior prevents invalid dropped files from bypassing constraints and gives current and future upload workflows one reusable UI contract.

## 🔍 How

- Wrap `useDropzone` inside `FileSelectorInput` and normalize library rejection codes before passing them to feature code.
- Keep the selected file list controlled by callers rather than duplicating it inside the dropzone.
- Build document acceptance rules from backend MIME types and extensions, and enforce the configured maximum upload size.
- Parse accepted CSV file text through `usePapaParse` with trimmed headers, headers enabled, and empty lines skipped.
- Surface generic geometry rejections through the existing translated parse-error UI.

## 🔗 Related API PR

None. This refactor does not require an API change.

## 🧪 Testing

- `yarn lint`
- `yarn test` — 107 files and 983 tests passed
- `yarn build`
- `yarn translations:check-sorted`
- `yarn translations:check-en`
- `yarn node scripts/translations/sync-with-en.js --check`

Manual verification can cover picker, keyboard, accepted/rejected drag-and-drop, responsive layout, and controlled file removal through the new `FileSelectorInput` Storybook stories.

## 📸 Previews

No screenshots included. The visual treatment is documented by the empty, selected-files, disabled, rejected, and single-file Storybook states.


ClemRz (changes requested):
Solid refactoring overall — the encapsulation goal is well-achieved, the test coverage is good, and the generation-counter pattern in `Step2` is a thoughtful way to handle rapid file replacement. A few things worth addressing before merge.

### Issues (Must Fix)

- **[Step2.jsx:87]** `setSelectedFile(file)` is called inside the `readString` `complete` callback after a stale-generation guard, but it uses the closure's `file` variable, not the guarded parse result. Because `readString` is synchronous when `complete` is provided (PapaParse calls `complete` immediately on the same tick as `readString`), this is safe **today**. However, if the guard `if (parseGeneration !== parseGenerationRef.current) return;` fires, the state is correctly skipped — but if it **doesn't** fire, `setSelectedFile(file)` should be inside the guard block alongside `setRowErrors`, `updateAttribute('importData', …)` and `updateAttribute('fileImported', …)`. Right now `setSelectedFile` sits _outside_ the inner stale-check, meaning a file that was superseded between `file.text()` resolving and `readString` completing could still set `selectedFile` to the old file object. Move `setSelectedFile(file)` inside the `if (parseGeneration !== parseGenerationRef.current) return;` check (i.e., after the guard, alongside the other state updates):
  ```suggestion
        complete: results => {
          if (parseGeneration !== parseGenerationRef.current) return;
          const errors = [
            ...results.errors.map(e => ({
              errorMessage: `Import error ${e.message}`,
              row: e.row + 2
            })),
            ...checkData(results.data, selectedType, formatMessage)
          ];
          const isValid = errors.length === 0;
          setSelectedFile(file);
          setRowErrors(errors);
          updateAttribute('importData', isValid ? results.data : undefined);
          updateAttribute('fileImported', isValid);
        }
  ```
  *(The suggestion is identical to the current code — the fix is to confirm this is intentional and add a comment explaining why `setSelectedFile` is safe here given PapaParse's sync dispatch. Without that comment the placement looks like a bug.)*

- **[AddFileForm/index.jsx:228]** The `handleRejections` fallthrough case maps any rejection that is neither `TOO_LARGE` nor `TYPE_NOT_ACCEPTED` (i.e., `TOO_MANY_FILES`, `TOO_SMALL`, or `UNKNOWN`) to the `'error on file name'` i18n key with the message *"The following file name is invalid: {file}."* That text is factually wrong for a `TOO_MANY_FILES` rejection and will confuse users. Add a branch for `TOO_MANY_FILES`, or change the fallback message to a generic `'This file was rejected.'`/`'error on file type'`-style key that is not filename-specific:
  ```suggestion
      if (reasons.includes(REJECTION_REASONS.TOO_MANY_FILES)) {
        return formatMessage(
          {
            id: 'This file was rejected.',
            defaultMessage: 'This file was rejected: {file}.'
          },
          { file: fileName }
        );
      }
      return formatMessage(
        {
          id: 'error on file name',
          defaultMessage: 'The following file name is invalid : {file}.'
        },
        { file: fileName }
      );
  ```

- **[Step2.jsx:51-55]** The `eslint-disable-next-line react-hooks/exhaustive-deps` suppression on the mount-only `useEffect` lacks a written justification comment, which violates the project rule in `AGENTS.md` ("Never use `// eslint-disable` without a written justification in the preceding comment"). Consider adding a comment like:
  ```suggestion
    // Run only on mount: intentionally omit updateAttribute and dispatch from
    // deps to avoid re-running on every render. These are stable callbacks but
    // are not memoised at the call site, so listing them would restart the
    // effect on every render and repeatedly wipe import state mid-session.
    // eslint-disable-next-line react-hooks/exhaustive-deps
  ```

- **[ShapefileImport.jsx:30]** `ACCEPTED_MIME` is still a comma-separated extension string (`.geojson,.json,.zip`), which `normalizeAccept` converts into `{ 'application/octet-stream': ['.geojson', '.json', '.zip'] }`. `application/octet-stream` is itself a valid MIME type (generic binary), so react-dropzone's `attr-accept` will accept **any** file whose `file.type` is `application/octet-stream` regardless of extension — meaning a plain binary with that MIME could slip through on some browsers. Consider converting `ACCEPTED_MIME` to the object form to make the intent explicit and avoid the catch-all MIME key:
  ```suggestion
  const ACCEPTED_MIME = {
    'application/geo+json': ['.geojson', '.json'],
    'application/zip': ['.zip'],
    'application/x-zip-compressed': ['.zip']
  };
  ```

- **[FileSelectorInput/index.jsx:186-204]** On mobile the `inputEl` (`<input {...getInputProps()} />`) is rendered as a direct child of `<Stack>`, outside any element that has `getRootProps()` applied. The `open()` call on the Button works around this, but dropping a file onto the mobile layout has no `onDrop` handler attached (no `getRootProps` means no react-dropzone event binding), so drag-and-drop is silently unavailable on mobile. This may be intentional (mobile UX often skips D&D), but it's worth either documenting or wrapping the Stack with `getRootProps` so behavior is consistent.

- **[Step2.unit.test.jsx:210-239]** The `'returning to the step resets…'` test calls `waitFor` after the first render to confirm that `updateAttribute('fileImported', false)` is called, but then re-renders the same `ctx` object. The `updateAttribute` mock was already called on mount with `'importData', undefined` and `'fileImported', false`, so the second render's assertions (`toHaveBeenCalledWith('importData', undefined)`) are trivially satisfied by the first render's calls if `mockClear` is skipped — consider clearing `ctx.updateAttribute` before the second render as well (as the `removing the file` test correctly does).

### Nitpicks (Optional)

- **[Step2.jsx:18-27]** `formatBytes` re-implements a standard size formatter. It's clean and correct, but `bytes <= 0` returning `'0 B'` would render `'0 B'` for a 0-byte file that might need a different message. Minor edge case, fine to leave as-is.

- **[FileSelectorInput/index.jsx:21-27]** The `REJECTION_REASONS` values (`'file-too-large'`, etc.) intentionally differ from react-dropzone's own codes (`'file-too-large'` is the same, `'file-invalid-type'` maps to `'file-type-not-accepted'`), which is the point of the abstraction. A one-line comment noting "values are intentionally different from react-dropzone's `e.code` strings — see `DROPZONE_REJECTION_MAP`" would make the design decision visible to the next reader without having to cross-reference.

- **[Step2.jsx:116]** `onFilesAdd={([file]) => file && parseFile(file)}` — the guard `file && …` is sensible but `react-dropzone` already only calls `onDrop(accepted, …)` when `accepted.length > 0`, so the `file &&` check is dead code. Fine to keep for safety, but worth knowing.

Paul-AUB:
Addressed in `c6920db6`.

- ✅ Fixed — clarified in `Step2` that `selectedFile` and parsed import state are committed under the same generation guard. No move was needed: the early return already guards every following state update, independently of PapaParse's callback timing.
- ✅ Fixed — changed the document upload fallback to the existing generic `This file was rejected.` translation instead of incorrectly reporting a filename error for `TOO_MANY_FILES`, `TOO_SMALL`, or unknown rejections.
- ✅ Fixed — added an explicit justification immediately before the `react-hooks/exhaustive-deps` suppression for the mount-only import reset.
- ✅ Fixed — changed the shapefile accept configuration to explicit GeoJSON and ZIP MIME mappings, removing the `application/octet-stream` catch-all for this caller.
- ✅ Fixed — documented that the mobile file selector is intentionally picker-only and therefore omits desktop drag-and-drop handlers.
- ⏭️ Not fixed — the remount test already clears `ctx.updateAttribute` after unmounting and before the second render, so its assertions cannot be satisfied by calls from the first mount.
- ⏭️ Not fixed — `formatBytes(0)` remains `0 B`; that is an accurate size display, while empty-file validity belongs to rejection or parsing behavior.
- ✅ Fixed — documented that the public rejection-reason values intentionally differ from react-dropzone's codes and are mapped at the component boundary.
- ⏭️ Not fixed — retained the defensive `file && parseFile(file)` guard; it is harmless and protects the callback if it is ever invoked directly with an empty array.

Validation: `yarn lint` (0 errors, 0 warnings) and the focused `Step2`/`FileSelectorInput` Vitest suite (15 tests passed).

ClemRz (changes requested):
Good follow-up — all the must-fix items from the first review have been addressed cleanly, and the `ShapefileImport` MIME fix and the eslint-disable justification are exactly right. The generation-counter explanation in the comment is clear. A few new things came up during this pass.

### Issues (Must Fix)

- **[AddFileForm/index.jsx:188–229]** In `AddFileForm`, `onFilesAdd` and `onFileRejections` are called sequentially from the same react-dropzone `onDrop` callback. In React 18 (automatic batching), both `setErrors` calls — one inside `updateFiles` and one inside `handleRejections` — are batched within the same event. Because `handleRejections` runs last, it always overwrites the `setErrors` call from `updateFiles`. In a mixed drop (some files accepted, some rejected), any domain errors produced by `validateAndBuildFileEntries` (duplicate filenames, missing extensions) for the accepted files are silently discarded, and only the rejection messages survive. Consider merging both error sets rather than overwriting:
  ```suggestion
  const handleRejections = rejections => {
    const messages = rejections.map(({ fileName, reasons }) => {
      if (reasons.includes(REJECTION_REASONS.TOO_LARGE)) {
        return formatMessage(
          {
            id: 'error on file size',
            defaultMessage:
              'The following file is too big: {file}. Max accepted size: {maxSize}'
          },
          {
            file: fileName,
            maxSize: `${MAX_SIZE_OF_UPLOADED_FILES / 1000000} Mo`
          }
        );
      }
      if (reasons.includes(REJECTION_REASONS.TYPE_NOT_ACCEPTED)) {
        return formatMessage(
          {
            id: 'error on file type',
            defaultMessage: 'The file type is not accepted: {file}.'
          },
          { file: fileName }
        );
      }
      return formatMessage({
        id: 'This file was rejected.',
        defaultMessage: 'This file was rejected.'
      });
    });
    setErrors(prev => [...prev, ...messages]);
  };
  ```

- **[FileSelectorInput.unit.test.jsx:15–19, Step2.unit.test.jsx:37–40]** Both new test files import the full `en.json` and suppress `MISSING_TRANSLATION` via `onError` rather than explicitly listing the keys the component uses. The project convention (see `testing.md` and `DetachEntranceSection.test.jsx`) is to define only the keys the component under test actually renders. Using the full file masks missing-key regressions (a new `formatMessage` call with a typo'd key will silently fall through instead of failing the test). Consider switching to explicit key maps for the keys rendered by each component. The full-file approach isn't wrong in absolute terms — it's just not the convention here.

- **[FileSelectorInput/index.jsx:152–161]** The `extensions` prop is display-only (rendered as the hint label), but it is not fed into `getInputProps`/`accept`. A caller can pass `accept={{ 'application/pdf': ['.pdf'] }}` and `extensions={['.csv']}` and the hint will lie about what the picker accepts. Consider at least a PropTypes comment noting that `extensions` must reflect `accept`, or derive the displayed extensions from `accept` directly (when it is in object form). This won't cause a runtime error but is an easy source of confusion in future integrations.

### Nitpicks (Optional)

- **[Step2.jsx:17–27]** `formatBytes` is local to `Step2` and duplicates a common pattern. If a shared `formatBytes` utility doesn't exist in `src/utils/`, moving it there would make it available to other components that need to display file sizes (e.g. future upload progress). Fine as-is for now.

- **[Step2.jsx:18]** `!Number.isFinite(bytes) || bytes <= 0` returns `'0 B'` for a genuine 0-byte file. A 0-byte file would likely fail CSV parsing anyway, so this edge case is harmless in practice. Just worth noting if this utility ever gets extracted and reused in other contexts.

- **[FileSelectorInput/index.jsx:37–61]** The `normalizeAccept` comment mentions `application/octet-stream` is used intentionally for extension-only entries in the legacy string form (Therion formats). Since `AddFileForm` still builds a comma-separated string and passes it as `accept`, `normalizeAccept` will produce `application/octet-stream: ['.th', '.th2', …]` for those extensions. This is explicitly acknowledged in the comment and is the intended behavior — just flagging it so it doesn't resurface in a future review.

Paul-AUB:
All fixes applied and verified. Summary:

  Must Fix — AddFileForm setErrors overwrite (AddFileForm/index.jsx)
  Added a useRef flag that updateFiles sets and clears via queueMicrotask. handleRejections now uses a functional setErrors and appends to the accepted-side errors when the flag is set (mixed drop)
  or replaces them otherwise (rejections-only drop). Domain errors on accepted files are no longer clobbered.

  Suggestion #1 — explicit i18n key maps in tests (FileSelectorInput.unit.test.jsx, Step2.unit.test.jsx)
  Dropped the full-en.json import and MISSING_TRANSLATION suppression. Each test file now declares only the keys its component actually renders — matching the DetachEntranceSection.test.jsx /
  UploadStep.test.jsx convention — so a typo'd formatMessage id fails loudly.

  Suggestion #2 — derive extension hint from accept (FileSelectorInput/index.jsx)
  Added deriveExtensionsFromAccept and a memoized displayedExtensions that falls back to the values inside acceptObject when the caller didn't pass extensions. Callers using object-form accept no
  longer risk a hint that lies about what the picker accepts.

  Nitpick #1 & #2 — extracted formatBytes utility (utils/formatBytes.js, utils/formatBytes.unit.test.js, Step2.jsx)
  Moved the local formatBytes into src/utils/formatBytes.js and tightened the guards: a genuine 0 renders '0 B', invalid inputs (NaN / negative / undefined / null) render ''. Added unit tests for the
  utility. Step2 now imports the shared version.

  Tests run: 13/13 pass for FileSelectorInput.unit.test.jsx and formatBytes.unit.test.js. Step2.unit.test.jsx fails to load due to a pre-existing missing isomorphic-fetch module (reproduces on HEAD
  before my changes) — unrelated.

Paul-AUB:
Deployed in staging

urien:
Pour l'ajout de fichiers a un document j'ai cela
image
Et cela fonctionne > OK

Pur les données issues de capteur j'ai cela
Pas de Drag and Drop,
image

Pour l'import d'une géométrie c'est OK
Pour l'import CSV c'est OK

Paul-AUB:
Dropzone également implémentée désormais sur les observations 
image

urien (approved):
Parfait
Merci @Paul-AUB

ClemRz (approved):
The encapsulation goal is fully achieved, all previous must-fix items have been addressed cleanly, and the CI pipeline is green. A few small things worth calling out.

- **[UploadStep.jsx:254–259]** `handleFileRejections` fires a generic `'This file was rejected.'` toast for every rejection regardless of reason. `ACCEPTED_FILE_TYPES` covers three text-based MIME types so the only realistic rejection is a wrong type — a message like `'Only .csv, .tsv, or .txt files are accepted.'` would be more actionable for users. This is the one migration that skipped the type-vs-size discrimination the other callers implement.

- **[FileSelectorInput.stories.jsx:68, 74, 84, 94, 101, 105]** The Storybook stories pass `extensions` without leading dots (`['pdf', 'txt']`), while every production caller passes dotted extensions (`['.csv']`, `['.geojson', '.json', '.zip']`). The hint label in the stories renders "csv, pdf, txt" instead of ".csv, .pdf, .txt", so the stories don't accurately represent the production rendering. Consider aligning the story fixtures with the dotted convention.

### Nitpicks (Optional)

- **[Step2.unit.test.jsx:202]** The "removing the file" test locates the chip via `screen.getByText('good.csv').closest('.MuiChip-root')`. Coupling to a MUI class name is fragile in unit tests just as in Cypress tests — a MUI major bump that renames the class would silently break the selector. Adding a `data-testid` to the chip, or using `getByRole('button', { name: /good\.csv/ })` on the delete button directly, would make it resilient.

- **[formatBytes.js:2–3]** The comment says "binary units (KiB thresholds)" but the labels emitted are `KB`/`MB`/`GB` (SI notation). The 1024 threshold with SI suffix is the de-facto convention most browsers use, so this is fine in practice — the comment just creates a mild terminology mismatch. Saying "uses 1024-byte thresholds but displays SI labels (KB/MB/GB)" would be more precise.

- **[UploadStep.jsx:433]** The `files` prop passed to `FileSelectorInput` is `file ? [{ fileName: file.name }] : []`. The chip therefore shows only the filename with no size hint (unlike `Step2` which renders `formatBytes(selectedFile.size)` below the selector). This is intentional given the different layouts, but worth noting for consistency if someone adds size display to this flow later.

PR #1515, opened 2026-08-14
feat(admin): add role-based site preview

### LINKED ISSUE #1513: Admin: "View site as another role" mode with a floating, dismissible top overlay
# Admin: "View site as another role" mode with a floating, dismissible top overlay

As an administrator, I need to verify what other users actually see on
Grottocenter — which buttons, menus, pages, and actions are exposed to a
Moderator, a Leader, a regular User, or a non-authenticated visitor. Today I
can only guess by reading the code or by logging in as a test account, which
is slow, error-prone, and does not scale when we ship UI changes that touch
role-gated surfaces (admin menus, moderation queues, sensitivity controls,
private routes, etc.).

I want a low-friction way to preview the site from any other role without
logging out of my admin session.

A **"View site as" mode** that an admin can activate from the **Management
Tools** page. Once activated, a **floating, dismissible overlay banner**
(`position: fixed`) is pinned to the top of the viewport, above the AppBar,
on every page and every route, until the admin explicitly closes it. The
banner is the source of truth for the current viewing mode: it shows the
impersonated role and lets the admin switch it or stop the mode without
going back to the Management Tools page.

Behavior:

- **Reserved to administrators.** Non-admins never see the entry point or
  the banner.
- **Activation UI** lives on the Management Tools page: a small panel to
  pick a role and start the mode.
- **Impersonatable roles**: Moderator, Leader, User, **Not logged in**
  (i.e. anonymous visitor).
- **The banner floats on top** with `position: fixed`, stays anchored to
  the top of the viewport while the page scrolls, and covers/pushes the
  AppBar rather than living inside the normal content flow.
- **The banner controls the mode**: it displays the current viewing role,
  and offers "Switch role" and "Stop viewing as…" controls.
- **Persistence across navigation and refreshes**: the mode survives route
  changes and page refreshes in the same browser tab. It ends only when
  the admin dismisses the banner, logs out, or closes the tab.
- **Effect on the UI**: role-based gates (menus, admin/moderation entry
  points, private routes) reflect the impersonated role. Selecting
  "Not logged in" behaves as if no session existed — login button back in
  the AppBar, private routes blocked, personal identity hidden from the
  UI.
- **Limitation to make explicit in the banner or a tooltip**: API
  responses still come back with the admin's real privileges. This is a
  UI-visibility preview, not a permissions sandbox.

**ASCII mockups**

Entry point on the Management Tools page:

```
┌─ Management Tools ─────────────────────────────────────────┐
│  [Manage users]  [Pending documents]  [Duplicates]  ...    │
│                                                            │
│  ┌─ Preview the site as another role ─────────────────┐    │
│  │                                                    │    │
│  │  Role:  [ Moderator            ▼ ]                 │    │
│  │         Moderator                                  │    │
│  │         Leader                                     │    │
│  │         User                                       │    │
│  │         Not logged in                              │    │
│  │                                                    │    │
│  │  [  Start viewing as this role  ]                  │    │
│  └────────────────────────────────────────────────────┘    │
└────────────────────────────────────────────────────────────┘
```

Floating overlay while active (position: fixed at the top of the
viewport, above the AppBar; visible on every route, on every scroll
position, until dismissed):

```
   ╔════════════════════════════════════════════════════════════╗ ← position: fixed
   ║ 👁  Viewing site as MODERATOR   [ Switch ▼ ]   [ Stop ] ✕ ║   top: 0, full width,
   ╚════════════════════════════════════════════════════════════╝   z-index above AppBar
┌──┴─────────────────────────────────────────────────────────────┐
│ [ Grottocenter AppBar — sits below the floating banner ]       │
├────────────┬───────────────────────────────────────────────────┤
│            │                                                   │
│  Side      │   Page content, rendered as a Moderator would     │
│  menu      │   see it (admin-only entries hidden, moderation   │
│            │   queues visible, etc.)                           │
│            │                                                   │
└────────────┴───────────────────────────────────────────────────┘
```

"Not logged in" variant:

```
   ╔════════════════════════════════════════════════════════════╗
   ║ 👁  Viewing site as an ANONYMOUS VISITOR       [ Stop ] ✕ ║
   ╚════════════════════════════════════════════════════════════╝
┌──┴─────────────────────────────────────────────────────────────┐
│ [ AppBar shows "Log in" button, no avatar ]                    │
│ [ Private routes redirect to the "must be authenticated" page ]│
└────────────────────────────────────────────────────────────────┘
```

**Acceptance criteria**

- [ ] Only administrators can access the entry point on the Management
      Tools page and see the overlay banner.
- [ ] From the Management Tools page, the admin can start the preview by
      picking one of: Moderator, Leader, User, Not logged in.
- [ ] Once started, the banner floats at the top of the viewport with
      `position: fixed`, above the AppBar, on every route.
- [ ] The banner stays anchored to the top of the viewport while the
      page scrolls.
- [ ] The banner displays the current impersonated role in a way that is
      unambiguously distinct from normal browsing (color, icon).
- [ ] From the banner, the admin can switch to another role without
      navigating back to the Management Tools page.
- [ ] From the banner, the admin can stop the preview with a single
      click.
- [ ] The banner and the mode survive route changes and full page
      reloads inside the same tab.
- [ ] Closing the tab or logging out ends the mode.
- [ ] Selecting "Not logged in" produces a UI equivalent to a logged-out
      session (login button, private routes blocked, no personal
      identity shown).
- [ ] Role-gated UI (admin menus, moderation entry points, private
      routes, permission-gated queries) reflects the impersonated role
      while the mode is active.
- [ ] The banner or a tooltip on it makes clear that this is a UI
      preview, not a permissions sandbox — API responses still reflect
      the admin's real access.

**Describe alternatives considered**

- **Real backend impersonation** (a dedicated endpoint issuing a scoped
  token for the target user or role). More faithful but requires
  backend work and a security review. Out of scope for now; this issue
  is deliberately UI-only.
- **Test accounts, one per role.** Works but requires logging out/in,
  wiping state, and does not scale to reviewing PRs that touch
  role-gated UI on many pages.
- **Floating action button (FAB).** Discreet but easy to forget; a
  floating top banner is harder to miss, which matches the intent —
  make the mode unmistakable while active.

**Additional context**

Related helpers in the code that already scaffold role checks and are
the natural integration points: `src/helpers/AuthHelper.js` (`hasRole`),
`src/hooks/usePermissions.js`, `src/hooks/useUserProperties.js`,
`src/reducers/LoginReducer.js`, and the existing Management Tools entry
at `src/pages/Admin/ManageUsers.jsx` / `src/pages/Dashboard.jsx`.

Paul-AUB:
## 🤔 What

- Add an administrator-only launcher to preview Grottocenter as a Moderator,
  Leader, User, or anonymous visitor.
- Show a fixed, unmistakable indicator on every route with controls to switch
  role or stop the preview.
- Persist the selected role for the current browser tab and clear it on logout
  or login.
- Apply the effective role through the shared permission and user-property
  hooks, including logged-out UI behavior for the anonymous preview.
- Add translations for all supported languages, unit coverage, a Storybook
  story, and a Cypress navigation/refresh scenario.

Closes #1513.

## 🤷‍♂️ Why

Administrators need a quick way to verify role-gated menus, routes, actions,
and identity surfaces without repeatedly logging into separate test accounts.

## 🔍 How

The selected preview role is stored in `sessionStorage` and exposed through the
login reducer. Shared authorization helpers replace the real administrator
roles with the selected effective role while retaining a separate
`isRealAdmin` check for the launcher and indicator. Stale or forged preview
state is ignored for non-administrators, and unsupported roles are rejected.

The anonymous option makes `isAuth` false and hides user properties while the
real token remains available for API calls. The indicator includes an
accessible warning that this is a UI preview only and that API responses still
use the administrator's real privileges.

## 🔗 Related API PR

None. This feature is deliberately UI-only.

## 🧪 Testing

- `yarn lint`
- `yarn test` — 107 files, 977 tests passed
- `yarn build`
- `yarn translations:sync-with-en`
- `yarn translations:check-sorted`
- Added `cypress/e2e/impersonation.cy.js` for role switching, cross-route
  persistence, reload persistence, anonymous UI, and stopping the preview.
  Local execution was blocked before test startup by a broken cached Cypress
  15.19.0 executable (`bad option: --smoke-test` / `--ping`); no spec failure
  was reported.

Manual verification:

1. Log in as an administrator and open Management Tools.
2. Start a preview for each available role.
3. Navigate to another route and reload the page; the indicator and effective
   permissions should remain active.
4. Switch role from the indicator, then stop the preview with one click.

## 📸 Previews

Not included.


ClemRz (changes requested):
Overall this is a well-structured feature. The architecture is clean: impersonation state lives in the login slice, the security gate is applied once in `AuthHelper.hasRole` with an escape hatch via `{ ignoreImpersonation: true }`, and the UI surfaces (launcher + indicator) compose from that single source of truth. Good test coverage across unit, property, and Cypress layers. A few things need attention before merge.

### Issues (Must Fix)

- **[`packages/web-app/src/pages/Dashboard.jsx:229-234`]** The Dashboard's access guard redirects to `/` when `isAdmin || isModerator || isLeader` are all false. An admin impersonating "User" (or "Anonymous") loses all three flags via `hasRole`, satisfying `!hasDashboardAccess`, and gets immediately bounced to the homepage — right from the very page that hosts the launcher. The launcher then disappears from under their feet and they can no longer stop the preview from the Management Tools page (they would have to use the indicator pill instead, which still works, but the UX is jarring). Consider including `isRealAdmin` in the access check:
  ```suggestion
  const hasDashboardAccess =
    permissions.isAdmin || permissions.isModerator || permissions.isLeader || permissions.isRealAdmin;
  ```

- **[`packages/web-app/src/components/common/ImpersonationIndicator.stories.jsx:7`]** The story imports `loginReducer` with a relative path (`../../reducers/LoginReducer`) while the two new components use the `@/` alias for all internal imports. Inconsistency aside, it also bypasses the project convention stated in `AGENTS.md`:
  > All new code must use [`@/` alias] for imports that would otherwise require two or more `../` segments.
  ```suggestion
  import loginReducer from '@/reducers/LoginReducer';
  import ImpersonationIndicator from './ImpersonationIndicator';
  ```

- **[`packages/web-app/src/components/appli/ImpersonationLauncher.jsx:91-94`]** The role `<Select>` is disabled while `isImpersonating` is true. That means an admin who lands back on the Dashboard during an active preview session can't see which role they originally chose — the select is greyed-out and shows the *local* default `'Moderator'`, not the active impersonated role. Consider seeding `selectedRole` from `impersonatedRole ?? 'Moderator'` (read from `usePermissions()`) so the picker reflects the live state even when disabled.

- **[`packages/web-app/src/reducers/LoginReducer.property.test.js:160-210`]** The new impersonation tests are unit tests written inside the property-test file. That's fine given the existing file structure, but none of them use `fc.property`. For a file named `*.property.test.js` the convention is property-based testing throughout. Consider promoting the impersonation tests to proper `fc.property` properties — e.g. asserting `SET_IMPERSONATED_ROLE` with an arbitrary from `IMPERSONATABLE_ROLES` persists correctly, and an arbitrary *outside* the allowed set is always a no-op.

- **[`packages/web-app/cypress/e2e/impersonation.cy.js:28`]** The second test starts a preview with `cy.get('[data-testid="impersonation-start-button"]').click()` without first selecting a role, so it uses whatever the Select's default is. This implicitly tests Moderator. Making the selection explicit (as the first test does) makes the test self-documenting and more resilient if the default ever changes.

- **[`packages/web-app/src/components/common/ImpersonationIndicator.jsx:109-115`]** `aria-live="polite"` is correct for announcing state changes, but `role="status"` already implies `aria-live="polite"` per the ARIA spec — the two are redundant here. Either remove the explicit `aria-live` or change `role` to `region` with an `aria-label` to get a named landmark. Minor accessibility nitpick.

### Nitpicks (Optional)

- **[`packages/web-app/public/lang/en.json:1258`]** The English description reads: `"This only affects what the UI shows, be careful API responses still reflect your real access."` — "be careful" feels mid-sentence and informal. Consider: `"This only affects what the UI shows. API responses still reflect your real access."` (the other language files have a semicolon or period here, so the English one is the odd one out).

- **[`packages/web-app/src/components/appli/ImpersonationLauncher.jsx:78`]** The `defaultMessage` for `Impersonation launcher description` in the component (`'Be careful API responses still reflect your real access'`) doesn't match what's in `en.json` (`"This only affects what the UI shows, be careful API responses still reflect your real access."`). The default is only shown when the i18n key is absent, so this is cosmetic, but keeping them in sync avoids confusion during development.

Paul-AUB:
✅ Fixed — Dashboard access now preserves a real administrator's live preview, including Anonymous mode, while expired sessions still redirect. The launcher remains available on the dashboard, and the Manage users card continues to follow the effective previewed role.

✅ Fixed — `ImpersonationIndicator.stories.jsx` now imports `LoginReducer` through the `@/` alias.

✅ Fixed — The disabled launcher picker now displays `impersonatedRole` while a preview is active, including role changes made from the indicator.

✅ Fixed — The impersonation reducer scenarios now use `fc.property` with arbitraries for supported roles, unsupported roles, prior state, and preview clearing.

✅ Fixed — The banner-switch Cypress test explicitly selects Moderator before starting the preview. The Anonymous test also verifies that the dashboard remains open and the launcher reflects the active role.

✅ Fixed — Removed the redundant `aria-live="polite"`; `role="status"` retains the intended polite announcement behavior.

✅ Fixed — Updated the English launcher description to use two sentences.

✅ Fixed — Synchronized the launcher's `defaultMessage` with `en.json` and the indicator fallback.

ClemRz (approved):
All five issues and three nitpicks from the previous round are addressed. The follow-up commit fixes every item cleanly and the implementation is now in good shape. A few minor observations remain below, none of which block merge.

- **[`packages/web-app/src/reducers/LoginReducer.property.test.js:130-162`]** The "Property 9" describe block correctly covers `LOGOUT`, but the expanded state shape (which now includes `impersonatedRole`) is not asserted there. Consider adding `expect(nextState.impersonatedRole).toBeNull()` to the existing `LOGOUT` properties — the new "Impersonation state" block covers `CLEAR_IMPERSONATION` and `FETCH_LOGIN_SUCCESS`, but a direct property for `LOGOUT → impersonatedRole = null` would close the loop and keep the property suite exhaustive. Tiny addition:
  ```suggestion
  expect(nextState.authTokenDecoded).toBeNull();
  expect(nextState.impersonatedRole).toBeNull();
  ```

- **[`packages/web-app/src/hooks/useUserProperties.js:8-10`]** The Anonymous guard only fires for non-Anonymous impersonation modes when the admin has the `Anonymous` role. For other impersonated roles (Moderator, Leader, User) the hook still returns the real admin's decoded token, so ownership/self-checks (`userId.toString() === comment.author?.id.toString()`) continue to match the real admin's identity. This is the documented limitation ("UI preview only") and is clearly communicated in the banner, but a comment here would help future maintainers understand why only `Anonymous` is special-cased in this hook:
  ```suggestion
  // Only Anonymous collapses identity — other impersonated roles keep the
  // real token so that personal content (my documents, my profile) still
  // loads correctly. This is intentional: the feature is a UI visibility
  // preview, not a full identity switch.
  if (isRealAdmin && authState.impersonatedRole === 'Anonymous') return {};
  ```

### Nitpicks (Optional)

- **[`packages/web-app/src/reducers/LoginReducer.property.test.js:220`]** The `expect(sessionStorageMock.setItem).not.toHaveBeenCalled()` sits outside the `fc.property` callback, which is the right place for a cumulative post-run assertion. However, `vi.clearAllMocks()` in `beforeEach` resets call counts *before* the `it` block runs, not between `fc.property` iterations — so if any run incorrectly calls `setItem`, it would be caught. This is fine; just worth noting for the next reader that this is intentionally a post-`fc.assert` check and not a within-iteration assertion.

- **[`packages/web-app/src/components/common/ImpersonationIndicator.jsx:75-95`]** The indicator's pill is `position: fixed` with `top: theme.spacing(1)` (8px), centred horizontally. On mobile the AppBar is also fixed at the top. There's no `paddingTop` adjustment on the page content to compensate for the extra 40px or so the pill occupies when active, which means it could overlap content near the top edge while a preview is running. Since the pill is admin-only and the admin is expected to use it deliberately, this is a cosmetic gap, not a functional regression — but a follow-up ticket to add a body offset (or use an AppBar slot) when active would round out the feature.

Paul-AUB:
Cette feature est déployée en staging désormais, je vous laisse y jeter un oeil quand vous aurez un moment @ClemRz @urien 
image

Paul-AUB:
✅ Fixed — The logout property now generates login states with any supported impersonated role or no role, and asserts that `LOGOUT` resets `impersonatedRole` to `null`. The login-then-logout property asserts the same invariant.

✅ Fixed — `useUserProperties` now documents why Anonymous collapses identity while other role previews intentionally keep the real token so personal content and profile data continue to load.

⏭️ Not changed — The cumulative `sessionStorageMock.setItem` assertion is already intentionally placed after `fc.assert`, as the review notes; no code change is needed.

⏭️ Not changed — The possible mobile indicator overlap is cosmetic, explicitly identified as a non-regression and better handled as a separate layout follow-up rather than expanding this PR.

Validation: the focused `LoginReducer.property.test.js` suite passes (6 tests), and full `yarn lint` passes.

urien (approved):
Cela marche parfaitement,
Je n'ai pas fait le test de toutes les fonctionnalités :-)
Mais le fonctionnement est très simple et agréable

PR #1521, opened 2026-08-21
feat(app): improve workflows, document viewing, and error recovery

### LINKED ISSUE #1518: Afficher un message de succès lors de la validation de documents
Aucun message indiquant le succès de chaque validation ne s'affiche lorsque des documents sont validés ; les documents concernés disparaissent simplement de la liste, sans que la validation ne soit confirmée.

Image

Paul-AUB:
## 🤔 What

This PR makes the day-to-day Grottocenter experience clearer, safer, and easier
to recover from.

### Follow what is happening

Users can now open a dedicated, paginated **Recent changes** page instead of
only seeing a short homepage excerpt. It is easier to catch up on new activity
without losing the context of older changes.

Recent changes page

### Browse and contribute documents more comfortably

Document lists use the available space better, so images and files are easier
to scan. Image captions can be hidden when they get in the way, zoomed PDFs can
be panned horizontally, and more PDF image formats render correctly. Submitting
a document now ends with a clearer confirmation, while destructive and
secondary actions are visually more consistent.

Improved document browsing

### Give moderators immediate feedback

After validating or declining documents, moderators now see whether the action
succeeded or failed instead of having rows silently disappear. This makes a
lost permission, network error, or server error distinguishable from a completed
review. **Closes #1518.**

### Recover gracefully from unexpected errors

When a page fails unexpectedly, users get a clear recovery screen: retry the
page, return home, contact Wikicaves, or report the issue on GitHub. Technical
details can be copied in one click, making support requests more useful without
exposing error messages or stack traces.

Error recovery screen

### Make multi-step tasks and navigation easier

CSV imports, observation imports, and MFA enrollment now share a responsive
progress indicator, so people always know where they are in a multi-step flow.
The account and notification menus, side menu, search filters, entity-table
toolbar, notifications, and entrance forms also receive focused usability and
layout improvements.

Responsive wizard progress

## 🤷‍♂️ Why

The application already supported these workflows, but several of them made
people work harder than necessary: activity was difficult to browse, document
previews could be awkward, moderation feedback was ambiguous, and an unexpected
error gave no clear next step. This PR turns those dead ends and rough edges
into explicit, understandable actions.

## 🔍 How

- Add a paginated activity route and refresh the homepage activity list.
- Improve document layout, lightbox, PDF preview, submission confirmation, and
  action styling.
- Add explicit moderator notifications for document-validation outcomes.
- Replace the misspelled page-error boundary with an accessible fallback that
  resets on navigation.
- Reuse one responsive `WizardProgress` component across CSV import,
  observation import, and MFA enrollment.
- Add translations, unit coverage, Storybook coverage for the reusable wizard,
  and focused UI refinements across related screens.

## 🔗 Related API PR

None.

## 🧪 Testing

- [ ] Manual browser verification of the responsive document layout, PDF preview,
  lightbox, validation feedback, recent-changes pagination, and error recovery.

## 📸 Previews

Screenshots are included alongside the relevant user-facing improvements above.


ClemRz (changes requested):
A solid improvement batch — the error boundary, WizardProgress consolidation, moderator feedback, and activity feed are all well-motivated and the implementations are generally clean. A few things worth fixing before merge.

### Issues (Must Fix)

- **[`src/components/appli/PageErrorBoundary.jsx:27`, `PageErrorBoundary.test.jsx:5`]** `copyToClipboard` is imported from `../../helpers/clipboard` — a deprecated folder. `helpers/clipboard.js` is an existing file in that folder, but the project convention says *no new imports* from `helpers/`. This PR introduces the import in a brand-new file, which is exactly the pattern the convention is trying to stop. Move `clipboard.js` to `src/utils/clipboard.js` (it's a pure utility with no deprecated dependencies) and update the import path here, in `AnchorCopyButton.jsx`, and in the test. If moving it is out of scope for this PR, a dedicated `tech/clipboard-to-utils` follow-up works too — but the new file should not land with a fresh `helpers/` import in the meantime.

- **[`src/components/appli/DocumentValidation/Actions.jsx:58` — `success` effect]** The `success` useEffect depends on `selectedIds.length` but not on `selectedIds` itself. If `selectedIds` changes without its length changing (e.g. two different documents are individually replaced one for one), `formatMessage` will receive the right `count` but the `onProcessed` callback will fire with a stale closure over the old `selectedIds`. Safer to list `selectedIds` as a dependency rather than `selectedIds.length`, or to capture the count at the moment of mutation rather than reading it from state in the effect.

  ```diff
  -    selectedIds.length,
  +    selectedIds,
  ```

- **[`src/pages/homepage/RecentChanges.jsx:14–26`]** Raw pixel values are used for spacing inside `styled()` components — `padding: '32px 24px'`, `padding: '24px 16px'` (responsive variant), `gap: 8`, and `marginBottom: 20`. The architecture conventions require `theme.spacing()` (with integer or fractional factors) or MUI `sx` shorthand props for ordinary spacing. These should be expressed as `theme.spacing(4, 3)`, `theme.spacing(3, 2)`, `gap: theme.spacing(1)`, and `marginBottom: theme.spacing(2.5)`.

- **[`src/components/common/DocumentsList/ImageLightbox.jsx:101–133`]** Same raw-pixel spacing concern at a lower priority (in a `styled` template literal, so less visible). `gap: 4px`, `gap: 8px`, `padding: 12px`, `padding: 24px 16px 12px` are scattered through the lightbox styled components. These are UI-dense interactions so pixel-exact control is understandable, but keeping them consistent with `theme.spacing()` prevents subtle drift if the theme's base unit ever changes.

- **[`src/components/common/RecentChangesList.jsx:183,260`]** `fontSize: 20` and `fontSize: 48` are raw integer values in `sx` props rather than theme references. MUI interprets numbers in `sx` as `px` for `fontSize`, but the typography scale isn't in play. Consider `fontSize: 'medium'` / `theme.spacing(6)` or just leaving a comment that these are intentional icon size overrides, so they aren't silently changed when someone audits for spacing issues.

- **[`src/hooks/queries/useHomepageLists.js:26–36`]** `fetchRecentChangesPage` fetches `limit + 1` items to probe for a next page, then slices off the extra. This is a clean cursor pattern — just worth noting that `nextOffset: offset + Math.min(changes.length, limit)` can drift if the server returns fewer than `limit` items (sparse pages near the 500-item ceiling). A comment documenting the intent and the 500-cap would help future readers understand why `hasMore` sometimes returns `false` without `changes.length < limit` being the only signal.

- **[`src/components/appli/PageErrorBoundary.jsx:62–70`]** `technicalReport` is memoised but its `useMemo` deps include `formatMessage` (which is a stable reference from `react-intl`) and `location.pathname`. The current UTC date string is captured at the time the error screen first renders, not at the time the memo recomputes — this is actually correct behavior. However, if the user navigates within the error boundary without triggering a reset (which `resetKeys={[location.key]}` prevents for full key changes, but not for same-key hash changes), the timestamp stays stale. This is an edge case worth a brief comment so the next engineer doesn't "fix" it by adding `Date.now()` to the deps and making it re-render constantly.

- **[`src/components/appli/EntitiesForm/Document/DocumentSubmissionSuccess.jsx`]** `onSubmitAnotherArticle` is always required in the PropTypes, but in the parent (`index.jsx`) the button that calls it is only rendered when `isArticle` is true. Consider making `onSubmitAnotherArticle` optional in PropTypes and adding a default no-op, or keeping it required but documenting the coupling to `isArticle`.

- **[`src/pages/RecentChanges/index.jsx:19`]** `const retry = changes.length > 0 ? fetchNextPage : refetch;` — this correctly differentiates "retry a pagination fetch" from "retry the initial fetch", but the name `retry` is slightly misleading in the `fetchNextPage` branch (it's not a retry, it's an advance). A comment or rename to `onLoadMore` / `onRetry` split at the call site would clarify intent for the `RecentChangesList`'s `onRetry` prop.

### Nitpicks (Optional)

- **[`src/utils/recentChanges.js`]** `getRecentChangeKey` joins fields with `-` but some values (dates, entity types) can contain `-` themselves. Consider using `|` or `||` as the separator, or `JSON.stringify([...])`, to reduce the risk of accidental key collisions.

- **[`src/components/appli/ImportCSV/ImportContainer.jsx:4`]** The named import `{ MuiLinearProgress }` is immediately re-exported as `LinearProgress` via `styled`. It's a minor style point, but importing as `LinearProgress` and aliasing to `StyledLinearProgress` (or just extending inline) would remove one indirection.

- **[`AGENTS.md`]** The checklist item now reads "ESLint passes on the modified lintable files (`yarn lint` before a PR)". This is clearer than the old version — good improvement.

- **[`src/components/common/Form/WizardProgress.stories.jsx`]** The Storybook story is present — appreciated. Consider adding a "mobile" story that forces a narrow viewport (using `parameters.viewport`) so the `LinearProgress` fallback can be visually verified in Storybook without needing to resize the browser.

urien (approved):
C'est parfait @Paul-AUB

urien:
En effet cela fonctionne : tu as surmonté la régression fourbe ;-)

ClemRz (changes requested):
All three issues flagged in the previous round are resolved — clipboard moved to `utils/`, `selectedIds` is now a proper effect dependency in `Actions.jsx`, and the homepage section spacing uses `theme.spacing()`. The overall quality of this batch is high. A few smaller things worth addressing before merge.

### Issues (Must Fix)

- **[`src/components/appli/EntitiesForm/Document/DocumentSubmissionSuccess.jsx:95`]** `onSubmitAnotherArticle` is declared `PropTypes.func.isRequired`, but the button that calls it is only rendered when `isArticle` is true — so a non-article document form would satisfy the contract by passing any function, even one that does nothing meaningful. The real risk is the reverse: if a future caller forgets to wire this prop when `isArticle` is false, they get a PropTypes warning at runtime but no crash because the button is never rendered. This is a benign coupling today (the only call site always supplies the prop), but the PropTypes contract is misleading. The previous review flagged this as a Suggestion, and the author's stated reason was "keeping it required catches incomplete integrations" — that reasoning is valid only if every integration is also an article form. Since the prop is gated by `isArticle`, declaring it `PropTypes.func` (optional) with the parent always passing it is both honest and equally safe. Either approach is fine, but if it stays `isRequired`, a comment explaining the coupling to `isArticle` is needed so the next engineer doesn't silently pass a no-op. As-is, the mismatch between "required" and "conditionally rendered" is a readability issue that can mislead.

  ```suggestion
  // onSubmitAnotherArticle is only called when isArticle is true;
  // it is still declared required so call sites must always wire it up.
  onSubmitAnotherArticle: PropTypes.func.isRequired,
  ```
  (Add the comment rather than changing the declaration if you prefer to keep it required.)

- **[`src/pages/RecentChanges/index.jsx:19`]** `const handleRetry = changes.length > 0 ? fetchNextPage : refetch;` passes `fetchNextPage` as the `onRetry` prop when there are already changes loaded, but `RecentChangesList`'s `FetchErrorState` labels this prop "Retry". From the user's perspective, pressing "Retry" after a pagination failure should retry the last page, which is what `fetchNextPage` does — so the behaviour is correct. However, the name `handleRetry` reads ambiguously in the "has changes" branch (it's actually "load next page", not a retry). Consider splitting the prop names at the call site or adding a brief comment, e.g.:
  ```js
  // When changes are already shown, the error is on a page fetch — retry that
  // page. When there are none, the initial load failed — refetch from scratch.
  const handleRetry = changes.length > 0 ? fetchNextPage : refetch;
  ```

- **[`src/hooks/queries/useHomepageLists.js:36`]** `nextOffset: offset + Math.min(changes.length, limit)` handles sparse pages correctly but the 500-row API cap causes a subtlety: the last page before the cap will have `changes.length === limit` (extra-row probe returned `limit + 1` items, so `hasMore = true`), yet the next fetch may return fewer than `limit` items (the tail end of the 500-row window). That next fetch correctly sets `hasMore = false`. This all works as intended, but the current comment only mentions the cap, not the implication for the last page. A brief addition would help future readers:
  ```js
  // The API caps this feed at 500 rows. When the cap is hit, the last page
  // returns fewer than `limit` items → hasMore becomes false on the next probe.
  ```

- **[`src/components/appli/PageErrorBoundary.jsx:62`]** The `technicalReport` memo's deps include `location.pathname` but not `location.search` or `location.hash`. If the error occurs on a URL with query parameters (e.g. `/ui/documents?type=article`), the copied report will omit the query string, potentially making it harder to reproduce. Consider using the full `location.href` equivalent — `location.pathname + location.search + location.hash` — or `window.location.href` directly, so the report is complete. Since the memo is frozen at first render (intentionally), this doesn't introduce re-renders; it just makes the captured address more useful.

- **[`src/components/common/DocumentsList/DocumentsList.jsx`]** The diff adds masonry layout utilities and a denser document grid. Worth verifying (if not already done) that the `masonryUtils.js` column calculation accounts for containers with padding, since the column width calculation may otherwise produce a slightly wrong number of columns at border breakpoints. The accompanying test covers a representative range, so this is mainly a note to check against the actual rendered output.

### Nitpicks (Optional)

- **[`src/components/appli/EntitiesForm/Document/index.test.jsx:1`]** The file is named `index.test.jsx` but tests `DocumentSubmissionSuccess`, not the `index.jsx` default export (`HydratedDocumentSubmission`). The test would be more discoverable as `DocumentSubmissionSuccess.test.jsx`, matching the component filename.

- **[`src/pages/homepage/RecentChanges.jsx:62–70`]** The "See all recent changes" link wraps a `<Typography>` in an `<AppLink>`. The typography applies `textDecoration: 'underline'` via `sx`, but `AppLink` likely already handles underline styling. Consider whether the link's own underline styling is redundant with MUI's `Link` defaults (and whether `cursor: 'pointer'` is needed, since links are pointer by default).

- **[`.agents/skills/fix-review/SKILL.md`]** This new skill file is a useful addition to the agent workflow. One minor note: step 4 says "read the current state of that file, not just the diff hunk" — this is exactly the right instruction and matches how this review was conducted.

Paul-AUB:
Addressed the follow-up review in `7c8a4e6d8`:

- ✅ Fixed — documented why `onSubmitAnotherArticle` remains required even though it is only called for article submissions.
- ✅ Fixed — documented why retrying delegates to `fetchNextPage` when existing changes show that pagination failed, and to `refetch` when the initial load failed.
- ✅ Fixed — expanded the recent-changes pagination comment to explain how the shorter tail of the API's 500-row window closes pagination.
- ✅ Fixed — technical error reports now include the pathname, query string, and hash; added regression coverage for the complete address.
- ⏭️ Not changed — verified that the masonry layout does not manually calculate its column count: CSS Grid calculates columns from the grid's content box, while `disablePadding` removes the MUI List inset. The targeted masonry utility tests also pass.
- ✅ Fixed — renamed `index.test.jsx` to `DocumentSubmissionSuccess.test.jsx` so the tested component is discoverable by filename.
- ✅ Fixed — removed the redundant nested Typography underline and cursor styling; `AppLink` now owns the typography and link styling directly.

Validation: full `yarn lint` passes, and 5 targeted test files pass (11 tests).

Paul-AUB:
Fixed the broken view in commit [fd82daa](https://github.com/GrottoCenter/grottocenter-front/pull/1521/commits/fd82daaa044d726f456656fe324662fe209e1ea9), now it looks fine :

image

ClemRz (changes requested):
All three Issues flagged in round 2 have been addressed: `onSubmitAnotherArticle` now carries a clarifying comment alongside its `isRequired` declaration, `handleRetry` has a comment explaining the two branches, and the pagination comment was expanded. The full-address `location.pathname + location.search + location.hash` technical report fix is also in. The codebase is in good shape overall.

One small Issue remains before merge, plus a handful of lighter notes.

### Issues (Must Fix)

- **[`src/components/appli/EntitiesForm/Document/DocumentSubmissionSuccess.jsx:94`, `DocumentSubmissionSuccess.test.jsx`]** The comment added above `onSubmitAnotherArticle` reads:
  ```
  // onSubmitAnotherArticle is only called when isArticle is true;
  // it is still declared required so call sites must always wire it up.
  ```
  The intent is sound, but the test only exercises the `isArticle = true` branch — it never renders the component with `isArticle = false` and verifies that **no** call to `onSubmitAnotherArticle` occurs even when it is passed (which would confirm the "required but conditionally unused" contract). A non-article submission form would satisfy the `isRequired` PropType by passing a no-op, yet the component would silently render without the article-only hint text or button. The test should add a case for `isArticle = false` asserting that neither the article button nor the hint text appears, and that `onSubmitAnotherArticle` is never called. Without that coverage, the coupling documented in the comment is invisible to the test suite.

- **[`src/components/appli/AdvancedSearch/SearchElements.jsx:765–793`]** The `SearchActionButtons` component no longer accepts any props (`onReset`, `showReset`), and all call sites have been migrated. The reset behaviour has been folded into the `ActiveFilterChips` `onClearAll` callback instead. This is a cleaner design — just worth noting that the old "Reset stays available offline" comment was removed from `SearchActionButtons` without a replacement. The new `ClearAllButton` inside `ActiveFilterChips` is not wrapped in `OfflineDisabled`, which means "Clear all" is available offline. That is arguably correct (clearing local state needs no network), but it is worth an explicit comment on the `onClearAll` prop to confirm this was intentional, since the old code specifically called it out.

- **[`packages/web-app/vite/serviceWorkerBuildGuard.mjs`]** The guard reads the generated `sw.js` in the `closeBundle` post hook. This is a good defensive measure — the comment and pattern are clear. One edge case: the `readFile` call will throw `ENOENT` if the build did not produce a service worker (e.g. when running `vite build --mode development` with PWA disabled). Consider wrapping the read in a `try/catch` that swallows `ENOENT` specifically, so the guard fails silently on non-PWA builds rather than crashing the whole bundle step.

- **[`src/components/common/DocumentsList/DocumentsList.jsx:65–70`]** The masonry layout renders `gridAutoRows: ${MASONRY_ROW_HEIGHT}px`. `MASONRY_ROW_HEIGHT` is imported from `./utils/masonryUtils`. `DocumentsGridItem` reads `theme.spacing(1)` for the vertical gap, which resolves to `8px` by default. `calculateMasonryRowSpan` presumably takes both values as parameters — worth confirming in the util that the `verticalGap` parameter matches what the grid actually uses (i.e. `theme.spacing(1)` evaluated as a number), otherwise cards at height boundaries will be off by a row.

### Nitpicks (Optional)

- **[`src/components/appli/PageErrorBoundary.jsx:176`]** `<Button component={AppLink} … to="/" underline="none">` — passing `underline` to a `Button` has no effect (it is an `AppLink`/`Link` prop). The `underline="none"` can be dropped; `Button` already renders its content without the default link underline.

- **[`packages/web-app/src/pages/homepage/RecentChanges.jsx:52`]** `HistoryOutlinedIcon` is given `sx={{ fontSize: 28 }}` inline. Per the project convention (`AGENTS.md`), icon sizes should use theme-backed values (`theme.spacing(3.5)` = 28px, or the MUI named font-size tokens). This is a minor drift; `fontSize: theme.spacing(3.5)` would keep it consistent with the rest of the codebase.

- **[`.agents/skills/fix-review/SKILL.md`]** Solid new skill — the guidance in step 4 to "read the current state of that file, not just the diff hunk" is exactly what makes multi-round reviews tractable.

Paul-AUB:
Addressed the latest review in `4119e006b`:

- ✅ Fixed — added non-article coverage confirming the article-only hint and action stay hidden and `onSubmitAnotherArticle` is never called.
- ✅ Fixed — documented that clearing advanced-search form state deliberately remains available offline because it makes no network request.
- ⏭️ Not changed — the service worker guard should keep failing if `sw.js` is missing: the PWA plugin is always enabled for builds, including development-mode builds, so a missing artifact indicates a broken build that the guard should surface rather than hide.
- ✅ Verified — the masonry measurement uses `Number.parseFloat(theme.spacing(1))`, exactly matching the grid's `theme.spacing(1)` vertical gap; the focused boundary/rounding tests pass, so no code change was needed.
- ✅ Fixed — removed the ineffective `underline="none"` prop from the home button.
- ✅ Fixed — converted the homepage history icon's 28px size to `theme.spacing(3.5)`.

Validation: full `yarn lint` passes, and 2 focused test files pass (5 tests).

